PR #1, opened 2019-02-28
Initial implementation of the vm-memory crate

jiangliu:
Initial implementation of the vm-memory crate


jiangliu:
Please refer to following links for related discussions:
https://github.com/rust-vmm/community/issues/22
https://github.com/rust-vmm/community/issues/16

README.md
```diff
@@ -1,2 +1,89 @@
 # vm-memory
-Virtual machine's guest memory crate
+A library to manage and access virtual machine's physical memory.
```
andreeaflorescu: This seems like an in depth description of the crate that would probably have a better place in a design document. I find short readmes with links to other documents easier to follow (it also looks prettier on crates.io).
The readme is the front page of the crate, so usually what I notice is that people have the following subsections in it:
- Summary
- Minimal Usage Example
- Contributions (optional)
- Links to in depth documentation
- License

Some examples of such crates:
- [backtrace](https://crates.io/crates/backtrace)
- [serde](https://crates.io/crates/serde)
- [lazy_static](https://crates.io/crates/lazy_static)

I find it easier to have a `Summary` at the beginning so the user can get a grasp on what the crate is and what is it useful for. It also worth having information in the readme from the user perspective. How can you use the vm-memory crate? What is it doing for you?

README.md
```diff
@@ -1,2 +1,89 @@
...
+First, add the following to your `Cargo.toml`:
+```toml
+vm-memory = "0.1"
+```
+Next, add this to your crate root:
+```rust
+extern crate vm-memory;
+```
+
+# TODO List
```
andreeaflorescu: nit: Should we move this to a TODO file instead?

.gitignore
```diff
@@ -0,0 +1,3 @@
+/target
+**/*.rs.bk
+Cargo.lock
```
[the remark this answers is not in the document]
bonzini: Libraries should ignore Cargo.lock, only applications should track it.

https://github.com/rust-lang/cargo/issues/315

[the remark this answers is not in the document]
andreeaflorescu: While reproducible builds is not a must for libraries, keeping control of the dependencies should be. `Cargo.lock` helps in having a finer grained control of your dependencies. When upgrading the dependencies of your package you will not have a clear view of transient dependencies. Cargo.lock can help with that. It also helps in spotting if one of your dependencies has a license incompatible with Apache-2.0 which is something we should keep an eye on.

There are probably other ways to make sure that we do not increase the number of dependencies & that the license of dependencies is compatible, so if we find an alternative that can give us that without adding Cargo.lock to the repository I am fine with it.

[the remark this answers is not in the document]
bonzini: I understand why you want Cargo.lock, but it's not the policy for crates and I'd rather respect that, especially for a crate that has exactly 1 depedency (and I don't expect most initial rust-vmm crates to have much more than that).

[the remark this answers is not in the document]
andreeaflorescu: We already have 24 dependencies (transient + direct). Most of them are added by `tempfile`. This is a crate only used in tests. I guess we do not care about dev dependencies as they should not be included in the release binary. Cargo has a strange behaviour there because all dev dependencies are pulled even for a release build.

src/lib.rs
```diff
@@ -0,0 +1,289 @@
+// Portions Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Portions Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the THIRD-PARTY file.
+
+//! Provides a wrapper for allocating, handling and interacting with the guest memory regions.
```
andreeaflorescu: Can you add some examples of how the vm-memory is going to be used?

jiangliu: add example to lib.rs?
I have added an simple example to README.md.

.gitignore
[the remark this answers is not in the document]
bonzini: Yes, they are pulled because you can do `cargo test --release`. But the only actual dependency of the library is libc.

Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "vm-memory"
+version = "0.1.0"
+authors = ["Liu Jiang <gerry@linux.alibaba.com>"]
+
+[features]
+default = ["vm-memory-backend-mmap"]
```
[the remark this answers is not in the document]
bonzini: This works for me:

    index d91efba..52376a5 100644
    --- a/src/mmap.rs
    +++ b/src/mmap.rs
    @@ -339,9 +339,7 @@ impl GuestMemoryMmap {
                    }
                 }
     
    -            // seems like a false warning
    -            #[allow(clippy::redundant_closure)]
    -            let mapping = MmapRegion::new(range.1).map_err(|e| MmapError::SystemCallFailed(e))?;
    +            let mapping = MmapRegion::new(range.1).map_err(MmapError::SystemCallFailed)?;
                 regions.push(GuestRegionMmap {
                     mapping,
                     guest_base: range.0,

src/lib.rs
andreeaflorescu: The benefit of having examples in the code is that they're also run with cargo test and published along with the public interface and the documentation to [docs.rs](https://docs.rs/) whenever you run `cargo publish`. Also, it is a nice way to keep examples up to date with the code (cargo test will crash if the examples are out of date). From my point of view it is easier to start working with a crate if you have examples of how to use the public interface in the docs.

We can have this is as a nice to have feature. What are your thoughts on this?

src/mmap.rs
```diff
@@ -0,0 +1,585 @@
...
+                    .guest_base
+                    .checked_add(last.mapping.len() as GuestAddressValue)
+                    .map_or(true, |a| a > range.0)
+                {
+                    return Err(MmapError::MemoryRegionOverlap);
+                }
+            }
+
+            // seems like a false warning
+            #[allow(clippy::redundant_closure)]
```
andreeaflorescu: Hmm.. This is strange. It should work like:
```rust
let mapping = MmapRegion::new(range.1).map_err(MmapError::SystemCallFailed)?;
```
What rust version are you using? Maybe it is something related to the version?
I wrote a minimal example in [Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=27f4da2774fd8aaef232b5b81a5d0534).

DESIGN.md
```diff
@@ -0,0 +1,79 @@
...
+	- map a request address to a GuestMemoryRegion object and relay the request to it.
+	- handle cases where an access request spanning two or more GuestMemoryRegion objects.
+
+The VM memory consumers, such as virtio device drivers, vhost drivers and boot loaders etc, should only rely on traits and structs defined here to access VM's physical memory.
+
+### A Sample and Default Backend Implementation Based on mmap()
+Provide a default and sample implementation of the GuestMemory trait by mmapping VM's physical memory into current process. Three data structures are defined here:
+- MmapRegion: mmap a continous range of VM's physical memory into current and provide methods to access the mmapped memory.
+- GuestRegionMmap: a wrapper structure to map VM's physical address into (mmap\_region, offset) tuple.
+- GuestMemoryMmap: manage a collection of GuestRegionMmap objects for a VM.
```
sameo: I think here we want to highlight the fact that `GuestMemoryMmap` is the `GuestMemory` implementation for an mmap'ed volatile memory backend.
That's our top level VM memory interface for the mmap backend, and we should highlight that.

I'm thinking about adding a sub-section to this section with really simple examples on how  to use `GuestMemoryMmap`. I'll do so as a follow-up PR.

DESIGN.md
```diff
@@ -0,0 +1,79 @@
...
+	let addr: GuestAddress = ...
+        let buf = &mut [0u8; 5];
+	let result = guest_memory_mmap.find_region(addr).unwrap().write_slice(buf, addr);
+```
+- use following pattern for functionality to support request crossing region boundary:
+```rust
+	let guest_memory_mmap: GuestMemoryMmap = ...
+	let addr: GuestAddress = ...
+        let buf = &mut [0u8; 5];
+	let result = guest_memory_mmap.write_slice(buf, addr);
```
sameo: I think those 2 examples are a little confusing:

### Why is this example crossing the memory region boundary?
We need to be more explicit on how we build `guest_memory_mmap`, maybe something like that (Not built...):

```Rust
const GUEST_START: u64 = 0x1000;
const GUEST_SIZE: usize = 0x2000;

let guest_start_addr = GuestAddress(GUEST_START);
let guest_memory: GuestMemoryMmap::new(&vec![(guest_start_addr, GUEST_SIZE)]);

/// End address - 0x10
let buf_guest_addr = GuestAddress(GUEST_START + GUEST_SIZE as u64 - 0x10);

/// Slice is to large
let buf = &mut [0u8; 0x100];

let result = guest_memory_mmap.find_region(addr).unwrap().write_slice(buf, buf_guest_addr);
```

### Why does the first one fails while the second succeeds?

Not sure how this would build. How do you call `write_slice` on a `GuestMemoryMmap` ?:
```Rust
let result = guest_memory_mmap.write_slice(buf, addr);
```

Also, did you mean `write` and not `write_slice`?

DESIGN.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+### Utilities and Helpers
+Following utility and helper traits/macros are imported from the [crosvm project](https://chromium.googlesource.com/chromiumos/platform/crosvm/) with minor changes:
+- DataInit: Types for which it is safe to initialize from raw data. A type `T` is `DataInit` if and only if it can be initialized by reading its contents from a byte array. This is generally true for all plain-old-data structs.  It is notably not true for any type that includes a reference.
+- {Le,Be}\_{16,32,64}: Explicit endian types useful for embedding in structs or reinterpreting data.
+
+## Relationship among Traits and Structs
```
sameo: I would replace that with:

## Relationships between Traits, Structs and new Types.

DESIGN.md
```diff
@@ -0,0 +1,79 @@
...
+traits:
+- Address: AddressValue
+- GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> (must be implemented)
+- GuestMemory: Bytes<GuestAddress> (generic implementation)
+
+newtypes:
```
sameo: New Types

DESIGN.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+### Utilities and Helpers
+Following utility and helper traits/macros are imported from the [crosvm project](https://chromium.googlesource.com/chromiumos/platform/crosvm/) with minor changes:
+- DataInit: Types for which it is safe to initialize from raw data. A type `T` is `DataInit` if and only if it can be initialized by reading its contents from a byte array. This is generally true for all plain-old-data structs.  It is notably not true for any type that includes a reference.
+- {Le,Be}\_{16,32,64}: Explicit endian types useful for embedding in structs or reinterpreting data.
+
+## Relationship among Traits and Structs
+traits:
+- Address: AddressValue
```
sameo: For clarity, maybe be a little more explicit: `Address` implements `AddressValue`

DESIGN.md
```diff
@@ -0,0 +1,79 @@
...
+- GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> (must be implemented)
+- GuestMemory: Bytes<GuestAddress> (generic implementation)
```
sameo: Ditto.

DESIGN.md
```diff
@@ -0,0 +1,79 @@
...
+
+newtypes:
+- GuestAddress: Address\<u64\>
+- MemoryRegionAddress: Address\<u64\>
+
+structs:
+- MmapRegion: VolatileMemory
```
sameo: Here as well, replace `:` with `implements`

src/mmap.rs
jiangliu: yah, it works. Haven't realized that rust has sugar for this case.
Will wait for more comments and fix this in coming patches.

DESIGN.md
jiangliu: Really thanks for detailed review:)
Will refine the documentation with following patches.

DESIGN.md
sameo: nm, `GuestMemoryMmap` is a `GuestMemory`, who implements `Bytes`. So calling `write_slice` on it would make sense. But I still wonder if you meant `write` instead of `write_slice`...

src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+/// - data files
+/// - a proxy to access memory on remote
+pub trait Bytes<A> {
+    /// Associated error codes
+    type E;
+
+    /// Writes a slice to the region at the specified address.
+    /// Returns the number of bytes written. The number of bytes written can
+    /// be less than the length of the slice if there isn't enough room in the
+    /// region.
```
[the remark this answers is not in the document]
jiangliu: Yes, the only difference is that write_slice() verifies that the whole buffer has been write successfully.
But it's hard to enumerate possible error cases of write() because it depends on the backend implementation.

src/lib.rs
jiangliu: Good suggestion. 
How about covering it with a coming PR? We should also ensure the content generated by "cargo doc" 
is suitable.

alexandruag:
Hi, for how long is this still open for comments?

jiangliu:
Hope it could be closed within this week:)
If there are still improvements required, we may do them with other PRs.

sameo:
> Hi, for how long is this still open for comments?

Until at least 2 team members will approve it. We can definitely hold it until you review it.

alexandruag:
> > Hi, for how long is this still open for comments?
> 
> @alexandruag Until at least 2 team members will approve it. We can definitely hold it until you review it.

Thanks, I just had some comments but also wanted to really try and understand the entire thing first. Will post something before the end of this week if still allowed :D

sameo:
@alexandruag Did you have time to look at this PR?

alexandruag:
> @alexandruag Did you have time to look at this PR?

I still have to go through some things. It would be great if you can allow us a couple more days. I'm very very sorry about the delay, and it's totally understandable if you want to draw a line and move forward.

src/mmap.rs
```diff
@@ -0,0 +1,583 @@
...
+        })
+    }
+
+    /// Maps the `size` bytes starting at `offset` bytes of the given `fd`.
+    ///
+    /// # Arguments
+    /// * `fd` - File descriptor to mmap from.
+    /// * `size` - Size of memory region in bytes.
+    /// * `offset` - Offset in bytes from the beginning of `fd` to start the mmap.
+    pub fn from_fd(fd: &AsRawFd, size: usize, offset: libc::off_t) -> io::Result<Self> {
```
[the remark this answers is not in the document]
jiangliu: This interface may also be used to manage memfd backed guest memory, so the memfd file descriptors could be passed to other processes by the vhost-user crate.

README.md
```diff
@@ -1,2 +1,86 @@
-# vm-memory
-Virtual machine's guest memory crate
+# memory-model
```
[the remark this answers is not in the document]
jiangliu: sure, will fix it.

src/mmap.rs
```diff
@@ -0,0 +1,583 @@
...
+    }
+}
+
+impl Drop for MmapRegion {
+    fn drop(&mut self) {
+        // This is safe because we mmap the area at addr ourselves, and nobody
+        // else is holding a reference to it.
+        unsafe {
+            libc::munmap(self.addr as *mut libc::c_void, self.size);
+        }
```
sboeuf: A few comments regarding this behavior.

First, this assumes that part of the memory security is enforced through the code, instead of being enforced by the language itself. The assumption that nobody else is holding a reference is very hard to enforce...

The way `mmap()` behaves, we can clearly expect that `mmap()` would be called multiple times on the same address with different ranges, without having to call into `munmap()`. Calling `munmap()` would prevent us from atomically change the mapping at a specific fixed address.
This is a very specific case that I ran into while porting virtio-fs for Firecracker (DAX case), since the memory mapping that happens on behalf of the libfuse daemon is performed on top of the existing mapping holding the whole shared region.

Problem is, by having this `Drop` behavior, this forces the virtio-fs device to maintain a list of `MemoryMapping` structures if we don't want to see the mapping disappear. And even by maintaining a list, what happens when someone wants to `mmap()` on top of an address that is already tracked?

Ok let's try to take an example:
1. Map 16MB of anonymous pages at address 0x0
2. Map 4MB of pages from a file descriptor at address 0x200000 (2MB)
3. At this point, you have two entries in the list of `MemoryMapping`
4. Unmap 16MB of pages at address 0x0
5. At this point you only have one entry corresponding to the second mapping
6. Trying to access the page at address 0x200000 will generate a segfault

What I'm trying to say is that by enforcing the unmap from the code itself, we need deterministic behavior. In the case of virtio-fs, the hypervisor is doing actions (`mmap()` basically) on behalf of the vhost-user backend, that's why it seems pretty hard to maintain proper state every time a `mmap()` happens.

My question is, do we want to keep this behavior for any memory mapping, by maintaining a complex list of descriptors, or do we simply want to create another type of `MmapRegion`, like `MmapRegionRaw` that would not include any `Drop` behavior?

mcastelino: @sboeuf Actually this now really a `MmapRegionRaw`. This is a special type of region, where you want to free mapping to the mapped contents but not give us the virtual memory reservation.

To be able to do this atomically you cannot do an unmap, but you need to do a mmap of anonymous memory at the same virtual address with the same same size. Hence the VMM gets to hold on to the virtual address reservation.

The only assumption here is that there is an overarching region that covers all of these *special" region that will outlive all of these sub regions and that will be responsible for the actual mmap.

sboeuf: @mcastelino you're right! Having a special type handling the `Drop` with `mmap(MAP_ANONYMOUS)` instead of `munmap()` looks like the only clean way to address this issue.

@jiangliu @bonzini WDYT?

src/guest_memory.rs
```diff
@@ -0,0 +1,237 @@
...
+pub trait GuestMemory: AddressSpace<GuestAddress, Error> {
+    /// Invoke callback `f` to handle data in the address range [addr, addr + count).
+    ///
+    /// The address range [addr, addr + count) may span more than one AddressRegion objects, or
+    /// even has holes within it. So try_access() invokes the callback 'f' for each AddressRegion
+    /// object involved and returns:
+    /// - error code returned by the callback 'f'
+    /// - size of data already handled when encountering the first hole
+    /// - size of data already handled when the whole range has been handled
+    fn try_access<F>(&self, count: usize, addr: GuestAddress, mut f: F) -> Result<usize, Error>
```
alexandruag: I'm a bit apprehensive about this function and its uses throughout the PR. I'm mainly concerned about the way `Bytes` is implemented for `T: GuestMemory` (somewhere in a later commit), and the fact that we silently allow reads/writes across multiple regions (even if they have to be adjacent) by default. I think that cases where a read/write spills into another region are erroneous situations more often than not, and the implementation of `Bytes` for `T: GuestMemory` should not touch more than one region. 

To go around this limitation for cases where it actually makes sense, we can offer a helper method like `try_access` that iterates through all relevant regions and invokes some logic passed by the user. What do you think?

bonzini:
I will take a look at Alexandu's review and make some changes tomorrow.

src/mmap.rs
```diff
@@ -0,0 +1,1200 @@
...
+// safety.
+unsafe impl Send for MmapRegion {}
+unsafe impl Sync for MmapRegion {}
+
+impl MmapRegion {
+    /// Creates an anonymous shared mapping of `size` bytes.
+    ///
+    /// # Arguments
+    /// * `size` - Size of memory region in bytes.
+    pub fn new(size: usize) -> Result<Self> {
```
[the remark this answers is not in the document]
bonzini: This can be added later when there is a need.

src/address.rs
```diff
@@ -80,13 +80,19 @@ pub trait Address:
     /// Returns the result of the add or None if there is overflow.
     fn checked_add(&self, other: Self::V) -> Option<Self>;
 
+    /// Returns the result of the add and a flag identifying whether there was overflow
+    fn overflowing_add(&self, other: Self::V) -> (Self, bool);
```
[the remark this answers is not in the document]
bonzini: There is no particular usecase, but it's how DMA works. Same for `try_access` which you remarked on below.

src/endian.rs
```diff
@@ -0,0 +1,148 @@
+// Copyright 2017 The Chromium OS Authors. All rights reserved.
```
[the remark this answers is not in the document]
bonzini: It's a useful tool to use together with Bytes/VolatileMemory accesses, so we're putting it in the same crate.

Unfortunately due to orphan trait rules it's not possible to place it, together with Bytes and VolatileMemory, in a separate crate than GuestMemory.

src/mmap.rs
```diff
@@ -0,0 +1,1200 @@
...
+//! The main structs to access guest's memory are:
+//! - MmapRegion: mmap a continuous region of guest's memory into current process
+//! - GuestRegionMmap: map from guest physical address into mmapped offset
+//! - GuestMemoryMmap: manage a collection of GuestRegionMmap objects
+
+use libc;
+use std::io::{self, Read, Write};
+use std::mem;
+use std::ops::{BitAnd, BitOr};
+use std::os::unix::io::AsRawFd;
```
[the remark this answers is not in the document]
bonzini: Non-Unix platforms do not have `mmap` at all. If we want to support Windows later, we can make `from_fd` Unix-only.

bonzini:
> Why not switch to Rust 2018? (I know Firecracker is still using 2015 👎 )

Do you have some specific 2018 features in mind? Even `?` technically is one.

The only limit is that we cannot use nightly features because we want to build this code in distro packages and not just distribute it as crates (and distro rustc packages don't enable nightly features).

> I generally don't like wildcard imports (I don't how to properly call them) because they make code harder to understand, and every now and then you end up exposing something you don't want. What do you think about only using explicit imports, unless there is a very compelling reason to do otherwise?

Will fix.

> It looks like the crate would be easier to inspect if we split the code a bit more into modules. What do you think about placing things like Bytes, VolatileMemory, VolatileSlice, VolatileRef, etc. into they own modules (assuming this doesn't makes testing too awkward)?

Unfortunately it's not possible due to orphan trait rules:

    impl<T: GuestMemory> Bytes<GuestAddress> for T {
  
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type parameter `T` must be used as the type parameter for some local type


:(

Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "memory-model"
```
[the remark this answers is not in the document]
jiangliu: It has been renamed as "vm-memory" by later commits.

src/mmap.rs
```diff
@@ -336,6 +338,8 @@ impl GuestMemoryMmap {
                 }
             }
 
+            // seems like a false warning
+            #[allow(clippy::redundant_closure)]
             let mapping = MmapRegion::new(range.1).map_err(|e| MmapError::SystemCallFailed(e))?;
```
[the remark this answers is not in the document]
jiangliu: Thanks, I have fixed it according to Andreea's suggestion.

src/mmap.rs
```diff
@@ -0,0 +1,583 @@
...
+
+/// A backend driver to access guest's physical memory by mmapping guest's memory into current
+/// process.
+/// For a combination of 32-bit hypervisor and 64-bit virtual machine, only partial of guest's
+/// physical memory may be mapped into current process due to limited process virtual address
+/// space size.
+#[derive(Debug)]
+pub struct MmapRegion {
+    addr: *mut u8,
+    size: usize,
```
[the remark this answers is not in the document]
jiangliu: How about adding another abstraction of AddressSpace to support the vhost crate?
The idea is to manage guest's physical address space with an AddressSpace object, which contains a group of address regions. Each address region maintains type(memory or MMIO), properties(RWX), Usage, GPA, optional fd, optional offset into fd etc.
And GuestMemory is a mechanism for current process to access the guest memory and it could be generated from the AddressSpace object on demand. 
Currently qemu maps all the guest memory into current process, but we have some thoughts only to map partial or even none of guest memory into the hypervisor process. For the vhost-user crate, we could simple generate a group of address regions accessible by IO subsystem (or specific device backend) and pass the (GPA, fd, offset) tuple to vhost-user backends.
What's your thoughts here?
Thanks!

jiangliu:
> I will take a look at Alexandu's review and make some changes tomorrow.
Sorry for slow response, was on travel last week. So I will wait for your updates and handle remaining issues.

andreeaflorescu:
> > Why not switch to Rust 2018? (I know Firecracker is still using 2015 -1 )
> 
> Do you have some specific 2018 features in mind? Even `?` technically is one.
> 
> The only limit is that we cannot use nightly features because we want to build this code in distro packages and not just distribute it as crates (and distro rustc packages don't enable nightly features).

Rust 2018 is indeed not yet available in distros, maybe we can wait a bit before switching to Edition 2018. I think the current version in the ones that I've tried is 1.30.0.

src/mmap.rs
sboeuf: I didn't know about `std::mem::forget`, but this could be a potential solution to the problem too.
That being said, @mcastelino suggestion would be the cleanest approach IMO since we would make sure the memory region is reserved without leaving the previous mapping.

This can be part of a follow up PR, I don't mind, but I want to make sure we discuss and decide how we want to solve this asap, as this may influence the current code.

DESIGN.md
jiangliu: It's a special behavior to be backward compatible with firecracker, which fails when a write_slice() request crosses the memory region boundary. So the first example shows the way maintain current firecracker behavior. If it causes confusion, I will remove it.

jiangliu: Will improve the examples for readability:)

src/guest_memory.rs
```diff
@@ -0,0 +1,445 @@
...
+
+/// Type to encode offset in the guest physical address space.
+pub type GuestAddressOffset = <GuestAddress as AddressValue>::V;
+
+/// Represents a continuous region of guest physical memory.
+pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
+    /// Get the size of the region.
+    fn len(&self) -> GuestAddressValue;
+
+    /// Check wheather the region is empty.
```
ericho: Little typo on `wheater`

src/endian.rs
```diff
@@ -0,0 +1,148 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+//! Explicit endian types useful for embedding in structs or reinterpreting data.
+//!
+//! Each endian type is guarnteed to have the same size and alignment as a regular unsigned primiive
```
ericho: Typos on `guarnteed` and `primiive`

src/mmap.rs
```diff
@@ -0,0 +1,583 @@
+// Copyright (C) 2019 Alibaba Cloud Computing. All rights reserved.
```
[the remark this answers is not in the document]
bonzini: Because the contents use`VolatileMemory`, we would have to take multiple mutable references to an `MmapMut`.

src/mmap.rs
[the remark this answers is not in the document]
jiangliu: Currently we focus on Unix platforms, and a PR for windows support is on the fly. So we may improve it when enabling support of Windows platforms.

src/mmap.rs
```diff
@@ -0,0 +1,1200 @@
...
+        })
+    }
+
+    /// Maps the `size` bytes starting at `offset` bytes of the given `fd`.
+    ///
+    /// # Arguments
+    /// * `fd` - File descriptor to mmap from.
+    /// * `size` - Size of memory region in bytes.
+    /// * `offset` - Offset in bytes from the beginning of `fd` to start the mmap.
+    pub fn from_fd(fd: &AsRawFd, size: usize, offset: usize) -> Result<Self> {
```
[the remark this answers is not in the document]
jiangliu: Will improve it when enabling support of Windows platforms.

src/guest_memory.rs
jiangliu: The main target is to support memory hotplug(hot-adding). A request crossing region boundary may be valid if memory hot-adding is enabled. Providing a helper such as try_access() doesn't solve the issue because the consumer has no way to tell whether a memory region is hot-added or not so it will always use try_access().

andreeaflorescu: @jiangliu can we add another feature to vm-memory called "hot-plug" or something of that sort? @alexandruag would that be an acceptable compromise?

src/address_space.rs
```diff
@@ -0,0 +1,307 @@
...
+    fn checked_sub(&self, other: Self::V) -> Option<Self>;
+
+    /// Returns the result of the subtraction.
+    /// Only use this when `other` is guaranteed not to underflow.
+    fn unchecked_sub(&self, other: Self::V) -> Self;
+}
+
+#[macro_export]
+macro_rules! impl_address_ops {
+    ($T:ident, $V:ty) => {
```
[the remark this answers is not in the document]
jiangliu: It will cause compilation errors like:
error: expected expression, found `GuestAddress`
   --> src/address.rs:111:17
    |
111 |                 $T(value)
    |                 ^^ expected expression
    | 
   ::: src/guest_memory.rs:107:1
    |
107 | impl_address_ops!(GuestAddress, u64);
    | ------------------------------------- in this macro invocation

src/guest_memory.rs
bonzini: I don't think it is a matter of compromising; one behavior is right and the other is wrong.

Hotplug is an example, but in general this is just how DMA works, and at its heart a VMM is just emulating hardware. Firecracker does not implement this behavior but it's incorrect. The purpose of rust-vmm should also be to avoid this kind of shortcuts and do things the right way.

src/volatile_memory.rs
```diff
@@ -0,0 +1,656 @@
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn copy_to<T>(&self, buf: &mut [T])
+    where
+        T: DataInit,
+    {
+        let mut addr = self.addr;
+        for v in buf.iter_mut().take(self.size / size_of::<T>()) {
+            unsafe {
```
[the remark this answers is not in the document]
bonzini: @zachreizner This code comes from crosvm, can you reply to @alexandruag?

I agree that we might want to use compiler barrier + read_unaligned instead of read_volatile. However I think we should keep it as is and fix it later, to avoid delaying this pull request further.

src/mmap.rs
jiangliu: It's a special requirement to support virtio-fs, and we haven't considered such a use case before.

Let's first talk about the requirements of the virtio-fs driver. Per my understanding, to support virtio-fs devices, an MMIO window is created in guest's physical address space, and the virtiofsd dynamically  map/remap contents from files into the MMIO window. Is that correct?

If it's so, how about separating the virtio-fs logic from the vm-memory crate? We could do it in this way:
1) create a GuestRegionMmap object to allocate the virtual address and optionally anonymous memory for the MMIO window.
2) introduce a new VirtioFsManager to manage the virtual address allocated in step 1  as a pool and provides interfaces to
     a) allocate virtual address from the pool to map file content into the MMIO window.
     b) free virtual addresses into the pool when a file is deleted/closed.
     c) reclaim virtual address by LRU when the pool runs out of virtual address.
     d) map/remap file content to specific virtual address with mmap(MAP_FIXED).
3) the VirtioFsManager object out-lives the GuestRegionMmap object for the MMIO window.
By this way, we don't need to change the vm-memory crate and just introduce a new manager for virtio-fs.

As a bonus, we may make a minor improvement to the vm-memory crate to support mmap(PROT_NONE) in step 1, so that page fault happens when guest accesses a virtual address in the MMIO window which hasn't been assoicated with any file content yet.

src/guest_memory.rs
```diff
@@ -0,0 +1,237 @@
...
+///
+/// Notes:
+/// - On ARM64, a 32-bit hypervisor may be used to support a 64-bit guest. For simplicity,
+/// u64 is used to store the the raw value no matter the guest a 32-bit or 64-bit virtual machine.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub struct GuestAddress(pub u64);
+impl_address_ops!(GuestAddress, u64);
+
+/// Type of the raw value stored in a GuestAddress object.
+pub type GuestAddressValue = <GuestAddress as AddressValue>::V;
```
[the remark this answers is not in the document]
jiangliu: Seems that rust prefers associated types against generic parameters, because it needs more typing when using a trait with generic parameter and sometimes the code with generic parameters looks ugly.

src/guest_memory.rs
```diff
@@ -0,0 +1,445 @@
...
+//! Traits to track and access guest's physical memory.
+//!
+//! To make the abstraction as generic as possible, all the core traits only define methods to
+//! access the address space are defined here, and they never define methods to manage (create,
+//! delete, insert, remove etc) address spaces. By this way, the address space consumers
+//! (virtio device drivers, vhost drivers and boot loaders etc) may be decoupled from the address
+//! space provider (typically a hypervisor).
+//!
+//! Traits and Structs
+//! - GuestAddress: represents a guest physical address (GPA).
```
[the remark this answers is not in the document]
jiangliu: Guest suggestions:)

src/guest_memory.rs
alexandruag: Thanks for the added context! There are obviously a lot of things that I'm missing, but it's not super clear that one behavior is correct/desirable every single time, whereas the other shouldn't ever make sense for anyone :( 

How about decoupling any particular behavior from the interface and leaving it up the implementation? In this case, we remove the `try_access` entirely from `GuestMemory`, and leave every method that relies on it unimplemented. We can then use `try_access` in the reference implementation, where we might also (at some point) add a parameter/feature/compilation flag that configures whether cross-region access is allowed.

alexandruag:
Hello, sorry for not being able to attend the meeting today.

>Do you have some specific 2018 features in mind? Even ? technically is one.

I don't fully grasp all the enhancements brought by the 2018 edition, but I really like not having to explicitly `extern crate ...` for imported crates and the much nicer path resolution system when `use`-ing/referencing stuff. Some additional features might also start growing on me/other ppl as we find out about them. The downside is, as Andreea mentioned, that older toolchains are no longer usable, but it should become less of a problem as time goes by. Also, the fact that so many distros provide toolchain versions that old is a bit sad.

>Unfortunately it's not possible due to orphan trait rules:
>impl<T: GuestMemory> Bytes<GuestAddress> for T {
>^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type parameter T must be used as the type parameter for some local type

I though the orphan rule does not apply when splitting things into separate modules (within the same crate), but anyway that comment was essentially about refactoring, which can be done at a later time.

src/guest_memory.rs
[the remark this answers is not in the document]
alexandruag: Ok, maybe we can pick up this discussion in another PR, since there are some nuances here, if ppl aren't terribly annoyed by it already :D

jiangliu:
Paolo has mentioned a good point that virtual machines mimic physical machines:)
For physical machines, a request crossing memory region boundary is allowed for several cases, such as
1) the memory hot plug case we have mentioned,
2) a request crossing NUMA node boundary,
3) even a request crossing physical memory and MMIO boundary is valid.

alexandruag:
> Paolo has mentioned a good point that virtual machines mimic physical machines:) For physical machines, a request crossing memory region boundary is allowed for several cases, such as 1) the memory hot plug case we have mentioned, 2) a request crossing NUMA node boundary, 3) even a request crossing physical memory and MMIO boundary is valid.

I guess what I wanna ask is this: are there any downsides to having an interface (the `GuestMemory` trait) that is agnostic in terms of cross-region memory access, especially since users have full control over which implementation they choose?

jiangliu:
On Mar 21, 2019, at 12:32 AM, Alexandru Agache <notifications@github.com> wrote:
> 
> Paolo has mentioned a good point that virtual machines mimic physical machines:) For physical machines, a request crossing memory region boundary is allowed for several cases, such as 1) the memory hot plug case we have mentioned, 2) a request crossing NUMA node boundary, 3) even a request crossing physical memory and MMIO boundary is valid.
> 
> I guess what I wanna ask is this: are there any downsides to having an interface (the GuestMemory trait) that is agnostic in terms of cross-region memory access, especially since users have full control over which implementation they choose?
> 
How about reimplementing the try_access() method for a specific GuestMemory implementation if the default implementation isn’t desired? 
Seems that may solve the concern:)

bonzini:
> I guess what I wanna ask is this: are there any downsides to having an interface (the GuestMemory trait) that is agnostic in terms of cross-region memory access, especially since users have full control over which implementation they choose?

The advantage is in code reuse.

src/guest_memory.rs
```diff
@@ -96,14 +99,59 @@ impl Display for Error {
...
+            return Err(Error::InvalidGuestAddress(addr));
+        }
+        Ok(MemoryRegionAddress(offset))
+    }
+
+
+    /// Return a slice corresponding to the data in the region; unsafe because of
+    /// possible aliasing.  Return None if the region does not support slice-based
+    /// access.
+    unsafe fn as_slice(&self) -> Option<&[u8]> {
```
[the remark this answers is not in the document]
jiangliu: Currently they are used by Bytes<GuestAddress>::read_into_stream, write_from_stream}.

alexandruag:
> The advantage is in code reuse.

It doesn't look like the argument of code reuse is significant in this context. `try_access` is essentially a helper method. It brings convenience, but does not enrich the `GuestMemory` trait in terms of expressiveness (that is another reason why I consider it should not be a part of `GuestMemory` itself). If its code reuse value is high, we can do the following:
- Alter the definition of `GuestMemory { ... }` to `GuestMemory: Bytes<GuestAddress> { ... }`.
- Move `try_access` (and the implementations it enables) to another trait, say `pub trait PartiallyImplementedGuestMemory: GuestMemory`. Anyone willing to reuse the code can derive from the new trait.

Getting back to the *to cross or not to cross thing* for a bit, I acknowledge the multitude of use-cases which require cross-region semantics, but please consider these points:

- At the end of the day, the `GuestMemory` trait cannot enforce any particular semantics in this regard; an implementation is free to override any method, including `try_access`. This means users have to be mindful of the implementation anyway.

- `GuestMemory` (or anything else for that matter) does not ascribe any precise meaning to `GuestMemoryRegion`. All regions might very well be the same for most implementations, but nothing stops others from differentiating between adjacent regions (among other things).

To sum it up, having something like `try_access` as part of `GuestMemory` feels wrong because not only are we conflating abstraction with implementation, but we're also doing it based on what I still consider to be an overgeneralization. Apologies for the long rant :D

src/mmap.rs
mcastelino: @jiangliu I am not sure if this requirement is limited to virtio-fs. Will we have a need for something similar when managing the PCI Hole and BAR's within the PCI Hole. @bonzini may have a better idea. Will the reqions corresponding to the BARs be allocated on the fly. If the regions corresponding to the BARs are allocated on the fly, then the current implementation of `GuestMemoryMmap` should probably need to support adding regions.

sboeuf: @jiangliu 
>Let's first talk about the requirements of the virtio-fs driver. Per my understanding, to support virtio-fs devices, an MMIO window is created in guest's physical address space, and the virtiofsd dynamically map/remap contents from files into the MMIO window. Is that correct?

Yes this is correct.

Introducing a `VirtioFsManager` seems like a lot to address a problem that is more generic as we could end up doing multiple `mmap` in any situation that requires submappings withing an existing primary mapping.

With your solution, you would create a new entry holding onto an `MmapRegion` every time a new submapping happens. And at the end, when all those entries go out of scope, they get dropped with `Drop`, which means a bunch of `munmap()` are being called. This works but seems more complex than the solution proposed by @mcastelino.

Eventually, I think we would want to be able to define that a `MmapRegion` holds a list of submapping that could be implementing a different `Drop` (doing an `mmap` instead).

Last thing, the mapping could be handled by a manager as you proposed, but if we end up going down this path, we should have this manager being defined as a trait that can be reused by any device. We cannot expect every device to handle this use case in its own way.

alexandruag:
Hi again! Taking a step back, it seems there are a couple of ongoing discussions, and also some comments yet to be answered. Would it make sense to merge the stack of commits from this PR, so we can then proceed with the various conversation topics on multiple PRs/issues? (and I guess comments can be answered even when they're on a merged PR). How do you ppl think we should proceed?

sboeuf:
@alexandruag 
>Hi again! Taking a step back, it seems there are a couple of ongoing discussions, and also some comments yet to be answered. Would it make sense to merge the stack of commits from this PR, so we can then proceed with the various conversation topics on multiple PRs/issues? (and I guess comments can be answered even when they're on a merged PR). How do you ppl think we should proceed?

If nobody has any objection about the current code, thinking that it might break future add-on and expectation of the larger memory model, then we can definitely go ahead and merge this so that follow up PRs can happen.

src/guest_memory.rs
bonzini: How is it useful _not_ to cross regions? IMO the point of this crate is exactly to provide a good implementation for things that are not so obvious, and this qualifies perfectly...

src/address.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// Returns the result of the subtraction and a flag identifying whether there was overflow
+    fn overflowing_sub(&self, other: Self::V) -> (Self, bool);
+
+    /// Returns the result of the subtraction.
+    /// Only use this when `other` is guaranteed not to underflow.
+    fn unchecked_sub(&self, other: Self::V) -> Self;
+}
+
+#[macro_export]
+macro_rules! impl_address_ops {
```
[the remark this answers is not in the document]
bonzini: I've just removed the export.

bonzini:
I am pushing the fixes for most of the reviews to my fork, and will channel it to @jiangliu via a pull request. Then we can approve and merge this.

bonzini:
> At the end of the day, the GuestMemory trait cannot enforce any particular semantics in this regard; an implementation is free to override any method, including try_access. This means users have to be mindful of the implementation anyway.

It doesn't, but then you can also implement `Add` so that it subtracts... It would just be a bug.

The case of differentiating between adjacent regions should be handled by `try_access` as well. If an implementation wants to differentiating between adjacent regions and finds `try_access` to lack in this respect, they should fix vm_memory instead of overriding it or worse vendoring it.

jiangliu:
Hi sameo,
	I have merged Paolo’s new patch set and clean up the commit history a bit.
And just send it to Paolo for review yesterday. So send out a new PR today after addressing a suggestion from Paolo.
Thanks!

src/mmap.rs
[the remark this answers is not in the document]
ericho: Ohh I see, thanks for the info :)

jiangliu:
Hi all,
        Thank you all for reviewing the PR. With patches to address the review comments, the commit history became a little messy. So I clean up the patch series by folding those trivial patches into other commits, the result patch set becomes much more cleaner.  A new PR is created to ease reviewing and merging. This PR will be closed once the new PR gets merged.
        https://github.com/rust-vmm/vm-memory/pull/4

src/guest_memory.rs
```diff
@@ -20,11 +20,14 @@
...
 use volatile_memory;
 use std::fmt::{self, Display};
-use std::io;
+use std::io::{self, Read, Write};
 use std::ops::{BitAnd, BitOr};
 use std::convert::From;
+
 use Bytes;
 
+static MAX_ACCESS_CHUNK: usize = 4096;
```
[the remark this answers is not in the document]
jiangliu: Guess it's based on typical physical page sizes.

src/mmap.rs
```diff
@@ -0,0 +1,1200 @@
...
+        self.addr
+    }
+
+    unsafe fn as_slice(&self) -> &[u8] {
+        // This is safe because we mapped the area at addr ourselves, so this slice will not
+        // overflow. However, it is possible to alias.
+        std::slice::from_raw_parts(self.addr, self.size)
+    }
+
+    unsafe fn as_mut_slice(&self) -> &mut [u8] {
```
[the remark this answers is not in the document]
jiangliu: The GuestMemoryRegion::as_mut_slice() is designed to support the case of relaying guest memory access to remote process by proxy. For example, the guest's memory is not mapped into the current process, and it relies on a proxy to access the guest's memory on a remote process.

sameo:
> Hi all,
> Thank you all for reviewing the PR. With patches to address the review comments, the commit history became a little messy. So I clean up the patch series by folding those trivial patches into other commits, the result patch set becomes much more cleaner. A new PR is created to ease reviewing and merging. This PR will be closed once the new PR gets merged.
> Please refer to #4 for the new PR.

Hi @jiangliu 
I personally prefer to keep it within the same PR, so that the complete history lives in the same PR thread. IMO it's ok to squash commits and force push into that PR. If people really want to look at the PR history, they can inspect your branch.
I'd suggest to do so (squash and force push) in that PR and make it exactly like PR #4, unless someone (@andreeaflorescu @bonzini @alexandruag @sboeuf ?) disagrees...

jiangliu:
Sure, I will force push to the PR and close PR #4.

bonzini:
I think there was a little misunderstanding about GuestUsize; it's `GuestAddressValue` that should be renamed instead. Can you squash in my commit [eaee4937](https://github.com/jiangliu/vm-memory/pull/2/commits/eaee4937eccd18cf365a7883eb4bbe59442a2511)?

(I tried fixing some of the volatile issues, but ran afoul of rust-lang/rust#43408).

src/volatile_memory.rs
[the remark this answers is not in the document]
bonzini: Fixed now via `#[repr(packed)]`.

src/volatile_memory.rs
```diff
@@ -0,0 +1,656 @@
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn copy_from<T>(&self, buf: &[T])
+    where
+        T: DataInit,
+    {
+        let mut addr = self.addr;
+        for &v in buf.iter().take(self.size / size_of::<T>()) {
+            unsafe {
```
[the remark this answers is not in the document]
bonzini: Fixed now via `#[repr(packed)]`.

sboeuf:
LGTM

Issue #6, opened 2019-03-26
Chainsawing of firecracker memory_model went a bit too far


bonzini:
A few useful methods were removed; we should copy them back from firecracker to ease porting.

PR #7, opened 2019-03-26
Add more methods

bonzini:
I have noticed some missing methods while trying to port Firecracker to rust-vmm/vm-memory. This pull request adds them.

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -50,8 +50,6 @@ pub enum Error {
     PartialBuffer { expected: usize, completed: usize },
     /// Requested backend address is out of range.
     InvalidBackendAddress,
-    /// Requested offset is out of range.
-    InvalidBackendOffset,
 }
 
 impl From<volatile_memory::Error> for Error {
@@ -93,7 +91,6 @@ impl Display for Error {
                 completed, expected,
             ),
             Error::InvalidBackendAddress => write!(f, "invalid backend address"),
-            Error::InvalidBackendOffset => write!(f, "invalid backend offset"),
         }
     }
 }
@@ -333,9 +330,6 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
             buf.len(),
             addr,
             |offset, _count, caddr, region| -> Result<usize> {
-                if offset >= buf.len() {
-                    return Err(Error::InvalidBackendOffset);
-                }
                 region.write(&buf[offset as usize..], caddr)
             },
         )
@@ -346,25 +340,13 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
             buf.len(),
             addr,
             |offset, _count, caddr, region| -> Result<usize> {
-                if offset >= buf.len() {
-                    return Err(Error::InvalidBackendOffset);
-                }
                 region.read(&mut buf[offset as usize..], caddr)
             },
         )
     }
 
     fn write_slice(&self, buf: &[u8], addr: GuestAddress) -> Result<()> {
-        let res = self.try_access(
-            buf.len(),
-            addr,
-            |offset, _count, caddr, region| -> Result<usize> {
-                if offset >= buf.len() {
-                    return Err(Error::InvalidBackendOffset);
-                }
-                region.write(&buf[offset as usize..], caddr)
-            },
-        )?;
+        let res = self.write(buf, addr)?;
         if res != buf.len() {
             return Err(Error::PartialBuffer {
                 expected: buf.len(),
@@ -375,16 +357,7 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
     }
 
     fn read_slice(&self, buf: &mut [u8], addr: GuestAddress) -> Result<()> {
-        let res = self.try_access(
-            buf.len(),
-            addr,
-            |offset, _count, caddr, region| -> Result<usize> {
-                if offset >= buf.len() {
-                    return Err(Error::InvalidBackendOffset);
-                }
-                region.read(&mut buf[offset as usize..], caddr)
-            },
-        )?;
+        let res = self.read(buf, addr)?;
         if res != buf.len() {
             return Err(Error::PartialBuffer {
                 expected: buf.len(),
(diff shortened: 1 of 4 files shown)


bonzini:
Fixes #6

src/guest_memory.rs
```diff
@@ -175,19 +181,89 @@ pub trait GuestMemory {
...
+    ///     println!("Total memory size = {} KB", total_size);
+    ///     Ok(())
+    /// # }
+    /// ```
+    fn map_and_fold<F, G, T>(&self, init: T, mapf: F, foldf: G) -> T
+    where
+        F: Fn((usize, &Self::R)) -> T,
+        G: Fn(T, T) -> T;
+
+    /// Returns true if the given address is within the memory range available to the guest.
```
alexandruag: Comment not really related to the function. Can also fix later (and was probably just copied like that anyway :D).

src/guest_memory.rs
```diff
@@ -125,6 +125,12 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
     /// Get minimum (inclusive) address managed by the region.
     fn min_addr(&self) -> GuestAddress;
 
+    /// Get maximum (inclusive) address managed by the region.
+    fn end_addr(&self) -> GuestAddress {
```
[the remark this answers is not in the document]
bonzini: Let's get rid of max_addr and do start_addr/end_addr.

src/mmap.rs
```diff
@@ -181,8 +182,20 @@ impl GuestRegionMmap {
...
     }
 
-    fn as_volatile_slice(&self) -> VolatileSlice {
-        self.mapping.as_volatile_slice()
+    /// Convert an absolute address into an address space (GuestMemory)
+    /// to a host pointer, or return None if it is out of bounds.
+    pub fn get_host_address(&self, addr: MemoryRegionAddress) -> Option<*mut u8> {
+        // Safe because we've just range-checked addr using check_address.
+        self.check_address(addr)
+            .map(|addr| unsafe { self.as_ptr().offset(addr.raw_value() as isize) })
```
alexandruag: On second thought, this looks potentially fishy for systems where `sizeof(isize) <= 32`, because we could have a legit memory region larger than `isize::MAX` bytes. Does this look like a real concern?

Otherwise, all good.

alexandruag: I meant `<= 4` :(

bonzini: Yes, it should use `wrapping_offset`. It is not even unsafe, though I'm not sure why...

src/mmap.rs
```diff
@@ -181,8 +182,21 @@ impl GuestRegionMmap {
...
 
-    fn as_volatile_slice(&self) -> VolatileSlice {
-        self.mapping.as_volatile_slice()
+    /// Convert an absolute address into an address space (GuestMemory)
+    /// to a host pointer, or return None if it is out of bounds.
+    pub fn get_host_address(&self, addr: MemoryRegionAddress) -> Option<*mut u8> {
+        // Not sure why wrapping_offset is not unsafe.  Anyway this
+        // is safe because we've just range-checked addr using check_address.
+        self.check_address(addr)
+            .map(|addr| self.as_ptr().wrapping_offset(addr.raw_value() as isize))
```
alexandruag: Hard to say why one is unsafe while the other isn't; they compile down to different intrinsics whose sources I couldn't find during a cursory search :( 

Anyway, apologies for not providing a more complete comment the first time: this still looks potentially wrong because of the `as isize` conversion. For example, in the 32 bit case, the reinterpretation `raw_value() as isize` might turn intro a negative number if the memory region and `addr` are large enough, and the result will be wrong (for 64 bit it's still theoretically possible, but kinda unrealistic). I guess we can just check against this condition as well.

bonzini: The important thing is that the `MemoryRegionAddress`'s u64 has been checked already to be in range of `self.len()`, so it must fit in 32 bits.

At this point, even if it's a negative number after the `isize` cast, it wraps around just fine.

alexandruag: Oh yeah, I forgot how these numbers actually work :( My bad

src/guest_memory.rs
```diff
@@ -440,10 +411,8 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
         F: Write,
     {
         self.try_access(count, addr, |offset, len, caddr, region| -> Result<usize> {
-            // Something bad happened...
-            if offset >= count {
-                return Err(Error::InvalidBackendOffset);
-            }
+            // Check if something bad happened before doing unsafe things.
```
dagrh: That's aq bit of a fluffy comment!

src/guest_memory.rs
```diff
@@ -175,19 +196,89 @@ pub trait GuestMemory {
...
+    ///     let total_size = mem.map_and_fold(
+    ///         0,
+    ///         |(_, region)| region.len() / 1024,
+    ///         |acc, size| acc + size
+    ///     );
+    ///     println!("Total memory size = {} KB", total_size);
+    ///     Ok(())
+    /// # }
+    /// ```
+    fn map_and_fold<F, G, T>(&self, init: T, mapf: F, foldf: G) -> T
```
andreeaflorescu: Can't we have those as separate methods aka `map()` and `fold()`? Or not have them at all actually.

I never liked it in Firecracker either and was hoping to get rid of it at some point. We are using it exactly once in Firecracker to handle `KVM_GET_DIRTY_LOG`. It think we can do better.

bonzini: Hmm I wonder if `-> impl Iterator<Item = X>`  would let us implement `map`... I didn't like it very much either, even though I am actually using it here so that makes two...

alexandruag: Well, we could provide access to regions via iterators instead of (or in addition to) the `with_regions_*` methods, and then we get `map`, `fold`, and all the other iterator enabled goodies for free :D

bonzini: I tried adding regions_iter but he problem is that iterators have lifetime parameters... I think that would also apply to map() but I am not sure. I think we can improve it later.

andreeaflorescu: @bonzini do you mind adding an issue about this one? Just to keep it mind for future refactoring :D

PR #9, opened 2019-03-31
Add GuestMemory method to return an Iterator

### LINKED ISSUE #8: Add GuestMemory method to return an Iterator
Currently we have methods such as `with_regions`, `with_regions_mut` and `map_and_fold` that are just papering over the lack of iterator-based access to the regions in GuestMemory. If possible we should try to add a better and more powerful abstraction.

jiangliu:
Fix #8

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -193,7 +193,7 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
 /// - handle cases where an access request spanning two or more GuestMemoryRegion objects.
 ///
 /// Note: all regions in a GuestMemory object must not intersect with each other.
-pub trait GuestMemory<'a> {
+pub trait GuestMemory<'a>: Clone {
     /// True if the guest memory is mapped into the current process and supports direct access.
     const DIRECT_ACCESS: bool = Self::R::DIRECT_ACCESS;
 
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -201,63 +201,8 @@ pub trait GuestMemory<'a> {
     /// Gets an iterator over the entries in the collection.
     fn iter(&'a self) -> Self::I;
 
-    /// Perform the specified action on each region.
-    /// It only walks children of current region and do not step into sub regions.
-    fn with_regions<F, E>(&self, cb: F) -> std::result::Result<(), E>
-    where
-        F: Fn(usize, &Self::R) -> std::result::Result<(), E>;
-
-    /// Perform the specified action on each region mutably.
-    /// It only walks children of current region and do not step into sub regions.
-    fn with_regions_mut<F, E>(&self, cb: F) -> std::result::Result<(), E>
-    where
-        F: FnMut(usize, &Self::R) -> std::result::Result<(), E>;
-
-    /// Applies two functions, specified as callbacks, on the inner memory regions.
-    ///
-    /// # Arguments
-    /// * `init` - Starting value of the accumulator for the `foldf` function.
-    /// * `mapf` - "Map" function, applied to all the inner memory regions. It returns an array of
-    ///            the same size as the memory regions array, containing the function's results
-    ///            for each region.
-    /// * `foldf` - "Fold" function, applied to the array returned by `mapf`. It acts as an
-    ///             operator, applying itself to the `init` value and to each subsequent elemnent
-    ///             in the array returned by `mapf`.
-    ///
-    /// # Examples
-    ///
-    /// * Compute the total size of all memory mappings in KB by iterating over the memory regions
-    ///   and dividing their sizes to 1024, then summing up the values in an accumulator.
-    ///
-    /// ```
-    /// # #[cfg(feature = "backend-mmap")]
-    /// # fn test_map_fold() -> Result<(), ()> {
-    /// # use vm_memory::{GuestAddress, GuestMemory, GuestMemoryRegion, mmap::GuestMemoryMmap};
-    ///     let start_addr1 = GuestAddress(0x0);
-    ///     let start_addr2 = GuestAddress(0x400);
-    ///     let mem = GuestMemoryMmap::new(&vec![(start_addr1, 1024), (start_addr2, 2048)]).unwrap();
-    ///     let total_size = mem.map_and_fold(
-    ///         0,
-    ///         |(_, region)| region.len() / 1024,
-    ///         |acc, size| acc + size
-    ///     );
-    ///     println!("Total memory size = {} KB", total_size);
-    ///     Ok(())
-    /// # }
-    /// ```
-    fn map_and_fold<F, G, T>(&self, init: T, mapf: F, foldf: G) -> T
-    where
-        F: Fn((usize, &Self::R)) -> T,
-        G: Fn(T, T) -> T;
-
     /// Get maximum (inclusive) address managed by the region.
-    fn end_addr(&self) -> GuestAddress {
-        self.map_and_fold(
-            GuestAddress(0),
-            |(_, region)| region.end_addr(),
-            std::cmp::max,
-        )
-    }
+    fn end_addr(&self) -> GuestAddress;
 
     /// Convert an absolute address into an address space (GuestMemory)
     /// to a relative address within this region, or return None if
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -116,6 +116,9 @@ pub type GuestUsize = <GuestAddress as AddressValue>::V;
 /// Represents a continuous region of guest physical memory.
 #[allow(clippy::len_without_is_empty)]
 pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
+    /// True if the guest memory is mapped into the current process and supports direct access.
+    const DIRECT_ACCESS: bool;
+
     /// Get the size of the region.
     fn len(&self) -> GuestUsize;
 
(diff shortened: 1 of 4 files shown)


src/guest_memory.rs
```diff
@@ -166,16 +166,22 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
 /// - handle cases where an access request spanning two or more GuestMemoryRegion objects.
 ///
 /// Note: all regions in a GuestMemory object must not intersect with each other.
-pub trait GuestMemory {
-    /// Type of objects hosted by the address space.
-    type R: GuestMemoryRegion;
+pub trait GuestMemory<'a> {
```
alexandruag: It's so sad that `impl Trait` doesn't work with trait methods yet, because then we wouldn't have needed the lifetime parameter for `GuestMemory` (or so it seems). Do you ppl think it's significantly awkward/cumbersome to have `GuestMemory<'a>` instead of simply `<GuestMemory>`? I think there might be a way to shift the parameter over to the implementation (but it still won't be beautiful).

jiangliu: It's caused me a whole day to work out this solution:(
And any better solutions are welcomed so I could get better understanding of rust:)

bonzini: I think we should hold off this pull request until we get some more experience migrating Firecracker to vm-memory (I have some initial code but nothing that I tested seriously or attempted to split into sensible commits). I'd rather avoid the risk of having an API that's not very usable; in particular if anything that has a `&GuestMemory` in Firecracker needs to become generic over the lifetime, it'd be a huge pain to do the conversion...

jiangliu: Good suggestion, will give a try to migrate firecracker to the vm-memory crate:)

jiangliu:
Just a cleanup, so I will remove it from the patch set.

src/guest_memory.rs
alexandruag: I wrote a quick and somewhat dirty way of removing the generic lifetime parameter from `GuestMemory`, which is now associated with the concrete type implementing it. However, as written in the commit message, I think a lot of the awkwardness associated with passing lifetimes around can be avoided in this case by using `impl GuestMemory` instead of referring to the actual type. Anyway, if you wanna take a look, the changes are in the `nolife` branch from my fork of `vm-memory`. The commit is added on top of the latest version of Gerry's code.

alexandruag: Oh snap, I forgot the `mmap` feature is conditionally compiled so I didn't test things properly :( ... pls disregard my previous message until I figure out if what I wanted to do can actually work :(

bonzini:
Nice! Can you add some comments explaining the reason for the PhantomData?

src/guest_memory.rs
bonzini: Ahah it did seem a bit unlikely that it would compile, but I didn't trust my Rust knowledge too much...

alexandruag: Ok so I found a bit more time to look into this. As long as we use associated types for the `Iterator`/`Item`, that forces us to fully declare the type as part of the implementing struct, which will require a lifetime that cannot be filled in via elision, so it has to be a parameter of either the trait or its implementation. Moreover, if we associate it with the implementation, there seems to be no way of reconciling the `iter` method from over there with the one in the trait, because the trait uses an anonymous lifetime (via `&self`) in its declaration, whereas the implementation has to link to the explicitly provided one :(

Something like the existence of an implicit `'self` lifetime (associated with the current object) would have helped, but it's not supported. Returning an `impl Trait` from `iter` would also help, but it's also not supported (yet) for traits.

That being said, there are still a couple of options that come to mind in order to avoid the lifetime parameter on `GuestMemory`:
- Return a `Box<dyn Iterator<Item=&Self::R>>` from `GuestMemory::iter`. However, this involves both heap allocation and dynamic dispatch, which might very well be a concern.
- If it's reasonable to assume that indexing over the regions in a `GuestMemory` is an efficient operation, we can add a `region(index: usize)` method to `GuestMemory` and then define an object that acts as an iterator over any kind of type implementing `GuestMemory` along the lines of
```rust
struct GuestMemoryIterator<'a, T> {
    inner: &'a T,
    next_position: usize,
}

impl<'a, T: GuestMemory> Iterator for GuestMemoryIterator<'a, T> {
    type Item = &'a T::R;
    
    fn next(&mut self) -> Option<Self::Item> {
        let res = self.inner.region(next_position)
        self.next_position += 1;
        res
    }   
}
```
If indexing is not necessarily supposed to be efficient, then we could consider some other helper function that provides a reference to the next region based on more information than just the current position. With `GuestMemoryIterator` in place, we can write `GuestMemory::iter` as
```rust 
fn iter(&self) -> GuestMemoryIterator<Self> {
    GuestMemoryIterator {
        inner: &self,
        next_position: 0,
    }
}
```
I can try and actually implement this if the assumptions make sense. Anyway, that's as much as I want to think about lifetimes for today :(

jiangliu: I have tried this way but given up:(
If the GuestMemory object is implemented by array-like data structures, it's perfect because we could use a simple index to implement an iterator. But it gets harder if the GuestMemory object is implemented by tree-like data structures.

alexandruag: Well, at least we tried. I'll switch my efforts towards seeing how much of a nuisance that lifetime parameter really is (maybe elision takes out most of the awkwardness). Also, I'll start looking into the implications of switching Firecracker to the new `GuestMemory` paradigm, with traits and maybe even lifetimes and all that.

jiangliu: Seems the right solution to remove the lifetime parameter should be "associated type constructors". 
FYI:
https://github.com/rust-lang/rfcs/blob/master/text/1598-generic_associated_types.md
http://smallcultfollowing.com/babysteps/blog/2016/11/02/associated-type-constructors-part-1-basic-concepts-and-introduction/
http://smallcultfollowing.com/babysteps/blog/2016/11/04/associated-type-constructors-part-3-what-higher-kinded-types-might-look-like/

But "associated type constructors" is still a nightly feature, we can't rely on it yet:(
So how about adopting the solution proposed by @alexandruag as a temporary solution to remove lifetime parameter and switching to associated type constructors when it's stable?
The switch from the temporary solution to associated type constructors should almost be transparent to GuestMemory consumers and only minor changes to GuestMemory providers.

alexandruag: Hi Gerry, which temporary solution are you referring to? The one from the commit link I posted doesn't work (and will take some time until the shame wears off), in case it wasn't obvious :D When I get a bit of time, I'll have a look on how awkward using that lifetime parameter for `GuestMemory` really is.

bonzini:
Hey people, I placed my initial attempt at converting firecracker at https://github.com/bonzini/firecracker branch master.

A more sensible way to do it would be to split like:
* Rename memory-model classes to the vm-memory equivalents
* Rename memory-model methods to the vm-memory (especially Bytes) equivalents
* Add "type GuestUsize = u32;" to memory-model, fix compilation
* Finally switch to vm-memory and add the new traits `use`s.

The vm-memory is not properly vendored, it's hacked to add the mmap feature to Cargo.toml. And so on... still if it saves you a couple hours, all for the best.

andreeaflorescu:
Hey @bonzini, thanks. I will open an issue in the Firecracker repository to switch to vm-memory and link the commit from your fork.

We will wait for the vm-memory to be published to crates.io before using it. Before doing the switch we want to make sure we have good code coverage and the necessary mechanisms in place to ensure that future PRs won't decrease it. I will open issues in vm-memory for the other things that I'll find.

jiangliu:
I have tried to migrate the vhost crate in firecracker to the vm-memory crate, it works with one more patch upon this PR.

andreeaflorescu:
@jiangliu we are currently trying to move away from vhost. See this RFC with the design doc: https://github.com/firecracker-microvm/firecracker/pull/1044

jiangliu:
Thanks @andreeaflorescu , happy to see the progress on support of vsock. I will keep current firecracker's vsock implementation as an example:)

andreeaflorescu:
@jiangliu what is the status here? Do we still want to merge this one?

src/mmap.rs
```diff
@@ -376,6 +367,8 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
 }
 
 impl GuestMemoryRegion for GuestRegionMmap {
+    const DIRECT_ACCESS: bool = true;
```
bonzini: Why is DIRECT_ACCESS useful?

src/guest_memory.rs
```diff
@@ -175,6 +178,11 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
     unsafe fn as_mut_slice(&self) -> Option<&mut [u8]> {
         None
     }
+
+    /// Convert an absolute address in an address space (GuestMemory) to a host pointer,
+    /// or return None if it is out of bounds. It always returns None when `DIRECT_ACCESS` is false.
+    /// The returned pointer may be used for ioctls etc.
+    fn get_host_address(&self, addr: MemoryRegionAddress) -> Option<*mut u8>;
```
bonzini: This can be implemented in terms of `as_mut_slice()`. Also please make the result `Option<NonNull<u8>>`.

jiangliu:
> @jiangliu what is the status here? Do we still want to merge this one?

Hi Paolo,
        The rust GAT(Generic Associated Type) feature is still unstable. Merging this patch will force all clients to carry on an extra lifetime parameter, which is a little heavy. So suggest to wait until that the GAT feature gets stable.

bonzini:
Agreed.

PR #10, opened 2019-04-03
Port the mmap backend to windows

tkreuzer:
This pull request implements the Windows version of the mmap-backend.

It is an initial version that probably contains some issues wrt to comments and formatting and there are still failing tests, but the main functionality works fine.

I suggest as an additional change to rename the *_raw_fd functions to _raw_file and add a RawFile trait on top of RawFd for unix and RawHandle for Windows. Please let me know what you think about this.

### DIFF
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -20,27 +20,27 @@
 //! - [GuestMemoryMmap](struct.GuestMemoryMmap.html): provides methods to access a collection of
 //! GuestRegionMmap objects.
 
-use libc;
 use std::io::{self, Read, Write};
 use std::ops::Deref;
-use std::os::unix::io::AsRawFd;
-use std::ptr::null_mut;
 use std::sync::Arc;
 
 use address::Address;
 use guest_memory::*;
-use volatile_memory::{self, compute_offset, VolatileMemory, VolatileSlice};
+use volatile_memory::VolatileMemory;
 use Bytes;
 
-/// A backend driver to access guest's physical memory by mmapping guest's memory into the current
-/// process.
-/// For a combination of 32-bit hypervisor and 64-bit virtual machine, only partial of guest's
-/// physical memory may be mapped into current process due to limited process virtual address
-/// space size.
-#[derive(Debug)]
-pub struct MmapRegion {
-    addr: *mut u8,
-    size: usize,
+#[cfg(unix)]
+pub use mmap_unix::MmapRegion;
+
+#[cfg(windows)]
+pub use mmap_windows::MmapRegion;
+
+// For MmapRegion
+pub(crate) trait AsSlice {
+    unsafe fn as_slice(&self) -> &[u8];
+
+    #[allow(clippy::mut_from_ref)]
+    unsafe fn as_mut_slice(&self) -> &mut [u8];
 }
 
 /// Errors that can happen when creating a memory map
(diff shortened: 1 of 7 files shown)


andreeaflorescu:
I haven't looked at the code yet, but I think we should also add windows hosts to test this crate as future PRs might break the windows build. It seems like Travis also has [support for Windows](https://blog.travis-ci.com/2018-10-11-windows-early-release), but I have no idea how to enable those builds. Anyone has knowledge about this?

src/mmap.rs
```diff
@@ -507,18 +410,37 @@ mod tests {
...
 
     #[test]
+    #[cfg(unix)]
     fn map_invalid_size() {
         let e = MmapRegion::new(0).unwrap_err();
         assert_eq!(e.raw_os_error(), Some(libc::EINVAL));
     }
 
     #[test]
+    #[cfg(windows)]
```
jiangliu: Could we move these test cases into mmap_unix and mmap_windows? So we could remove many condition compilations.

src/mmap_windows.rs
```diff
@@ -0,0 +1,220 @@
+// Copyright (C) 2019 Alibaba Cloud Computing. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Portions Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Portions Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the THIRD-PARTY file.
```
jiangliu: Could you please help to replace above copyright claims with yours? We haven't contributed to it yet:)

src/mmap_unix.rs
```diff
@@ -0,0 +1,161 @@
...
+        self.addr
+    }
+}
+
+pub trait AsSlice {
+    unsafe fn as_slice(&self) -> &[u8];
+
+    #[allow(clippy::mut_from_ref)]
+    unsafe fn as_mut_slice(&self) -> &mut [u8];
+}
```
jiangliu: How about moving the AsSlice trait into file mmap.rs? So mmap.rs file defines and makes use of the AsSlice trait, and mmap_unix.rs/mmap_window.rs implements AsSlice for MmapRegion.
index f5cee36..9f40c8e 100644
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -29,16 +29,19 @@ use guest_memory::*;
 use volatile_memory::VolatileMemory;
 use Bytes;
 
-#[cfg(unix)]
-use mmap_unix::AsSlice;
 #[cfg(unix)]
 pub use mmap_unix::MmapRegion;
 
-#[cfg(windows)]
-use mmap_windows::AsSlice;
 #[cfg(windows)]
 pub use mmap_windows::MmapRegion;
 
+pub(crate) trait AsSlice {
+    unsafe fn as_slice(&self) -> &[u8];
+
+    #[allow(clippy::mut_from_ref)]
+    unsafe fn as_mut_slice(&self) -> &mut [u8];
+}
+
 /// Errors that can happen when creating a memory map
 #[derive(Debug)]
 pub enum MmapError {

tkreuzer: Done

tkreuzer:
@sboeuf Done. I intentionally didn't squash the commits, since it tends to make reviews harder by trashing the existing comments and squashing can always be done when merging. Anyway, I squashed it now.

petrutlucian94:
For the record, here are test results for this patch (run against Windows): https://buildkite.com/lpetrut/vm-memory-ci/builds/53

We'll have those test jobs integrated into the official CI soon.

tkreuzer:
@sboeuf Done.

Issue #11, opened 2019-04-04
MmapError is not exported


sameo:
The mmap API returns `MmapError` but it's not exported.
For callers to propagate the right errors from the vm-memory crate, this needs to be exported.

PR #12, opened 2019-04-04
mmap: Export MmapError

sameo:
mmap callers need this to properly propagate the memory model errors.

Fixes: #11

Signed-off-by: Samuel Ortiz <sameo@linux.intel.com>


sameo:
@bonzini @jiangliu PTAL

PR #19, opened 2019-05-16
volatile_memory: add VolatileArrayRef and &std::sync::atomic::Atomic* accessors

bonzini:
Allow easily access to an array stored in a VolatileMemory object, removing
the need to do multiplications and pointer offsetting.

### DIFF
--- a/src/volatile_memory.rs
+++ b/src/volatile_memory.rs
@@ -25,7 +25,7 @@ use std::cmp::min;
 use std::fmt;
 use std::io::{self, Read, Write};
 use std::marker::PhantomData;
-use std::mem::size_of;
+use std::mem::{align_of, size_of};
 use std::ptr::copy;
 use std::ptr::{read_volatile, write_volatile};
 use std::result;
@@ -47,6 +47,8 @@ pub enum Error {
     Overflow { base: usize, offset: usize },
     /// Taking a slice whose size overflows `usize`.
     TooBig { nelements: usize, size: usize },
+    /// Trying to obtain a misaligned reference.
+    Misaligned { addr: usize, alignment: usize },
     /// Writing to memory failed
     IOError(io::Error),
     /// Incomplete read or write
@@ -67,6 +69,9 @@ impl fmt::Display for Error {
                 "{:?} elements of size {:?} would overflow a usize",
                 nelements, size
             ),
+            Error::Misaligned { addr, alignment } => {
+                write!(f, "address 0x{:x} is not aligned to {:?}", addr, alignment)
+            }
             Error::IOError(error) => write!(f, "{}", error),
             Error::PartialBuffer {
                 expected,
@@ -105,6 +110,33 @@ pub fn compute_offset(base: usize, offset: usize) -> Result<usize> {
         Some(m) => Ok(m),
     }
 }
+
+/// Trait for objects for which a reference can be extracted safely out of a
+/// VolatileSlice.  Objects that implement this trait must consist exclusively
+/// of atomic types from std::sync::atomic, except for AtomicPtr<T>.
+pub unsafe trait AtomicValued: Sync + Send {}
+
+// also conditionalize on #[cfg(target_has_atomic) when it is stabilized
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicBool {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicI8 {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicI16 {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicI32 {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicI64 {}
+unsafe impl AtomicValued for std::sync::atomic::AtomicIsize {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicU8 {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicU16 {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicU32 {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicU64 {}
+unsafe impl AtomicValued for std::sync::atomic::AtomicUsize {}
 
 /// Trait for types that support raw volatile access to their data.
 pub trait VolatileMemory {
@@ -150,6 +182,41 @@ pub trait VolatileMemory {
             // This is safe because the pointer is range-checked by get_slice, and
             // the lifetime is the same as self.
             Ok(VolatileArrayRef::<T>::new(slice.addr, n))
+        }
+    }
+
+    /// Gets a reference to T at `offset`.  The resulting pointer must be
+    /// aligned, or the function fails.
+    ///
+    /// To use this safely, the caller must guarantee that there are no other
+    /// users of the given chunk of memory for the lifetime of the result.
+    unsafe fn aligned_as_ref<T: ByteValued>(&self, offset: usize) -> Result<&T> {
+        let slice = self.get_slice(offset, size_of::<T>())?;
+        slice.check_alignment(align_of::<T>())?;
+        Ok(&*(slice.addr as *const T))
+    }
+
+    /// Gets a reference to T at `offset`.  The resulting pointer must be
+    /// aligned, or the function fails.
+    ///
+    /// To use this safely, the caller must guarantee that there are no other
+    /// users of the given chunk of memory for the lifetime of the result.
+    unsafe fn aligned_as_mut<T: ByteValued>(&self, offset: usize) -> Result<&mut T> {
+        let slice = self.get_slice(offset, size_of::<T>())?;
+        slice.check_alignment(align_of::<T>())?;
+        Ok(&mut *(slice.addr as *mut T))
+    }
+
+    /// Gets a reference to T at `offset`.  The resulting pointer must be
+    /// aligned, or the function fails.
+    fn get_atomic_ref<T: AtomicValued>(&self, offset: usize) -> Result<&T> {
+        let slice = self.get_slice(offset, size_of::<T>())?;
+        slice.check_alignment(align_of::<T>())?;
+
+        unsafe {
+            // This is safe because the pointer is range-checked by get_slice, and
+            // the lifetime is the same as self.
+            Ok(&*(slice.addr as *const T))
         }
     }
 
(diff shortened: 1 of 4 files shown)


bonzini:
That's an interesting idea. I will take a look.

bonzini:
The main obstacle here is that `VolatileSlice` has some extra methods to copy in and out of the slice, compared to `VolatileArrayRef`, and I don't think they make sense for `VolatileArrayRef` (better: they do make sense, but with a different type bound, they can be `T: ByteValued` for VolatileSlice but they should not be generic for `VolatileArrayRef`. However, I've added a From implementation to go from `VolatileArrayRef<u8>` to `VolatileSlice`, and a bunch of other improvements.

src/volatile_memory.rs
```diff
@@ -671,6 +679,223 @@ impl<'a, T: ByteValued> VolatileRef<'a, T> {
...
+        unsafe { VolatileSlice::new(self.addr, self.nelem * self.element_size()) }
+    }
+
+    /// Does a volatile read of the I-th value pointed to by this ref.
+    pub fn ref_at(&self, index: usize) -> VolatileRef<'a, T> {
+        // Safe because the memory has the same lifetime and points to a subset of the
+        // memory of the VolatileArrayRef.
+        unsafe {
+            let byteofs = (self.element_size() * index) as isize;
+            let ptr = self.as_ptr().offset(byteofs);
```
alexandruag: I still don't understand how `ptr::offset` and `ptr::wrapping_offset` work/differ from each other :(, but the former does state *the computed offset, in bytes, cannot overflow an isize*, otherwise undefined behavior might happen. Is that a concern here?

src/volatile_memory.rs
```diff
@@ -671,6 +679,223 @@ impl<'a, T: ByteValued> VolatileRef<'a, T> {
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn copy_to(&self, buf: &mut [T]) -> usize {
+        let mut addr = self.addr;
+        let mut i = 0;
+        for v in buf.iter_mut().take(self.len()) {
+            unsafe {
+                *v = read_volatile(addr as *const Packed<T>).0;
+                addr = addr.add(self.element_size());
```
alexandruag: Seems like `ptr::add` also states undefined behavior might happen if the result overflows an `isize`.

src/volatile_memory.rs
```diff
@@ -671,6 +679,223 @@ impl<'a, T: ByteValued> VolatileRef<'a, T> {
...
+    /// # }
+    /// ```
+    pub fn copy_from(&self, buf: &[T]) {
+        let mut addr = self.addr;
+        for &v in buf.iter().take(self.len()) {
+            unsafe {
+                // Safe because the pointers are range-checked when the slices
+                // are created, and they never escape the VolatileSlices.
+                write_volatile(addr as *mut Packed<T>, Packed::<T>(v));
+                addr = addr.add(self.element_size());
```
alexandruag: Again with the `ptr::add` and overflowing an `isize` :-s

src/volatile_memory.rs
```diff
@@ -103,6 +108,33 @@ pub fn compute_offset(base: usize, offset: usize) -> Result<usize> {
     }
 }
 
+/// Trait for objects for which a reference can be extracted safely out of a
+/// VolatileSlice.  Objects that implement this trait must consist exclusively
+/// of atomic types from std::sync::atomic, except for AtomicPtr<T>.
+pub unsafe trait AtomicValued: Sync + Send {}
```
alexandruag: I imagine `AtomicValued` mainly functions as a marker trait, but would it be useful to have `AtomicValued: ByteValued` instead of only `Sync + Send`? I guess that way we can use all those functions with `T: ByteValued` trait bounds.

src/volatile_memory.rs
```diff
@@ -146,6 +178,61 @@ pub trait VolatileMemory {
...
+                alignment: align_of::<T>(),
+            })
+        } else {
+            Ok(&mut *(slice.addr as *mut T))
+        }
+    }
+
+    /// Gets a reference to T at `offset`.  The resulting pointer must be
+    /// aligned, or the function fails.
+    fn get_atomic_ref<T: AtomicValued>(&self, offset: usize) -> Result<&T> {
```
alexandruag: I was wondering, isn't atomicity orthogonal to pointer aliasing? If so, shouldn't this function be marked as `unsafe`, with the same contract as the previous two (*to use this safely, the caller must guarantee that there are no other users of the given chunk of memory for the lifetime of the result*)?  Moreover, in that case, does this become equivalent to `aligned_as_ref` if `AtomicValued: ByteValued`?

src/volatile_memory.rs
```diff
@@ -146,6 +178,61 @@ pub trait VolatileMemory {
         }
     }
 
+    /// Gets a reference to T at `offset`.  The resulting pointer must be
+    /// aligned, or the function fails.
+    ///
+    /// To use this safely, the caller must guarantee that there are no other
+    /// users of the given chunk of memory for the lifetime of the result.
+    unsafe fn aligned_as_ref<T: ByteValued>(&self, offset: usize) -> Result<&T> {
```
alexandruag: The body of these functions seems pretty boilerplate :-s Can we extract it into a helper method/macro and reuse that instead?

src/volatile_memory.rs
bonzini: I'll add a check against `isize` in `get_array_ref`. And especially, we need to range-check here!

src/volatile_memory.rs
bonzini: Here there is an implicit check against `self.len()` (thanks to `.take(self.len())`) so it's enough to check that the size fits an `isize` in get_array_ref.

src/volatile_memory.rs
bonzini: Here there is an implicit check against self.len() (thanks to .take(self.len())) so it's enough to check that the size fits an isize in get_array_ref.

src/volatile_memory.rs
bonzini: Unfortuanately that's not possible because `std::sync::atomic::Atomic*` types are not `Copy`, which `ByteValued` requires.

bonzini: But I think it's better this way, as it enforces that `AtomicValued` types are accessed with the proper memory orderings, and not via `VolatileRef` and friends.

src/volatile_memory.rs
bonzini: Actually the point of atomic types is _exactly_ that they can be aliased safely.

Atomic types, if you ignore all the crazy parts with `std::sync::atomic::Ordering`, behave very similarly to an int type wrapped with a mutex, e.g. `AtomiicU8` -> `Mutex<u8>`. In particular, atomic types provide interior mutability just like `Mutex`, i.e. they can be accessed concurrently even in the presence of aliases. This can be done because concurrent accesses to atomic types do not create a [data race](https://doc.rust-lang.org/nomicon/races.html) (yes, I'm showing off by linking the nomicon but you don't have to read it if you trust my comparison with `Mutex`:)), 

"Not creating a data race" of course says nothing about correctness; it's only about safety. However that should be enough to answer your question..

src/volatile_memory.rs
bonzini: I added check_alignment.

src/volatile_memory.rs
alexandruag: Thanks for the explanation! I don't think linking to anything is showing off; it's great to be pointed towards more information. I was initially worried about interpreting the same piece of memory as both atomic using this method, and also as some non-atomic value in some other place in the code, but I now realized this would be a breakage of the unsafe contract for the non-atomic access so yeah looks fine. 

Last question: so you're basically saying that data races are the only reason why pointer aliasing can lead to undefined behavior? (I promise I'll have a much better grasp of memory models some day)

bonzini: > also as some non-atomic value in some other place in the code, but I now realized this would be a breakage of the unsafe contract for the non-atomic access so yeah looks fine.

Actually even if you did, safety would be guaranteed by the contract of the `ByteValued` trait together with usage of `read_volatile`/`write_volatile`, but I agree that "don't do it" is the best thing to do with respect to mixing atomic and non-atomic accesses.

> so you're basically saying that data races are the only reason why pointer aliasing can lead to undefined behavior

Yes, pretty much! Again, this is only talking about undefined behavior, correctness is of course a whole other level of complexity: data races is why aliasing `Rc<T>` leads to undefined behavior, but there's more to `Arc<T>` than just making the reference count atomic.

src/volatile_memory.rs
```diff
@@ -632,7 +726,7 @@ impl<'a, T: ByteValued> VolatileRef<'a, T> {
     }
```
jiangliu: Should we check that the pointer satisfy the alignment requirement of T? read_volatile/write_volatile() requires the that the pointer is correctly aligned.

src/volatile_memory.rs
```diff
@@ -646,38 +740,264 @@ impl<'a, T: ByteValued> VolatileRef<'a, T> {
...
+
+impl<'a, T: ByteValued> VolatileArrayRef<'a, T> {
+    /// Creates a reference to raw memory that must support volatile access of `T` sized chunks.
+    ///
+    /// To use this safely, the caller must guarantee that the memory at `addr` is big enough for
+    /// `nelem` values of type `T` and is available for the duration of the lifetime of the new
+    /// `VolatileRef`. The caller must also guarantee that all other users of the given chunk of
+    /// memory are using volatile accesses.
+    pub unsafe fn new(addr: *mut u8, nelem: usize) -> VolatileArrayRef<'a, T> {
+        VolatileArrayRef {
```
jiangliu: Check alignment of addr too? Should we change the return value to Result<VolatileArrayRef<'a, T>>?

src/volatile_memory.rs
```diff
@@ -96,6 +111,33 @@ pub fn compute_offset(base: usize, offset: usize) -> Result<usize> {
...
+unsafe impl AtomicValued for std::sync::atomic::AtomicBool {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicI8 {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicI16 {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicI32 {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicI64 {}
+unsafe impl AtomicValued for std::sync::atomic::AtomicIsize {}
```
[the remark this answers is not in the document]
bonzini: No, `AtomicIsize` and `AtomicUsize` existed before other integer atomics were stabilized in 1.34.0; the feature is there to allow using vm-memory with older versions of Rust.

src/volatile_memory.rs
```diff
@@ -96,6 +111,33 @@ pub fn compute_offset(base: usize, offset: usize) -> Result<usize> {
...
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicU8 {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicU16 {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicU32 {}
+#[cfg(feature = "integer-atomics")]
+unsafe impl AtomicValued for std::sync::atomic::AtomicU64 {}
+unsafe impl AtomicValued for std::sync::atomic::AtomicUsize {}
```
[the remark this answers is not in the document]
bonzini: Same here, see https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html which lists 1.0.0 as the first version where it existed.

src/volatile_memory.rs
bonzini: No, the pointer is to `Packed<T>`, not `T`, and `Packed<T>` does not have alignment requirements because it's `#[repr(C, packed)]`.

src/volatile_memory.rs
bonzini: Same here, `VolatileArrayRef::ref_at` goes through `VolatileRef` which does not have alignment requirements.

PR #22, opened 2019-05-28
Add buildkite pipelines

petrutlucian94:
At the moment, we're using Travis for vm-memory tests and buildkite
for other crates.

This change adds a few buildkite pipeline definitions, which can
later on be moved to a common CI repository.

There are currently 4 pipelines, basically a matrix based on the
agent OS (Linux and Windows) and the feature list (default and
backend-mmap).

We're using the same workflow as the other CIs, going with the
Docker buildkite plugin and pointing to docker images that provide
Rust build tools.


petrutlucian94:
A few matters that we may want to figure out before merging this:

* Do we need separate buildkite jobs for arm or x86 for the vm-memory crate?
* Should we have additional agent tags so that we can just pick the os type (e.g. windows, linux, etc) and ignore the host cpu arch (arm or x86) or host type (bare metal or vm)?
* Do we need musl jobs?
* Do we actually need a separate pipeline that covers the "backend-mmap" feature? This is the only available implementation at the moment.

andreeaflorescu:
> A few matters that we may want to figure out before merging this:
> 
>     * Do we need separate buildkite jobs for arm or x86 for the vm-memory crate?

What do you mean by jobs? I would like to have the vm-memory tested on both platforms for each PR if that was the question.

>     * Should we have additional agent tags so that we can just pick the os type (e.g. windows, linux, etc) and ignore the host cpu arch (arm or x86) or host type (bare metal or vm)?

The way I see it is we would add two more tags in Buildkite for the OS as well (one windows, one linux). This means that we would end up with 4 tags: windows, linux, arm.metal, x86.metal. Each agent will have 2 tags: one for the OS and one for the arch. We don't currently take into account the host type, but in the future we might as from what I understand Intel plans to test it with nested as well. @sameo can probably give more details there. As for the arch I think we need two different tags as well because we want to be able to run each pipeline on the corresponding agent configuration (OS and arch). 

>     * Do we need musl jobs?

What do you mean by jobs? We need to test the musl build as well as Firecracker uses musl and we would like to eventually switch to vm-memory as well. Do you mean if we should have a separate pipeline for musl?

>     * Do we actually need a separate pipeline that covers the "backend-mmap" feature? This is the only available implementation at the moment.

Hmm, I am not sure what the best approach here is. I think we should test the vm-memory with and without the backend-mmap feature because projects that use this crate can implement their own backend that fits their needs so we should test that the default configuration (with no features) works as well. It will probably become a nightmare to maintain pipelines for each feature, I think we should think this true and come with some proposals around how this is going to work out.

petrutlucian94:
Thanks for feedback. By jobs I meant buildkite steps. which may or may not be part of the same pipeline. I'll update the PR shortly.

petrutlucian94:
The last commit doesn't look to good. We've ended up with lots of pipelines and duplicated stuff that is difficult to maintain. I haven't even defined all 8 possible pipelines, it's just so that we can discuss the options we have.

A couple of ideas to make it better:
* try to use dynamic pipelines https://github.com/buildkite/dynamic-pipeline-example
    * our pipeline "generator" script should be as clean as possible, otherwise this will lead to even more headaches
    * we'd probably end up with a single pipeline with multiple generated steps
* use subdirs, so that pipeline.linux.mmap.x86.gnu becomes
    .buildkite/pipelines/linux/mmap/x86/gnu
* simplify step names (e.g. unittests-gnu-arm becomes unnittests) since the additional details should be part of the pipeline name

To get Windows tested quicker, I guess the best way to move forward is to just have two pipelines: one for Linux (similar to the kvm ioctls one https://buildkite.com/rust-vmm/kvm-ioctls-ci/settings) and one for Windows. The linux one can continue to define multiple steps, depending on the platform and standard library.

.buildkite/pipeline.linux.yml
```diff
@@ -0,0 +1,139 @@
...
+      platform: x86_64.metal
+    plugins:
+      - docker#v3.0.1:
+          privileged: true
+          image: "fandree/rust-vmm-dev"
+          always-pull: true
+
+  - label: "unittests-gnu-arm-mmap"
+    commands:
+     - cargo test --features=backend-mmap
```
andreeaflorescu: Same as for x86 unit tests, we can have only one step with all features.

.buildkite/pipeline.linux.yml
```diff
@@ -0,0 +1,139 @@
...
+  - label: "unittests-gnu-x86-mmap"
+    commands:
+     - cargo test --features=backend-mmap
+    retry:
+      automatic: false
+    agents:
+      platform: x86_64.metal
+    plugins:
+      - docker#v3.0.1:
+          privileged: true
```
andreeaflorescu: I don't think you need to be privilegded to run these unit tests. Only the unit tests in KVM need privilegded as they need permission to open /dev/kvm and Buildkite does not support passing `--device` to the container.

.buildkite/pipeline.linux.yml
```diff
@@ -0,0 +1,139 @@
...
+          image: "fandree/rust-vmm-dev"
+          always-pull: true
+
+  - label: "unittests-gnu-x86-mmap"
```
andreeaflorescu: For unit tests we can do ```cargo test --all-features``` instead of having one step/feature. We can have `unittests-gnu-x86` and `unittests-gnu-x86-mmap` as one feature with the --all-features flag.

.buildkite/pipeline.linux.yml
```diff
@@ -0,0 +1,139 @@
...
+    retry:
+      automatic: false
+    agents:
+      platform: x86_64.metal
+    plugins:
+      - docker#v3.0.1:
+          image: "fandree/rust-vmm-dev"
+          always-pull: true
+
+  - label: "check-warnings-x86"
```
andreeaflorescu: Also add a check warnings for aarch64.

.buildkite/pipeline.windows.yml
```diff
@@ -0,0 +1,94 @@
...
+      automatic: false
+    agents:
+      platform: x86_64
+      os: windows
+    plugins:
+      - petrutlucian94/docker#v3.1.1:
+          image: "lpetrut/rust_win_buildtools"
+          always-pull: true
+
+  - label: "unittests-msvc-x86-mmap"
```
andreeaflorescu: Same comment with cargo test --all-features

.buildkite/pipeline.linux.yml
```diff
@@ -0,0 +1,139 @@
+steps:
+  - label: "build-gnu-x86"
+    commands:
+     - cargo build --release
+    retry:
+      automatic: false
+    agents:
+      platform: x86_64.metal
```
andreeaflorescu: I noticed that for the windows pipeline you also specify an "os" tag. Shouldn't we do the same for the linux pipeline? I can re-start the agent and add a linux tag to it.

andreeaflorescu:
@petrutlucian94 forgot to ask: what Rust version are you using in the windows container? In the linux one we currently have rust 1.33.0. I think it makes sense to use the same rust version when testing the code.

.buildkite/pipeline.linux.yml
petrutlucian94: Sure, works for me. This was one of the open questions I had before.

.buildkite/pipeline.linux.yml
petrutlucian94: Right, I'll drop this.

.buildkite/pipeline.linux.yml
petrutlucian94: That would be great. On one hand, it's more consistent and we avoid targeting the wrong agents by mistake.

I'll add os tags here.

.buildkite/pipeline.linux.yml
andreeaflorescu: Sorry, I didn't really get that. For unit tests I think it's perfectly fine to have it all features tested at once. I think only the builds should be separate.

.buildkite/pipeline.linux.yml
andreeaflorescu: I added the os=linux tag to the buildkite agents.

petrutlucian94:
I'm using 1.34.1, I just let rustup-init get the latest stable version. Indeed, it would make sense to use the same version.

We should probably pin the version and maybe define a constant within the rust-vmm-ci repo, which could be updated every now and then.

.buildkite/pipeline.linux.yml
petrutlucian94: Great, thanks for clarifying this.

andreeaflorescu:
@petrutlucian94 we merged the initial version of the rust-vmm-container. I was also using the latest Rust stable version, but that's not an ideal situation for reproducible builds. I added an issue about this here: https://github.com/rust-vmm/rust-vmm-container/issues/5

I think it makes a lot of sense to have all the containers used for running the CI in the same repository and when we are upgrading to a newer version we just have the update all Dockerfiles in the same repository. Can you submit a PR with the windows container you are using as well?

BTW, the container builds are currently hosted in my personal docker hub account, but once I fix some of the pressing issues with the current Dockerfiles, I was thinking about moving them to the rust-vmm organization on DockerHub that was created by @sameo. The pressing issues I am talking about are: https://github.com/rust-vmm/rust-vmm-container/issues/6 and https://github.com/rust-vmm/rust-vmm-container/issues/5

petrutlucian94:
Sounds good. I can send a PR against https://github.com/rust-vmm/rust-vmm-container/

.buildkite/pipeline.linux.yml
```diff
@@ -0,0 +1,158 @@
...
+      os: linux
+    plugins:
+      - docker#v3.0.1:
+          image: "fandree/rust-vmm-dev"
+          always-pull: true
+
+  steps:
```
andreeaflorescu: The pipeline upload fails with:
```
2019-05-30 11:58:11 FATAL  Pipeline parsing of "pipeline.linux.yml" failed (Failed to parse pipeline.linux.yml: line 14: did not find expected '-' indicator)
--
```

There should be only on "steps" per yaml file. The additional ones need to be removed. Same for the windows pipeline. This might be a leftover from the previous design.

petrutlucian94:
Whops, sorry about that.

.buildkite/pipeline.linux.yml
petrutlucian94: Whops, sorry about that. I've updated the commit.

andreeaflorescu:
@petrutlucian94 works like a charm! <3
I try to fix the musl unit tests, I think I know what's the problem with the failed test.
Can you also add another commit to delete the travis.yaml since we are switching to buildkite?

petrutlucian94:
Once we merge this, we may want to trigger the tests again for the other PRs.

andreeaflorescu:
@petrutlucian94 yap! The branches are going to be out of sync anyway so they'll have to rebase on top of these commits.

PR #26, opened 2019-06-07
Add CODEOWNERS file to auto-assign code reviewers

### LINKED ISSUE #15: Add CODEOWNERS file
Please add a CODEOWNERS file with the details of the maintainers. Please use the following format as outlined in: https://help.github.com/en/articles/about-code-owners

Most projects can simply follow the wildcard syntax. e.g.

`* @owner1 @owner2`

Not only does this provide a way to see who is responsible or this repository they will also automatically be notified of incoming PR reviews.

jiangliu:
Fix #15

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @alexandruag @bonzini @jiangliu @tkreuzer


CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @andreeaflorescu @bonzini @jiangliu @tkreuzer
```
andreeaflorescu: I don't think I should be the owner of this one. I haven't contributed at all. I believe code owners should be people with a very good understanding of the code.
I would propose you only add @bonzini @jiangliu (maybe @alexandruag if he also wants to maintain the Linux side of it) and @tkreuzer for the Windows side of things.

rbradford:
I would like to see all the people listed in this file (@alexandruag @bonzini @jiangliu @tkreuzer) approve this PR as acknowledgement that they will be autoassigned the job of reviewing changes on this repository.

andreeaflorescu:
Hmmm, it seems like this PR needed one one approval. Should I revert it till we get a response from @tkreuzer?

jiangliu:
How adding another patch instead of revert if @tkreuzer has objections?

andreeaflorescu:
Sounds good to me.

PR #29, opened 2019-06-11
Another iteration of supporting file backed memory regions

alexandruag:
This PR is based on code from Sebastien's original proposal (#20) and the subsequent comments and conversations.

As per Paolo's suggestion, the offset is now represented as an `u64`. I didn't name the new struct `FileRange`, because it should've also kept a `length` field, and that information is already available via the `len()` method of the `GuestMemoryRegion`. Instead I've called it `FileOffset` (will rename on demand), and it only keeps the `Arc<File>` and `offset`.

The Linux code is clearly ahead of the Windows one right now, and some harmonization is in order. I've just created a Windows VM on my machine, but didn't get the chance to install Rust and test things yet. It would be good to know if ppl are ok with the proposed way of implementing this functionality on Linux, and then we can bring the Windows implementation up to date as well (and add some more tests to this PR).

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -29,8 +29,10 @@
 
 use std::convert::From;
 use std::fmt::{self, Display};
+use std::fs::File;
 use std::io::{self, Read, Write};
 use std::ops::{BitAnd, BitOr};
+use std::sync::Arc;
 
 use address::{Address, AddressValue};
 use bytes::Bytes;
@@ -114,6 +116,40 @@ impl_address_ops!(MemoryRegionAddress, u64);
 
 /// Type of the raw value stored in a GuestAddress object.
 pub type GuestUsize = <GuestAddress as AddressValue>::V;
+
+/// Represents the start point within a `File` that backs a `GuestMemoryRegion`.
+#[derive(Clone, Debug)]
+pub struct FileOffset {
+    file: Arc<File>,
+    start: u64,
+}
+
+impl FileOffset {
+    /// Creates a new `FileOffset` object.
+    pub fn new(file: File, start: u64) -> Self {
+        FileOffset::from_arc(Arc::new(file), start)
+    }
+
+    /// Creates a new `FileOffset` object based on an exiting `Arc<File>`.
+    pub fn from_arc(file: Arc<File>, start: u64) -> Self {
+        FileOffset { file, start }
+    }
+
+    /// Returns a reference to the inner `File` object.
+    pub fn file(&self) -> &File {
+        &self.file.as_ref()
+    }
+
+    /// Return a reference to the inner `Arc<File>` object.
+    pub fn arc(&self) -> &Arc<File> {
+        &self.file
+    }
+
+    /// Returns the start offset within the file.
+    pub fn start(&self) -> u64 {
+        self.start
+    }
+}
 
 /// Represents a continuous region of guest physical memory.
 #[allow(clippy::len_without_is_empty)]
@@ -513,5 +549,17 @@ mod tests {
         let a = GuestAddress(0xff);
         assert_eq!(Some(GuestAddress(0x0f)), a.checked_sub(0xf0));
         assert!(a.checked_sub(0xffff).is_none());
+    }
+
+    #[test]
+    fn test_file_offset() {
+        let file = tempfile::tempfile().unwrap();
+        let start = 1234;
+        let file_offset = FileOffset::new(file, start);
+        assert_eq!(file_offset.start(), start);
+        assert_eq!(
+            file_offset.file() as *const File,
+            file_offset.arc().as_ref() as *const File
+        );
     }
 }
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -200,6 +200,9 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
             .and_then(|offset| self.check_address(MemoryRegionAddress(offset)))
     }
 
+    /// Returns information regarding the file and offset backing this memory region.
+    fn file_offset(&self) -> Option<&FileOffset>;
+
     /// Return a slice corresponding to the data in the region; unsafe because of
     /// possible aliasing.  Return None if the region does not support slice-based
     /// access.
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -20,8 +20,9 @@
 //! - [GuestMemoryMmap](struct.GuestMemoryMmap.html): provides methods to access a collection of
 //! GuestRegionMmap objects.
 
-use std::io::{self, Read, Write};
+use std::io::{Read, Write};
 use std::ops::Deref;
+use std::result;
 use std::sync::Arc;
 
 use address::Address;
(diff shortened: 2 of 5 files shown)


src/mmap.rs
```diff
@@ -64,12 +71,14 @@ pub struct GuestRegionMmap {
 
 impl GuestRegionMmap {
     /// Create a new memory-mapped memory region for guest's physical memory.
-    /// Note: caller needs to ensure that (mapping.len() + guest_base) doesn't wrapping around.
-    pub fn new(mapping: MmapRegion, guest_base: GuestAddress) -> Self {
-        GuestRegionMmap {
+    pub fn new(mapping: MmapRegion, guest_base: GuestAddress) -> result::Result<Self, Error> {
+        if guest_base.0.checked_add(mapping.len() as u64).is_none() {
+            return Err(Error::InvalidGuestRegion);
+        }
```
jiangliu: Should we also check that the offset and len are page aligned? Or just let the mmap syscall to check them?

src/guest_memory.rs
```diff
@@ -512,4 +546,12 @@ mod tests {
         assert_eq!(Some(GuestAddress(0x0f)), a.checked_sub(0xf0));
         assert!(a.checked_sub(0xffff).is_none());
     }
+
+    #[test]
+    fn test_file_offset() {
+        let file = tempfile::tempfile().unwrap();
+        let start = 1234;
+        let file_offset = FileOffset::new(file, start);
+        assert_eq!(file_offset.start(), start);
```
jiangliu: How about adding a test case for non-page-aligned offset?

rbradford:
@sboeuf What's your thoughts on this?

sboeuf:
@rbradford 
>@sboeuf What's your thoughts on this?

I'm planning to review this today :)

src/mmap.rs
sboeuf: I guess it's better to check it ourselves since we cannot assume that both Linux and Windows `mmap` syscall will check for that, right?

src/mmap.rs
```diff
@@ -315,54 +320,53 @@ impl GuestMemoryMmap {
...
+        Self::with_files(ranges.iter().map(|r| (r.0, r.1, None)))
+    }
 
-        Ok(Self {
-            regions: Arc::new(regions),
-        })
+    /// Creates a container and allocates anonymous memory for guest memory regions.
+    /// Valid memory regions are specified as a Vec of (Address, Size, Option<FileOffset>)
+    /// tuples sorted by Address.
+    pub fn with_files<A, T>(ranges: T) -> result::Result<Self, Error>
```
sboeuf: Looks good, I feel that long term the constructor `new()` could be removed.

src/mmap.rs
```diff
@@ -315,54 +320,53 @@ impl GuestMemoryMmap {
...
+    where
+        A: Borrow<(GuestAddress, usize, Option<FileOffset>)>,
```
sboeuf: Rust question: why is this Borrow trait needed here? What prevent us from using references in the code below?

src/mmap.rs
alexandruag: It was used in some tests and it looks more convenient when someone doesn't require file mappings. I can remove it if you are totally against it, but otherwise looks relatively useful for such a small method IDK :-s

src/mmap.rs
alexandruag: I used `Borrow` here because I wanted more flexibility in the `Item` the iterator returns. For example, if we would have stated the bound `T: IntoIterator<Item=&(GuestAddress, usize, Option<FileOffset>)>`, the current implementation of `new` would not work, because the `map` adapter creates an iterator that returns actual tuples (`(GuestAddress, usize, Option<FileOffset>)`) instead of references to them, and I can't simply add a `&` in front of the result because the compiler complains the reference might not live long enough. With `Borrow`, it works because we accept iterators whose items are anything which can be borrowed as `&(GuestAddress, usize, Option<FileOffset>)`.

`Borrow` is relatively similar to `AsRef` and `Deref`, but unlike them comes with the following blanket implementations: 
```rust
impl<'_, T> Borrow<T> for &'_ T
where
    T: ?Sized
```
and
```rust
impl<T> Borrow<T> for T
where
    T: ?Sized
```
which guarantee that both `(GuestAddress, usize, Option<FileOffset>)` and `&(GuestAddress, usize, Option<FileOffset>)` implement `Borrow<(GuestAddress, usize, Option<FileOffset>)`, and are thus usable as iterator item types in `with_files`.

src/mmap.rs
sboeuf: Fine with me, you can keep it.

src/mmap.rs
sboeuf: Makes sense, thanks for the explanation :)

src/mmap_unix.rs
```diff
@@ -52,53 +74,75 @@ impl MmapRegion {
...
+        Self::build(
+            Some(file_offset),
+            size,
+            libc::PROT_READ | libc::PROT_WRITE,
+            libc::MAP_NORESERVE | libc::MAP_SHARED,
+        )
+    }
+
+    /// Creates a new mapping based on the provided arguments.
+    pub fn build(
```
sboeuf: I think that half of this function should be generic to unix and windows, therefore should be defined in `mmap.rs`. The file offset processing is the part that should be common.

src/mmap_unix.rs
```diff
@@ -52,53 +74,75 @@ impl MmapRegion {
...
-        let addr = unsafe {
-            libc::mmap(
-                null_mut(),
-                size,
-                libc::PROT_READ | libc::PROT_WRITE,
-                libc::MAP_SHARED,
-                fd.as_raw_fd(),
-                offset,
-            )
+    pub fn from_file(file_offset: FileOffset, size: usize) -> Result<Self> {
```
sboeuf: We need to modify the same way the function `from_fd` in `mmap_windows.rs`. This should be pretty straightforward.

src/mmap_unix.rs
alexandruag: Make sense, I refactored this by adding the `check_file_offset` function to `mmap.rs` in the same commit.

alexandruag:
@bonzini @jiangliu can you please take a look?

src/mmap.rs
```diff
@@ -45,15 +51,42 @@ pub(crate) trait AsSlice {
...
+) -> result::Result<(), MmapRegionError> {
+    let file = file_offset.file();
+    let start = file_offset.start();
+
+    if let Some(end) = start.checked_add(size as u64) {
+        if let Ok(metadata) = file.metadata() {
+            if metadata.len() < end {
+                return Err(MmapRegionError::MappingPastEof);
+            }
+        }
```
jiangliu: Do we need to handle the case when file.metadata() fails?

sboeuf: We could create a special Error case for it.
Once @alexandruag updates it, we need to re-approve the PR.

sboeuf:
The green button is very tempting :)

src/mmap.rs
jiangliu: We could merge the big PR first and submit another bugfix PR for this issue.
Review a small bugfix PR is much easier:)

PR #32, opened 2019-06-17
guest_memory: Fix region access violation

bjzhjing:
While call try_access from read_from, in try_access, it does not
check if start + len will cross over the end of the current region.
If it will, once pass the region to the closure, where the region
be converted to slice, and data is picked up there, from the start
to the end = start + len, then, region access violation will happen,
since start + len is out of range of the slice.

Get the len in a differen way in try_access to ensure it will not
cross over the region to avoid such error.

Signed-off-by: Cathy Zhang <cathy.zhang@intel.com>

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -334,9 +334,9 @@ pub trait GuestMemory {
         let mut total = 0;
         while let Some(region) = self.find_region(cur) {
             let start = region.to_region_addr(cur).unwrap();
-            let cap = region.len() as usize;
-            let len = std::cmp::min(cap, count - total);
-            match f(total, len, start, region) {
+            let cap = region.len() - start.raw_value();
+            let len = std::cmp::min(cap, (count - total) as GuestUsize);
+            match f(total, len as usize, start, region) {
                 // no more data
                 Ok(0) => break,
                 // made some progress
@@ -492,6 +492,19 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
 #[cfg(test)]
 mod tests {
     use super::*;
+    #[cfg(feature = "backend-mmap")]
+    use crate::{GuestAddress, GuestMemoryMmap};
+    #[cfg(feature = "backend-mmap")]
+    use std::io::Cursor;
+
+    #[cfg(feature = "backend-mmap")]
+    fn make_image(size: u8) -> Vec<u8> {
+        let mut image: Vec<u8> = Vec::with_capacity(size as usize);
+        for i in 0..size {
+            image.push(i);
+        }
+        image
+    }
 
     #[test]
     fn offset_from() {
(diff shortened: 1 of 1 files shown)


src/guest_memory.rs
```diff
@@ -295,8 +295,8 @@ pub trait GuestMemory {
         let mut total = 0;
         while let Some(region) = self.find_region(cur) {
             let start = region.to_region_addr(cur).unwrap();
-            let cap = region.len() as usize;
-            let len = std::cmp::min(cap, count - total);
+            let size = region.len() as usize - start.raw_value() as usize;
```
jiangliu: Sorry, at a second glance found it's unsafe to cast region.len() and start.raw_value() as usize.
usize is 32bit with a 32bit VMM but region.len() and start.raw_value() are both u64.

bjzhjing:
@chao-p  Yes, I encounter this error while adding hugepages support, in the case with hugepagesize=2M, it fails to write elf image to memory. After add this code change, elf image could be loaded.

bonzini:
Looks good but it needs a test case. Also the name of the variable "cap" is better: `size` is accurate in the buggy code but not in the fixed one.

src/guest_memory.rs
bonzini: Yes, this must be `GuestUsize`.

bjzhjing: It seems this is not the only case in guest_memory.rs regarding the potential truncate issue, I suggest to have a separate PR opened to handle such issue.

bjzhjing:
@bonzini I thought "cap" represents the capacity of the region, but here the parameter is only the length/size of the sub-set data in the region. Anyway, it's changed back to be "cap", and a test case is added for this change :-P

src/guest_memory.rs
bonzini: Still, this should use `GuestUsize`.

src/guest_memory.rs
```diff
@@ -295,8 +295,8 @@ pub trait GuestMemory {
         let mut total = 0;
         while let Some(region) = self.find_region(cur) {
             let start = region.to_region_addr(cur).unwrap();
-            let cap = region.len() as usize;
-            let len = std::cmp::min(cap, count - total);
+            let cap = region.len() - start.raw_value();
+            let len = std::cmp::min(cap as usize, count - total);
```
jiangliu: There is still a risk that cap is bigger than usize on 32-bit platforms.
How about:
            let cap = region.len() - start.raw_value();
            let len = std::cmp::min(cap, (count - total) as GuestUsize);	   
           match f(total, len as usize, start, region) {

bonzini: Yes, it's better. @bjzhjing can you adjust?

bjzhjing: Sure, change is done!

jiangliu:
@bjzhjing hi, this PR has some conflicts with the latest master branch, so could you help to rebase the PR?

bjzhjing:
@jiangliu Rebase to the newest version, and submit the commit again.

bjzhjing:
Is it possible to merge the change and close the PR ASAP, since Gerry and Paolo have approve the change previously. There is no update during the latest rebase.

jiangliu:
> Is it possible to merge the change and close the PR ASAP, since Gerry and Paolo have approve the change previously. There is no update during the latest rebase.

Still need another approver:(

PR #35, opened 2019-06-21
[DNM] Fix windows for PR #29

sboeuf:


sboeuf:
@alexandruag this PR is not intended to be merged, but more a reference based on your PR #29, which I rebased on top of `master` and where I added one new commit to fix windows build and unit tests.
Please squash it into your PR if you agree with the code. The point is to prevent from breaking Windows, but I think @tkreuzer should submit a follow up PR to add more tests once #29 gets merged.

One thing to notice, I followed this [document](https://zork.net/~st/jottings/rust-windows-and-debian.html) to be able build the code and the tests for windows, this was very handy.

jiangliu:
Should we close this PR now?

Issue #38, opened 2019-07-26
Windows: leaked mmap regions


petrutlucian94:
We're incorrectly using VirtualFree in the MmapRegion Drop trait implementation, for which reason it will be silently leaking memory. This function will fail if the size is specified when releasing memory.

PR #39, opened 2019-07-26
[windows] Fix MmapRegion drop

petrutlucian94:
At the moment, we're not properly deallocating memory mapped regions.
The "VirtualFree" function will fail if the region size is specified
when using the "MEM_RELEASE" flag.

This change fixes it, also printing a message when VirtualFree fails,
which may help discovering memory leaks more easily.

Closes: #38

### DIFF
--- a/src/mmap_windows.rs
+++ b/src/mmap_windows.rs
@@ -20,6 +20,8 @@ use std::ptr::{null, null_mut};
 
 use libc::{c_void, size_t};
 
+use winapi::um::errhandlingapi::GetLastError;
+
 use crate::guest_memory::FileOffset;
 use crate::mmap::AsSlice;
 use crate::volatile_memory::{self, compute_offset, VolatileMemory, VolatileSlice};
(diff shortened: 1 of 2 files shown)


jiangliu:
It's still not `cargo fmt` clean, please help to fix it.

petrutlucian94:
> It's still not `cargo fmt` clean, please help to fix it.

My bad, should be fine now. I usually try to avoid getting passed 80 characters per line but rustfmt doesn't seem to like the way I'm splitting lines :).

src/mmap_windows.rs
```diff
@@ -219,8 +221,20 @@ impl Drop for MmapRegion {
     fn drop(&mut self) {
         // This is safe because we mmap the area at addr ourselves, and nobody
         // else is holding a reference to it.
+        // NOTE(lpetrut): The size must be set to 0 when using MEM_RELEASE,
```
bonzini: No need to add your name here. :)

petrutlucian94: Sure, sorry about that. It's just a habit from working on other projects that requested notes, TODOs, etc to include the name/nickname of the one adding those.

Issue #36, opened 2019-07-01
Switch to rust-vmm-ci


andreeaflorescu:
Use rust-vmm-ci as a gitsubmodule to test the code. Since vm-memory has non-default features, a custom pipeline is also required.

For more details about how to make the switch, please check the [rust-vmm-ci readme](https://github.com/rust-vmm/rust-vmm-ci)

PR #40, opened 2019-07-31
switch to rust-vmm-ci

jiangliu:
Removed the existing buildkite pipeline and added the rust-vmm-ci
submodule as this is the agreed upon way of testing all crates in
rust-vmm.

Fixes: #36


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "rust-vmm-ci"]
```
andreeaflorescu: You're missing the `coverage_config.json` file.

jiangliu: fixed

PR #42, opened 2019-08-26
 mmap: Add a type parameter to the memory mapped regions 

### LINKED ISSUE #34: Create types of memory regions
I think the next step for the vm-memory is to define types (Acpi, Ram, Persistent, ...) of memory regions.

The first use case I have for this is to support `virtio-pmem` properly from the cloud-hypervisor VMM. The idea is that virtio-pmem memory map the backing file into the VMM address space, to make it available to the VM. So this is a clear use case where we want the portion of memory exposed from virtio-pmem to be part of the GuestMemory as a separate region. But this region cannot be considered the same as the region representing the RAM, otherwise the virtio device implementations will be broken, thinking that it is part of the VM RAM.
That's why having types associated with regions could be a simple way for any code in the VMM to check if the region should be considered or ignored, depending on what you're trying to achieve.

I think this should be added once #29 will be merged.

@bonzini @jiangliu @alexandruag WDYT?

sboeuf:
In order to let the caller identify regions based on the region type,
this patch introduces a new parameter as part of the region creation.

Fixes #34

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -267,10 +267,16 @@ pub trait GuestMemory {
     /// ```
     /// # #[cfg(feature = "backend-mmap")]
     /// # fn test_map_fold() -> Result<(), ()> {
-    /// # use vm_memory::{GuestAddress, GuestMemory, GuestMemoryRegion, mmap::GuestMemoryMmap};
+    /// # use vm_memory::{
+    ///     GuestAddress, GuestMemory, GuestMemoryRegion, mmap::GuestMemoryMmap,
+    ///     RegionType
+    /// };
     ///     let start_addr1 = GuestAddress(0x0);
     ///     let start_addr2 = GuestAddress(0x400);
-    ///     let mem = GuestMemoryMmap::new(&vec![(start_addr1, 1024), (start_addr2, 2048)]).unwrap();
+    ///     let mem = GuestMemoryMmap::new(&vec![
+    ///         (start_addr1, 1024, RegionType::Ram),
+    ///         (start_addr2, 2048, RegionType::Ram)
+    ///         ]).unwrap();
     ///     let total_size = mem.map_and_fold(
     ///         0,
     ///         |(_, region)| region.len() / 1024,
@@ -493,7 +499,7 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
 mod tests {
     use super::*;
     #[cfg(feature = "backend-mmap")]
-    use crate::{GuestAddress, GuestMemoryMmap};
+    use crate::{GuestAddress, GuestMemoryMmap, RegionType};
     #[cfg(feature = "backend-mmap")]
     use std::io::Cursor;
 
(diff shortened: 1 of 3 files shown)


sboeuf:
@jiangliu PTAL

sboeuf:
@sameo PTAL

jiangliu:
We are thinking about the relationship between vm-allocator(vm-resource) and vm-memory, so could we delay the review when vm-resource design has set down?

bonzini:
I think you should just not use `GuestMemoryMmap` in this case; having a random `u64` tag is not a very "Rust'y API.

sboeuf:
@bonzini should we reopen? (since you left a comment saying `looked at the wrong commit.`)

sboeuf:
@bonzini @jiangliu With this PR, I'm looking for a simple way to identify regions as belonging to certain type of memory. The goal being to reuse the `vm-memory` crate as much as possible so that I could rely on it to perform the `mmap` of regions related to `virtio-pmem` or `virtio-fs`. But then I need a way to differentiate the regions from the consumer's code perspective.

bonzini:
I edited the message to refer to types instead of tags, but I still agreed with the gist of it. The type enum is not very extensible. I can imagine quite a few kinds of memory regions backed by mmap-ed memory: RAM, ROM, PCI BARs from assigned devices (where the size of accesses matter and thus `as_slice`/`as_mut_slice` should return `None`), flash memory (acts as ROM when read but does not discard writes and instead goes into some kind of MMIO-based command mode).

Now that `GuestMemory` supports external reference counts, this kind of ancillary data could be kept in an external hash table. Unfortunately Rust doesn't make it easy to extend `GuestRegionMmap` in the users of the crate, but IMHO that's not a reason to put everything in `vm-memory`. Perhaps later on we could add a full-blown `GuestMemory` implementation that supports all of the usecases above (something like QEMU's memory API), but I'd rather keep `GuestMemory` as simple as possible.

sboeuf:
>Now that GuestMemory supports external reference counts, this kind of ancillary data could be kept in an external hash table. Unfortunately Rust doesn't make it easy to extend GuestRegionMmap in the users of the crate, but IMHO that's not a reason to put everything in vm-memory.

Because the type is related to each region, and since the regions are inside the `GuestMemoryMmap`, I thought putting the `type` information about each region as part of the `GuestRegionMmap` was the most logical.
Now if you want to externalize this out of `vm-memory` with some hash table created by the consumer of this crate, I guess we need some sort of identifier for each region, don't we?

bonzini:
> Because the type is related to each region, and since the regions are inside the GuestMemoryMmap, I thought putting the type information about each region as part of the GuestRegionMmap was the most logical.

And I must admit it is indeed. What I didn't like was forcing the definition of the enum into vm-memory. It's difficult to resolve the tension between staying simple and being sufficiently general-purpose.

Regarding the identifier for the region, the starting address could be one, but actually even the `Arc<GuestRegionMmap>` itself can be an identifier for the region now that we have that in vm-memory.

sboeuf:
Actually, maybe the right way to proceed from the user is to create one `GuestMemoryMmap` per type, this way it can manage set of regions per type, through each `GuestMemoryMmap`.

PR #43, opened 2019-08-26
mmap: Allow an existing GuestMemoryMmap to be extended

sboeuf:
The public API of vm-memory is pretty limited regarding the actions
that can be performed on a GuestMemoryMmap. The only allowed action
is to create a new instance of a GuestMemoryMmap based on a list of
memory ranges, but there are use cases where we need to extend an
existing instance by appending more memory regions to the existing
ones.

This PR creates a new public function that can be used from any
caller in order to add a new memory region to an existing guest memory.
Note that rework of the existing code was necessary to apply proper
factorization.

### DIFF
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -26,7 +26,6 @@ use std::fmt;
 use std::io::{Read, Write};
 use std::ops::Deref;
 use std::result;
-use std::sync::Arc;
 
 use crate::address::Address;
 use crate::guest_memory::{
@@ -363,9 +362,9 @@ impl GuestMemoryRegion for GuestRegionMmap {
 }
 
 /// Tracks memory regions allocated/mapped for the guest in the current process.
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct GuestMemoryMmap {
-    regions: Arc<Vec<GuestRegionMmap>>,
+    regions: Vec<GuestRegionMmap>,
 }
 
 impl GuestMemoryMmap {
@@ -427,9 +426,7 @@ impl GuestMemoryMmap {
             }
         }
 
-        Ok(Self {
-            regions: Arc::new(regions),
-        })
+        Ok(Self { regions })
     }
 
     /// Convert an absolute address into an address space (GuestMemory)
@@ -1000,8 +997,8 @@ mod tests {
             .map(|x| (x.0, x.1))
             .eq(iterated_regions.iter().map(|x| *x)));
 
-        assert_eq!(gm.clone().regions[0].guest_base, regions[0].0);
-        assert_eq!(gm.clone().regions[1].guest_base, regions[1].0);
+        assert_eq!(gm.regions[0].guest_base, regions[0].0);
+        assert_eq!(gm.regions[1].guest_base, regions[1].0);
     }
 
     #[test]
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -21,6 +21,7 @@
 //! GuestRegionMmap objects.
 
 use std::borrow::Borrow;
+use std::cmp::Ordering;
 use std::error;
 use std::fmt;
 use std::io::{Read, Write};
@@ -139,6 +140,26 @@ impl GuestRegionMmap {
         // is safe because we've just range-checked addr using check_address.
         self.check_address(addr)
             .map(|addr| self.as_ptr().wrapping_offset(addr.raw_value() as isize))
+    }
+}
+
+impl Ord for GuestRegionMmap {
+    fn cmp(&self, other: &GuestRegionMmap) -> Ordering {
+        self.guest_base.cmp(&other.guest_base)
+    }
+}
+
+impl PartialOrd for GuestRegionMmap {
+    fn partial_cmp(&self, other: &GuestRegionMmap) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Eq for GuestRegionMmap {}
+
+impl PartialEq for GuestRegionMmap {
+    fn eq(&self, other: &GuestRegionMmap) -> bool {
+        self.guest_base == other.guest_base
     }
 }
 
@@ -427,6 +448,29 @@ impl GuestMemoryMmap {
         }
 
         Ok(Self { regions })
+    }
+
+    /// Add a new region to the set of existing regions.
+    ///
+    /// # Arguments
+    ///
+    /// * `region` - The region to add.
+    ///               The region shouldn't overlap with the existing regions.
+    pub fn add_region(&mut self, region: GuestRegionMmap) -> result::Result<(), Error> {
+        match self.regions.binary_search(&region) {
+            Ok(_) => return Err(Error::MemoryRegionOverlap),
+            Err(pos) => {
+                if (pos > 0 && self.regions[pos - 1].end_addr() >= region.start_addr())
+                    || (pos < self.regions.len()
+                        && region.end_addr() >= self.regions[pos].start_addr())
+                {
+                    return Err(Error::MemoryRegionOverlap);
+                }
+                self.regions.insert(pos, region);
+            }
+        }
+
+        Ok(())
     }
 
     /// Convert an absolute address into an address space (GuestMemory)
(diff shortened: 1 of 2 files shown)


sboeuf:
@jiangliu PTAL

sboeuf:
@sameo PTAL

dagrh:
The question of a 'ram region' gets pretty thorny at times; do you mean ROM is included? Do you mean anything with a real piece of host memory pointer is included?
If you're updating the sizes then what are the rules for a caller iterating over it?
(Also were you trying to keep regions in order?)

sboeuf:
@dagrh 
>The question of a 'ram region' gets pretty thorny at times; do you mean ROM is included?

`ROM` can be a specific type of memory region if we somehow have the need to make the distinction between the two.

>Do you mean anything with a real piece of host memory pointer is included?

So far, that's my understanding of what are the memory regions.

>If you're updating the sizes then what are the rules for a caller iterating over it? (Also were you trying to keep regions in order?)

Well the iterating should be fine as long as we introduce some types. And then based on the type and the base address, the caller should know what needs to be done.
And I think that's joining my question about what is the assumption regarding the behavior of vm-memory.

jiangliu:
In my original thoughts, the GuestMemory just provides mechanisms to access guest's memory. We may build another abstraction AddressSpace, which holds all resources present in the guest's system bus.
Then we may generate GuestMemory objects from  the AddressSpace object on demand.
And the type of resources (the proposed `tag`) should be an attributes of entries in the AddressSpace object.

rbradford:
@jiangliu / @dagrh I think you're might be conflating the issue that is trying to be resolved in #42 with this one. Whatever the ideal design may be as it currently stands once the `GuestMemoryMmap` is constructed it is not possible to add any more regions to it. It supports adding multiple at the start but there are situations such as virtio-fs where they need to be added later.

We want to start using the support that @sboeuf added in cloud-hypervisor so that we can benefit from the `GuestMemoryMmap` managing our virtio-fs and virtio-pmem regions so that we don't have to.

Please consider reviewing it from the current implementation and behavioural perspective rather than a possible future ideal as this is a feature we need and hence why it's been proposed.

sboeuf:
@rbradford PTAL

src/mmap.rs
```diff
@@ -427,9 +447,30 @@ impl GuestMemoryMmap {
             }
         }
 
-        Ok(Self {
-            regions: Arc::new(regions),
-        })
+        Ok(Self { regions })
+    }
+
+    /// Add a new `GuestRegionMmap` to a vector of regions.
```
rbradford: nit: You're leaking the implementation details in the comment.  You can simply say "Add a new region to the set of existing regions."

sboeuf: done!

sboeuf:
@jiangliu @andreeaflorescu could you please take a quick look. The second commit introducing a new function `add_region()` is pretty obvious, but please take a deeper look at the first one removing the `Arc`. I need the list of regions to be mutable and I decided to remove the `Arc`, but another approach might be to modify `Arc` into an `Arc<Mutex<>>`. It really depends on what is the expected behavior from the consumer. I know that in case of `cloud-hypervisor`, we move the `GuestMemoryMmap` into an `Arc<Mutex>` so we don't need it from the structure itself, but I'm really opened to discussion on this :)

alexandruag:
Hi, will review tomorrow!

bonzini:
> Whatever the ideal design may be as it currently stands once the GuestMemoryMmap is constructed it is not possible to add any more regions to it. It supports adding multiple at the start but there are situations such as virtio-fs where they need to be added later.

In QEMU this is solved by adding a layer of indirection: the `AddressSpace` is mutable, but the equivalent of `GuestMemory` is actually the `FlatView` and it is immutable. This also has advantages in that we use RCU to avoid locking when accessing `FlatView`s. I suggest doing the same in vm-memory.

jiangliu:
> > Whatever the ideal design may be as it currently stands once the GuestMemoryMmap is constructed it is not possible to add any more regions to it. It supports adding multiple at the start but there are situations such as virtio-fs where they need to be added later.
> 
> In QEMU this is solved by adding a layer of indirection: the `AddressSpace` is mutable, but the equivalent of `GuestMemory` is actually the `FlatView` and it is immutable. This also has advantages in that we use RCU to avoid locking when accessing `FlatView`s. I suggest doing the same in vm-memory.

I like this idea. In a previous PR, we have an address_space abstraction.  And in our dragonball VMM, we are still using the address_space abstraction.

sboeuf:
@bonzini 
>In QEMU this is solved by adding a layer of indirection: the AddressSpace is mutable, but the equivalent of GuestMemory is actually the FlatView and it is immutable. This also has advantages in that we use RCU to avoid locking when accessing FlatViews. I suggest doing the same in vm-memory.

So you mean the number of regions is immutable? How do you perform memory hotplug? Do you have only one big region?
Not sure I follow the comparison but maybe what you're describing is the equivalent of having `GuestMemoryMmap` holding an `Arc<Mutex<Vec<GuestMemoryRegion>>>`. In this case we can make `GuestMemoryMmap` and all its methods immutable, since the `Arc<Mutex>` is the indirection layer in this case.

bonzini:
> So you mean the number of regions is immutable? How do you perform memory hotplug?

Devices don't use `GuestMemoryMmap` directly. It can work like this:

- the hotplug controller creates a new `GuestMemoryMmap` and stores it into a `crossbeam_epoch::Atomic<GuestMemoryMmap>` inside the hotplug controller itself. For example:

```rust
impl GuestAddressSpace {
   fn hotplug(&self, r: GuestRegionMmap) {
       let regions = self.regions_mutex.lock(); // a Mutex<Vec<Box<GuestRegionMmap>>>
       regions.add(Box::new(r));
       // each hotplug creates a new memory map:
       let mmap = GuestMemoryMmap::from_regions(regions.clone());
       let owned = crossbeam_epoch::Owned::new(mmap);
       current_mmap.store(owned, Release);
    }
}
```

- you create a new `GuestMemory` implementation that holds an `Arc<GuestAddressSpace>` and, for each method, forwards it to the `GuestMemoryMmap` like so:

```rust
impl GuestMemory for GuestMemoryIndirect {
    fn read(&self, buf: &mut [u8], addr: MemoryRegionAddress) -> guest_memory::Result<usize> {
        let guard = &crossbeam_epoch::pin();
        let mmap = guest_address_space.current_mmap.load_consume(guard);
        mmap.read(buf, addr)
    }
}
```
(I think that due to orphan rules you cannot implement `GuestMemory` directly on `Arc<GuestAddressSpace>` but I may be wrong) and then you pass the `GuestMemoryIndirect` to the devices.

sboeuf:
@bonzini 
Ok I understand now. That's an interesting way of solving the problem. This moves the "locking" (which is handled by by `Atomic` here) internally, and that's how you deal with immutable `GuestMemoryMmap`. And this would allow for better performances since no locking would need to be involved, either from the consumer or from the internals of the crate.

About `crossbeam_epoch`, is it acceptable to pull this dependency into the `vm-memory` crate?

And about `Vec<GuestRegionMmap>` it __must__ be a `Vec<Arc<GuestRegionMmap>>` since the cloning would cause duplication and at the time one of the two copies is dropped, the `Drop` trait would be called, causing the unmapping of the actual memory region.

bonzini:
> About crossbeam_epoch, is it acceptable to pull this dependency into the vm-memory crate?

It should be hidden behind a feature flag, but yes I think it's standard enough.

> And about Vec<GuestRegionMmap> it must be a Vec<Arc<GuestRegionMmap>> since the cloning would cause duplication and at the time one of the two copies is dropped, the Drop trait would be called, causing the unmapping of the actual memory region.

Right, and I think changing `GuestMemoryMmap` to have a `Vec<Arc<GuestRegionMmap>>` is acceptable for vm-memory so you could start from there.

sboeuf:
@sameo @rbradford @andreeaflorescu @alexandruag @jiangliu I'd like to hear your opinions on this :)

alexandruag:
The approach proposed by Paolo is very interesting. I'm super curious to have a look at the functionality from `crossbeam_epoch` to see what are the implications and whether what we need is simple enough to roll our own. 

Meanwhile, what do you think about `GuestMemoryMmap` being generic in terms of the underlying `GuestRegionMmap` collection? That way, we can continue using a simple `Vec` if we don't need more advanced functionality, or something along the lines of Paolo's proposition that also allows regions to be added, for more complex use cases. 

Seems like the collection type only needs to implement something like `len()`, `iter()` and `From<Vec<GuestRegionMmap>>` for the rest of the code to keep working right now.

bonzini:
> whether what we need is simple enough to roll our own.

I don't think it's a good idea to roll your own. Actually I have already rolled my own [here](https://github.com/bonzini/rust-rcu-qemu/commit/759a98e7594bca4037739d4b05552b1a9a1e09bd) and [here](https://github.com/bonzini/rust-rcu-qemu/commit/73a30baa7e08795b2cf96673b3456b18df9bfd3c) for learning purposes, but it's the kind of thing that is easy to do a toy implementation for, and quite hard to turn into something serious. (That repository also has a [C implementation with a Rust wrapper](https://github.com/bonzini/rust-rcu-qemu/commit/aa6aec708db3f404f2b39d18626f7c2e52612415) which _is_ production quality but Linux-only).

> What do you think about GuestMemoryMmap being generic in terms of the underlying GuestRegionMmap collection? That way, we can continue using a simple Vec if we don't need more advanced functionality, or something along the lines of Paolo's proposition that also allows regions to be added, for more complex use cases.

I think my proposal should still build on `GuestMemoryMmap` and be entirely optional, so `GuestMemoryMmap` can and should keep using a `Vec`.

alexandruag:
Thanks for the details Paolo! I gave it more thought (and I understand these things a bit better now), and your original proposal is clearly neater than what I was saying.

bonzini:
Let's close this pull request. Cloud-hypervisor can start experimenting with `crossbeam_epoch` now that #54 and #56 have been merged.

PR #44, opened 2019-09-01
Introduce trait GetHostAddress

jiangliu:
Previously the get_host_address() implenents the way to map guest physical address to host virtual
address, but it's implemented as a method of the GuestMemoryMmap struct. There's no formal way
for clients to state the dependency on the capability to map GPA to HVA. So introduce trait
GetHostAddress to describe the capability to map GPA to HVA, so clients may depend on it.

### DIFF
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -132,15 +132,6 @@ impl GuestRegionMmap {
             guest_base,
         })
     }
-
-    /// Convert an absolute address into an address space (GuestMemory)
-    /// to a host pointer, or return None if it is out of bounds.
-    pub fn get_host_address(&self, addr: MemoryRegionAddress) -> Option<*mut u8> {
-        // Not sure why wrapping_offset is not unsafe.  Anyway this
-        // is safe because we've just range-checked addr using check_address.
-        self.check_address(addr)
-            .map(|addr| self.as_ptr().wrapping_offset(addr.raw_value() as isize))
-    }
 }
 
 impl Deref for GuestRegionMmap {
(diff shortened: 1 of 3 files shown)


jiangliu:
@alexandruag @bonzini Could you please help to review this PR? The vhost crate depends on this PR.

src/mmap.rs
```diff
@@ -470,6 +464,15 @@ impl GuestMemory for GuestMemoryMmap {
     }
 }
 
+impl GetHostAddress for GuestMemoryMmap {
```
andreeaflorescu: The naming is a bit weird I think. How about `HostAddress`?

jiangliu: Renamed as HostAddress:)

bonzini:
Let's do this instead:

* add a `AddressNotAvailable` case to `guest_memory::Error`

* change both implementations of `get_host_address` so that they return a `Result<*mut u8, GuestMemoryError>` (adding an `.ok_or(InvalidBackendAddress)` to the implementation in GuestMemoryRegion)

* add a `get_host_address` in GuestMemoryRegion (with a default implementation that is just `Error(AddressNotAvailable)`)

* move the implementation of `get_host_address` for GuestMemoryMmap up to GuestMemory.

bonzini:
>  The proposed HostAddress trait explicitly states that the backend support mmapping guest memory into the vmm process, which is good for consumers to claim dependencies on the HostAddress trait.

This would be okay for some consumers, but not for those that want to support *some* regions that support `get_host_address` and some that don't. In those systems, using `get_host_address` at all would be ugly.

If we agree that `get_host_address` is constant across the lifetime of the region, vhost could still check for `get_host_address` support very early, and return an error if it fails.

jiangliu:
> > The proposed HostAddress trait explicitly states that the backend support mmapping guest memory into the vmm process, which is good for consumers to claim dependencies on the HostAddress trait.
> 
> This would be okay for some consumers, but not for those that want to support _some_ regions that support `get_host_address` and some that don't. In those systems, using `get_host_address` at all would be ugly.
> 
> If we agree that `get_host_address` is constant across the lifetime of the region, vhost could still check for `get_host_address` support very early, and return an error if it fails.

Deal, will update the PR:)

jiangliu:
> InvalidBackendAddress

Fixed, thanks:)

jiangliu:
> Just change to the right error and I will approve!

Fixed:)

PR #47, opened 2019-11-06
[Address] small improvements to the unit tests

serban300:
- move `Address` related unit tests to `address.rs`
- refactor and add more unit tests for `add` and `sub` operations
- add unit tests for `Address::new`, `BitAnd`, `BitOr`, `Default`

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -221,25 +221,72 @@ mod tests {
         assert_eq!(MockAddress(0x5055), a | 0x0005u64);
     }
 
-    #[test]
-    fn test_add_sub() {
-        let a = MockAddress(0x50);
-        let b = MockAddress(0x60);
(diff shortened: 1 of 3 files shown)


src/address.rs
```diff
@@ -176,3 +176,90 @@ macro_rules! impl_address_ops {
...
+        assert_ne!(a, c);
+        assert_ne!(c, a);
+    }
+
+    #[test]
+    #[allow(clippy::eq_op)]
+    fn test_cmp() {
+        let a = MockAddress(0x300);
+        let b = MockAddress(0x301);
+        assert!(a < b);
```
andreeaflorescu: This is a bit weird and clippy also complains about it. Let's just check that `a < b` instead.

serban300: Fixed.

PR #48, opened 2019-11-12
[GuestMemoryMmap] polish from_regions() method

serban300:
- polish `from_regions()` method
- add more unit tests for `GuestMemoryMmap` constructors

### DIFF
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -409,11 +409,11 @@ impl GuestMemoryMmap {
             return Err(Error::NoMemoryRegion);
         }
 
-        for i in 1..regions.len() {
-            let region = &regions[i];
-            let last = &regions[i - 1];
+        for window in regions.windows(2) {
+            let prev = &window[0];
+            let next = &window[1];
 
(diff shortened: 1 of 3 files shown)


src/mmap.rs
```diff
@@ -949,4 +1074,32 @@ mod tests {
         assert!(region.file_offset().is_some());
         assert_eq!(region.file_offset().unwrap().start(), offset);
     }
+
+    #[test]
```
bonzini: I am not sure this test adds much, it would fail even for a grammar fix.

serban300: Yes, that's true.

I can remove it, but then I would have to decrease the coverage with 0.1, because of these 1 added line that I cannot test:  `UnsortedMemoryRegions,`

bonzini: I think that's reasonable.

jiangliu: Coverage is just an hint, we needs useful code and tests:)

serban300: Ok. I removed the tests and decreased the coverage.

Issue #41, opened 2019-08-21
Improve performance of GuestMemoryMmap::find_region()


rbradford:
This function gets called a lot and appears in the performance reports when running cloud-hypervisor.

This currently has an O(n) algorithm and a simple fix I think would be to keep the vector of regions sorted and then use `Vec::binary_search_by` to give us an O(lg n) lookup. The added cost of sorting the regions when new ones get added would be minimal compared to the number of times `.find_region()` is called

rbradford:
cc @sboeuf @sameo

jiangliu:
Just curious, what scenarios make the find_region() as a hotspot?

bonzini:
I agree. The number of regions is usually small, but if there was a cut-off size below which binary search loses, it should be implemented in `Vec::binary_search_by`.

rbradford:
It appears in `perf top` traces when running under high network load. This because the the virtio-net code makes a lot of use of functionality to look into guest memory.

jiangliu:
Any plan to implement some cache policy into the find_region() hot path?

bonzini:
No plans to add caching, but the original issue with O(n) performance is fixed. Should this be closed? Or is `find_region()` still on the profiles?

jiangliu:
So let's close it first. We may open another if other performance hotspot appears later.

PR #49, opened 2019-11-16
Implemented Binary search to gain O(log n) lookup performance in find_region()

eminghuliev:
I have implemented Binary search in find_region which mentioned in #41 issue. There is UnsortedMemoryRegions mechanism to make sure sorting of memory regions during creating a new `GuestMemoryMmap` vector. That's why I just implemented binary search through sorted region map.

Signed-off-by: drmint80@gmail.com

### DIFF
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -447,14 +447,23 @@ impl GuestMemory for GuestMemoryMmap {
         self.regions.len()
     }
 
-    fn find_region(&self, addr: GuestAddress) -> Option<&GuestRegionMmap> {
-        for region in self.regions.iter() {
-            if addr >= region.start_addr() && addr <= region.end_addr() {
-                return Some(region);
-            }
-        }
-        None
-    }
+    fn find_region(&self, addr: GuestAddress) -> Option<&GuestRegionMmap> {           
+        let bs_region = self.regions.binary_search_by_key(&addr, |x| x.start_addr()); 
+        match bs_region {                                                             
+            Ok(x) => {                                                                
+                return Some(&self.regions[x]);                                        
+            }                                                                         
+            Err(x) if x > 0 => {                                                      
+                let size = addr.unchecked_sub(self.regions[x - 1].guest_base.0);      
+                if size.0 < self.regions[x - 1].len(){                                
+                    return Some(&self.regions[x - 1])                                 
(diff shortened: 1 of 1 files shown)


jiangliu:
I still have a question here. For most cases, we should have one or two memory ranges in an GuestMemory object, and the original search algorithm should be good enough. So what scenarios will have a relative more ranges in an GuestMemory object?

jiangliu:
Or any performance comparison with the previous linear search algorithm? Does it achieve good performance with 1-2 memory ranges only?

drmint80:
That's right it will give kinda performance when small range memory (even O(log n) in worst-case). As I know this project can be used for other purposes related to virtualization. If is there huge memory range for any other purposes, linear iteration can cause performance degradation.

jiangliu:
Thanks for the clarification:)

src/mmap.rs
```diff
@@ -447,14 +447,23 @@ impl GuestMemory for GuestMemoryMmap {
...
-    fn find_region(&self, addr: GuestAddress) -> Option<&GuestRegionMmap> {
-        for region in self.regions.iter() {
-            if addr >= region.start_addr() && addr <= region.end_addr() {
-                return Some(region);
-            }
-        }
-        None
-    }
+    fn find_region(&self, addr: GuestAddress) -> Option<&GuestRegionMmap> {           
+        let bs_region = self.regions.binary_search_by_key(&addr, |x| x.start_addr());
```
andreeaflorescu: This looks like it's changing the behavior of `find_region`. The function was previously returning the region that contains the `addr` passed as a parameter. Now it looks like it returns the region that starts with `addr`. Is this true?

drmint80: Return value is same with previous version. In previous version it just does iterate through regions. Now it returns just region with index (e.g regions[x]) which founded with BS.

drmint80: It does same thing still. In BS version, it searches exact match (starts with `addr`) if can't find exact value that starts with `addr` then searches nearest value (check `end_addr` in like previous version) in line 458.

andreeaflorescu: Ok, thanks. Let me check that again.

bonzini:
Closing to reopen and include a rebased version.

PR #50, opened 2019-12-04
[ByteValued] improve unit tests

serban300:
- Rename ByteValued macros for consistency
- improve ByteValued associated unit tests

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -192,17 +192,17 @@ pub trait Bytes<A> {
 }
 
 // All intrinsic types and arrays of intrinsic types are ByteValued. They are just numbers.
-macro_rules! array_data_init {
+macro_rules! byte_valued_array {
     ($T:ty, $($N:expr)+) => {
         $(
             unsafe impl ByteValued for [$T; $N] {}
         )+
     }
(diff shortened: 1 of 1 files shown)


bonzini:
I think PlainData is less clear than ByteValued (though that may not be a great name either); a ByteValued object can be sliced into bytes, while it's not clear what it means to be "plain" data. For example in C++ a "plain old data" type has no destructor, no assignment operator, etc. but it can have members of pointer type.

serban300:
> I think PlainData is less clear than ByteValued (though that may not be a great name either); a ByteValued object can be sliced into bytes, while it's not clear what it means to be "plain" data. For example in C++ a "plain old data" type has no destructor, no assignment operator, etc. but it can have members of pointer type.

Ok, Makes sense. I reverted the renamings.

bonzini:
@serban300 can you look into the CI failure so that we can merge the new tests? It seems like the buildkite configuration was broken.

andreeaflorescu:
@bonzini the Windows CI machine was not available. That's why the tests were failing. I re-triggered the Windows CI and now they're all passing.

PR #51, opened 2019-12-10
Fixes for CVE-2019-18960

sameo:
We want to be able to:

- Check that when adding an offset to a base guest address, we stay within a single region
- Check that a guest range stays within a single region before resolving to a host address

See https://github.com/firecracker-microvm/firecracker/commit/2c3e3728f51527a95c770df2413e82abbc206bbb

### DIFF
--- a/coverage_config.json
+++ b/coverage_config.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 78.9,
+  "coverage_score": 80.6,
   "exclude_path": "mmap_windows.rs",
   "crate_features": "backend-mmap"
 }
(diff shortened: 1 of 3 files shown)


bonzini:
I think this is the wrong approach. Instead, anything that wants to deal with ranges of guest addresses should use `try_access` or higher level `Bytes` or `VolatileMemory` methods. All of these provide correct range checking and slicing.

I will study more the firecracker CVE, but my gut feeling is that this should not happen at all if the vm-memory interfaces are used properly.

sameo:
@bonzini Firecracker does not use vm-memory yet, so I think they have different changes.
I looked a little deeper on what we're trying to solve here and I agree with you that things should be handled properly when calling into the corresponding vm-memory APIs. This is what we do from cloud-hypervisor so I believe this PR is not needed.
Closing it now.

PR #52, opened 2019-12-20
[GuestMemory] add examples for some methods

serban300:
Add examples for some methods declared in the Bytes trait:
- write_slice
- read_slice
- read_from
- write_to

Signed-off-by: Serban Iorga <seriorga@amazon.com>

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -21,7 +21,7 @@
 //! - [MemoryRegionAddress](struct.MemoryRegionAddress.html): represents an offset inside a region.
 //! - [GuestMemoryRegion](trait.GuestMemoryRegion.html): represent a continuous region of guest's
 //! physical memory.
-//! - [GuestMemory](trait.GuestMemroy.html): represent a collection of GuestMemoryRegion objects.
+//! - [GuestMemory](trait.GuestMemory.html): represent a collection of GuestMemoryRegion objects.
 //! The main responsibilities of the GuestMemory trait are:
 //!     - hide the detail of accessing guest's physical address.
 //!     - map a request address to a GuestMemoryRegion object and relay the request to it.
@@ -417,6 +417,25 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
         )
     }
 
+    /// # Examples
+    /// * Write a slice at guestaddress 0x200.
+    ///
+    /// ```
+    /// # #[cfg(feature = "backend-mmap")]
+    /// # use vm_memory::{Bytes, GuestAddress, mmap::GuestMemoryMmap};
+    ///
+    /// # #[cfg(feature = "backend-mmap")]
+    /// # fn test_write_u64() {
+    ///     let start_addr = GuestAddress(0x1000);
+    ///     let mut gm =
+    ///             GuestMemoryMmap::new(&vec![(start_addr, 0x400)]).expect("Could not create guest memory");
+    ///     let res = gm.write_slice(&[1, 2, 3, 4, 5], start_addr);
+    ///     assert!(res.is_ok());
+    /// # }
+    ///
+    /// # #[cfg(feature = "backend-mmap")]
+    /// # test_write_u64();
+    /// ```
     fn write_slice(&self, buf: &[u8], addr: GuestAddress) -> Result<()> {
         let res = self.write(buf, addr)?;
         if res != buf.len() {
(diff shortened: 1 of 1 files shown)


src/guest_memory.rs
```diff
@@ -386,6 +386,21 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
         )
     }
 
+    /// # Examples
```
aghecenco: Not related to your changes but can you fix the broken link (`GuestMemroy`) at line 24? (`[GuestMemory](trait.GuestMemroy.html)`)?

src/guest_memory.rs
```diff
@@ -386,6 +386,21 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
...
+    /// * Write a slice at guestaddress 0x200.
+    ///
+    /// ```
+    /// # #[cfg(feature = "backend-mmap")]
+    /// # fn test_write_u64() -> Result<(), ()> {
+    /// # use vm_memory::{Bytes, GuestAddress, mmap::GuestMemoryMmap};
```
aghecenco: Does the `use` need to live inside the `fn`?

andreeaflorescu: Let's not use fn in examples. It's easier to keep the code and examples in sync if we don't have functions because failures are caught by the CI. If you just return an Error when something goes wrong and don't check the error, failures go unnoticed.

src/guest_memory.rs
```diff
@@ -408,6 +439,26 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
...
 
+    /// # Examples
+    ///
+    /// * Read bytes from /dev/urandom
+    ///
+    /// ```
+    /// # use std::fs::File;
+    /// # use std::path::Path;
+    /// # #[cfg(feature = "backend-mmap")]
+    /// # fn test_read_random() -> Result<u32, ()> {
```
bonzini: You need to exclude this test from Windows builds.

src/guest_memory.rs
bonzini: This also applies to existing tests it would be nice if you could remove the use of `fn`, for example:

```diff
diff --git a/src/volatile_memory.rs b/src/volatile_memory.rs
index 1883c46..9178783 100644
--- a/src/volatile_memory.rs
+++ b/src/volatile_memory.rs
@@ -323,17 +323,14 @@ impl<'a> VolatileSlice<'a> {
     /// # use std::fs::File;
     /// # use std::path::Path;
     /// # use vm_memory::VolatileMemory;
-    /// # fn test_write_null() -> Result<(), ()> {
     /// let mut mem = [0u8; 32];
     /// let mem_ref = &mut mem[..];
-    /// let vslice = mem_ref.get_slice(0, 32).map_err(|_| ())?;
+    /// let vslice = mem_ref.get_slice(0, 32).unwrap();
     /// let mut buf = [5u8; 16];
     /// vslice.copy_to(&mut buf[..]);
     /// for v in &buf[..] {
     ///     assert_eq!(buf[0], 0);
     /// }
-    /// # Ok(())
-    /// # }
     /// ```
     pub fn copy_to<T>(&self, buf: &mut [T]) -> usize
     where
```

(but be careful with tests that try to open /dev/urandom).

serban300: I used fn in order to be able to specify `#[cfg(feature = "backend-mmap")]` since this is needed for using `GuestMemoryMmap` . I couldn't think of a better solution for this, but I did some changes.

src/guest_memory.rs
serban300: Done.

bonzini:
You're right that you have to keep the function. However `#[cfg(...)]` can also be applied to individual statements, so you can make the function return `()` and add at the bottom an `#[cfg(feature="backend-mmap")] test()` to ensure that tests run the function as requested by @aghecenco.

serban300:
> You're right that you have to keep the function. However `#[cfg(...)]` can also be applied to individual statements, so you can make the function return `()` and add at the bottom an `#[cfg(feature="backend-mmap")] test()` to ensure that tests run the function as requested by @aghecenco.

Done.

bonzini:
Can you mark as `#` the lines that are not strictly part of the example?

serban300:
> Can you mark as `#` the lines that are not strictly part of the example?

Sure. Done.

PR #53, opened 2019-12-20
A bunch of small changes, plus removing try_access

bonzini:
Hi all, this is a bunch of small changes, mostly cosmetic, that I had lying around. Most of the patches are small enough that it is probably unnecessary to create many pull requests for them.

The biggest is probably removing `try_access()`, in favor of an iterator-like object. Whether the code is nicer is in the eye of the beholder, I guess. I like the code of the users more, because it avoids a closure, but the code of the iterator is quite ugly as it often happens.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,6 @@ backend-mmap = []
 
 [dependencies]
 libc = ">=0.2.39"
-cast = ">=0"
 
 [target.'cfg(windows)'.dependencies.winapi]
 version = ">=0.3"
--- a/src/address.rs
+++ b/src/address.rs
@@ -19,7 +19,8 @@
 //! object and defines methods to access and manipulate it.
 
 use std::cmp::{Eq, Ord, PartialEq, PartialOrd};
-use std::ops::{Add, BitAnd, BitOr, Sub};
+use std::fmt::Debug;
+use std::ops::{Add, BitAnd, BitOr, Not, Sub};
 
 /// Simple helper trait used to store a raw address value.
 pub trait AddressValue {
@@ -29,10 +30,23 @@ pub trait AddressValue {
         + Eq
         + PartialOrd
         + Ord
+        + Not<Output = Self::V>
         + Add<Output = Self::V>
         + Sub<Output = Self::V>
         + BitAnd<Output = Self::V>
-        + BitOr<Output = Self::V>;
+        + BitOr<Output = Self::V>
+        + Debug
+        + From<u8>;
+
+    /// Return the value zero, coerced into the value type `Self::V`
+    fn zero() -> Self::V {
+        0u8.into()
+    }
+
+    /// Return the value zero, coerced into the value type `Self::V`
+    fn one() -> Self::V {
+        1u8.into()
+    }
 }
 
 /// Trait to represent an address within an address space.
@@ -87,6 +101,21 @@ pub trait Address:
     /// ```
     fn unchecked_offset_from(&self, base: Self) -> Self::V {
         self.raw_value() - base.raw_value()
+    }
+
+    /// Returns self, aligned to the given power of two.
+    fn checked_align_up(&self, power_of_two: Self::V) -> Option<Self> {
+        let mask = power_of_two - Self::one();
+        assert_ne!(power_of_two, Self::zero());
+        assert_eq!(power_of_two & mask, Self::zero());
+        self.checked_add(mask).map(|x| x & !mask)
+    }
+
+    /// Returns self, aligned to the given power of two.
+    /// Only use this when the result is guaranteed not to overflow.
+    fn unchecked_align_up(&self, power_of_two: Self::V) -> Self {
+        let mask = power_of_two - Self::one();
+        self.unchecked_add(mask) & !mask
     }
 
     /// Computes `self + other`, returning `None` if overflow occurred.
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -14,6 +14,7 @@ backend-mmap = []
 [dependencies]
 libc = ">=0.2.39"
 cast = ">=0"
+num-traits = ">=0"
```
andreeaflorescu: Can you add some details about why is this dependency needed?

src/guest_memory.rs
```diff
@@ -236,6 +237,68 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
...
+        Ok(range)
+    }
+
+    pub fn advance(&mut self, count: usize) {
+        assert!(count > 0 && count <= self._last_length.get());
+        self.total += count;
+        self.addr = match self.addr.overflowing_add(count as GuestUsize) {
+            (GuestAddress(0), _) => GuestAddress(0),
+            (result, false) => result,
+            (_, true) => panic!("guest address overflow"),
```
andreeaflorescu: Can you add a comment here to explain the need for the panic? I remember we discussed about this during KVM Forum as well. I generally don't expect panics in library code.

src/guest_memory.rs
```diff
@@ -236,6 +237,68 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
...
+    addr: GuestAddress,
+    count: usize,
+    total: usize,
+
+    // Stores the length of the current GuestMemoryRange to
+    // check for bugs.
+    _last_length: Cell<usize>,
+}
+
+impl<'a, M: GuestMemory + ?Sized> GuestMemoryRangeIter<'a, M> {
```
andreeaflorescu: Can you add comments to the public interface (pls pls pls)? Also some comments about what count and total are would be nice, it's not straightforward when reading the code the first time.

src/guest_memory.rs
```diff
@@ -494,15 +494,15 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
                 // It is safe to read from volatile memory. Accessing the guest
                 // memory as a slice is OK because nothing assumes another thread
                 // won't change what is loaded.
-                let bytes_written = dst.write(&src[start..end]).map_err(Error::IOError)?;
-                Ok(bytes_written)
+                dst.write_all(&src[start..end]).map_err(Error::IOError)?;
```
andreeaflorescu: I am not sure about this, but isn't the implementation of write different from the implementation of write_all? In some cases write_all is blocking while write is not blocking. Do we want this call to block until all bytes are written?

andreeaflorescu: nit: Unrelated to this line, but can you please re-work your commit message so you don't have to have the summary on two lines? "guest_memory: avoid multiple reads from the same range" should suffice as a summary.

Related to commit: 72b073aceadc72b6de0d2693c0235b3adc15e313

bonzini: I undid this change. However, write_all unfortunately must remain in the other branch, because reading could have side effects.

src/address.rs
```diff
@@ -86,6 +90,19 @@ pub trait Address:
         self.raw_value() - base.raw_value()
     }
 
+    /// Returns self, aligned to the given power of two.
+    fn checked_align_up(&self, power_of_two: Self::V) -> Option<Self> {
```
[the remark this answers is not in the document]
bonzini: Did both (which required adding `zero()` as well as `one()` as associated functions of `AddressValue`.

src/guest_memory.rs
bonzini: Done, in fact I made the whole of `GuestMemoryRangeIter` part of the public interface of `vm_memory`.

`total` is now named `so_far`.

src/guest_memory.rs
```diff
@@ -236,6 +237,108 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
...
+
+    /// The number of bytes for this range.
+    len: usize,
+}
+
+/// `GuestMemoryRangeIter` splits a memory range in contiguous regions
+/// (each represented by a `GuestMemoryRange`).  The caller can then
+/// process an entire range, or a part thereof, and proceed to the
+/// next one.
+pub struct GuestMemoryRangeIter<'a, M: GuestMemory + ?Sized> {
```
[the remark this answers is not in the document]
bonzini: Well, @andreeaflorescu didn't like the `try_access` code. :) For me the main advantage is actually that `GuestMemoryRangeIter` cannot be reimplemented/overridden. Also, while it is not a regular Rust `Iterator`, if you think it's desirable we could write both an `Iterator<Item=u8> and a `Read` implementation for `GuestMemoryRangeIter`.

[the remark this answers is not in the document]
andreeaflorescu: :(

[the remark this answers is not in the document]
alexandruag: Hi, apologies for the long silence :-s I'm still a bit more inclined towards keeping the `try_access` function as part of the interface, and not introducing a component like `GuestMemoryRangeIter` that cannot be overridden. I don't have a super clear picture in mind right now, but was definitely working towards that as part of an ongoing final pass over `vm-memory` before it gets published. Do you mind if we put the two iterator-related commits on hold and merge the rest of the PR in the interim? I can take care of rebasing, etc.

bonzini:
Sure, I will push the rest tomorrow

src/address.rs
```diff
@@ -86,6 +100,20 @@ pub trait Address:
...
+        let mask = power_of_two - Self::one();
+        assert_eq!(power_of_two & mask, Self::zero());
```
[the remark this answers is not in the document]
bonzini: Yes, why not.

bonzini:
@jiangliu can you approve?

andreeaflorescu:
Looks like this PR needs a manual rebase because the coverage also needs to be updated :(

PR #54, opened 2019-12-20
mmap: Protect each GuestMemoryMmap region with its own Arc

sboeuf:
The goal here is to ensure GuestMemoryMmap stays immutable, but that a
consuming layer on top of it could still use it to extend the effective
guest address space.

By moving the existing Arc from the outside of the Vec to the inside, we
make sure a GuestRegionMmap never gets released by mistake (which would
cause an unmap of the actual region), and we let the possibility for
each of these regions to be properly cloned into the new extended
instance of GuestMemoryMmap.

### DIFF
--- a/coverage_config.json
+++ b/coverage_config.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 78.9,
+  "coverage_score": 79.0,
   "exclude_path": "mmap_windows.rs",
   "crate_features": "backend-mmap"
 }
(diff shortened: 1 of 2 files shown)


sboeuf:
PR opened following the discussion [here](https://github.com/rust-vmm/vm-memory/pull/43#issuecomment-567877268) and [here](https://github.com/rust-vmm/vm-memory/pull/43#issuecomment-567910286).

sboeuf:
@bonzini I decided to simplify the PR by keeping the constructor the way it was working. There's actually no added value in having the one with the `Vec<Arc<GuestRegionMmap>>`, therefore I create each `Arc` from the initial `Vec<GuestRegionMmap>`.
This way, the PR really has minimal impact on the crate.

PR #55, opened 2019-12-23
mmap: implement Binary search in find_region()

bonzini:
This patch uses binary search to search in regions, as mentioned in issue #41;
the UnsortedMemoryRegions error already makes sure that memory regions are
sorted when a new GuestMemoryMmap is created.

Signed-off-by: Emin Ghuliev <drmint80@gmail.com>
Signed-off-by: Paolo Bonzini <pbonzini@redhat.com>

### DIFF
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -462,12 +462,13 @@ impl GuestMemory for GuestMemoryMmap {
     }
 
     fn find_region(&self, addr: GuestAddress) -> Option<&GuestRegionMmap> {
-        for region in self.regions.iter() {
-            if addr >= region.start_addr() && addr <= region.end_addr() {
-                return Some(region);
-            }
-        }
-        None
+        let index = match self.regions.binary_search_by_key(&addr, |x| x.start_addr()) {
+            Ok(x) => Some(x),
+            // Within the closest region with starting address < addr
+            Err(x) if (x > 0 && addr <= self.regions[x - 1].end_addr()) => Some(x - 1),
+            _ => None,
+        };
+        index.map(|x| self.regions[x].as_ref())
     }
 
     fn with_regions<F, E>(&self, cb: F) -> result::Result<(), E>


src/mmap.rs
```diff
@@ -448,10 +448,18 @@ impl GuestMemory for GuestMemoryMmap {
     }
 
     fn find_region(&self, addr: GuestAddress) -> Option<&GuestRegionMmap> {
-        for region in self.regions.iter() {
-            if addr >= region.start_addr() && addr <= region.end_addr() {
-                return Some(region);
+        let bs_region = self.regions.binary_search_by_key(&addr, |x| x.start_addr());
```
alexandruag: just a style proposal: I would merge this line into the following `match` by writing `match self.regions.binary_search_by_key(&addr, |x| x.start_addr()) { ... }`

src/mmap.rs
```diff
@@ -448,10 +448,18 @@ impl GuestMemory for GuestMemoryMmap {
...
+            Ok(x) => {
+                return Some(&self.regions[x]);
             }
+            Err(x) if x > 0 => {
+                let size = addr.unchecked_sub(self.regions[x - 1].guest_base.0);
+                if size.0 < self.regions[x - 1].len() {
+                    return Some(&self.regions[x - 1]);
+                }
+            }
+            _ => {}
```
alexandruag: just a nit: we could `return None` here and get rid of line 464.

src/mmap.rs
```diff
@@ -448,10 +448,18 @@ impl GuestMemory for GuestMemoryMmap {
...
     fn find_region(&self, addr: GuestAddress) -> Option<&GuestRegionMmap> {
-        for region in self.regions.iter() {
-            if addr >= region.start_addr() && addr <= region.end_addr() {
-                return Some(region);
+        let bs_region = self.regions.binary_search_by_key(&addr, |x| x.start_addr());
+        match bs_region {
+            Ok(x) => {
+                return Some(&self.regions[x]);
             }
+            Err(x) if x > 0 => {
```
alexandruag: We might be able to replace the following block with:

```rust
if self.regions[x - 1].end_addr() >= addr {
    return Some(self.regions[x - 1]);
}
```

Off-topic: this also kinda reminds me that `end_addr()` might not be most suggestive name for a method that returns the address of the last byte from the region (instead of that of the following byte). Do you think the name is actually ok, and in line with usual conventions (maybe `last_addr()` is more appropriate)?

bonzini: No, I don't&mdash;`last_addr` is definitely better.

src/mmap.rs
bonzini: Taking into account all of your suggestion, I rewrote everything as a match expression, at the cost of a slightly more complicated `if` clause.

PR #56, opened 2019-12-23
mmap: Introduce new constructor for extending GuestMemoryMmap

sboeuf:
In case the consumer of the API tries to extend an existing
GuestMemoryMmap with some new regions, it will need a special
constructor to accept a Vec<Arc<GuestRegionMmap>> instead of
Vec<GuestRegionMmap>. The existing regions coming from the existing
GuestMemoryMmap will come as a list of Arc<GuestRegionMmap>, which
is why the existing constructor cannot be reused in this case.

### DIFF
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -410,6 +410,22 @@ impl GuestMemoryMmap {
     ///               The regions shouldn't overlap and they should be sorted
     ///               by the starting address.
     pub fn from_regions(mut regions: Vec<GuestRegionMmap>) -> result::Result<Self, Error> {
+        Self::from_arc_regions(regions.drain(..).map(Arc::new).collect())
+    }
+
+    /// Creates a new `GuestMemoryMmap` from a vector of Arc regions.
+    ///
+    /// Similar to the constructor from_regions() as it returns a
+    /// GuestMemoryMmap. The need for this constructor is to provide a way for
+    /// consumer of this API to create a new GuestMemoryMmap based on existing
+    /// regions coming from an existing GuestMemoryMmap instance.
+    ///
+    /// # Arguments
+    ///
+    /// * `regions` - The vector of Arc regions.
+    ///               The regions shouldn't overlap and they should be sorted
+    ///               by the starting address.
+    pub fn from_arc_regions(regions: Vec<Arc<GuestRegionMmap>>) -> result::Result<Self, Error> {
         if regions.is_empty() {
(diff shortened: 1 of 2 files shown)


sboeuf:
@jiangliu @bonzini actually we need the new constructor, otherwise we won't be able to extend an existing list of regions.

alexandruag:
Hi everyone, it is ok if we wait and see where the design discussion started in #43 goes before adding more changes?

sboeuf:
@alexandruag well it'd be nice if we could merge this one since it's the logical follow up (which I should have included as part of the initial PR #54 that got merged).

bonzini:
Yes, it's useful because the owner might already have a `Vec<Arc<>>` and `GuestRegionMmap` is not cloneable.

@alexandruag - I think there are two parts in this discussion. The first is how to ensure that the user can recycle the same `GuestRegionMmap` across multiple `GuestMemoryMmap`s, which I think is useful functionality anyway. The is how to handle hotplug and more in general changes to the memory map (including e.g. PCI). This pull request is more on the side of the first part, so I'm more inclined to accept it already, but we can surely wait a bit longer.

sboeuf:
@alexandruag thank you for approving the PR, as it does indeed unblock things for `cloud-hypervisor` as we're about to add memory hotplug support.

src/mmap.rs
```diff
@@ -410,6 +410,22 @@ impl GuestMemoryMmap {
...
+    /// GuestMemoryMmap. The need for this constructor is to provide a way for
+    /// consumer of this API to create a new GuestMemoryMmap based on existing
+    /// regions coming from an existing GuestMemoryMmap instance.
+    ///
+    /// # Arguments
+    ///
+    /// * `regions` - The vector of Arc regions.
+    ///               The regions shouldn't overlap and they should be sorted
+    ///               by the starting address.
+    pub fn from_arc_regions(regions: Vec<Arc<GuestRegionMmap>>) -> result::Result<Self, Error> {
```
bonzini: Just one question, should this be `from_regions` or `from_arc_regions`? I am not familiar enough with Rust API design to know the answer.

sboeuf: What do you mean? I introduced a new constructor and since I was trying to keep backward compatibility, I didn't change the name of the original function `from_regions()`.

bonzini: Too much C++, Rust doesn't support function overloading. :)

PR #60, opened 2020-01-09
Rename guest memory `end_addr` to `last addr`

iulianbarbu:
`end_addr` is a method implemented both in `GuestMemoryRegion` and `GuestMemory`, which returns the maximum address contained by the memory region or guest memory respectively.

I suggest renaming `end_addr` to `last_addr`, and thus disambiguate what the method returns. 
In my opinion, the ambiguity comes from the vague `end` term, which is a noun, and does tangentially suggests that the address is at the end of the region/guest memory, but it is not clear enough that it refers to an included address.

I think a better word would be `last`, which is an adjective, and it does a better job in describing what kind of `addr` the method returns, namely the last address of a region/guest memory.

Signed-off-by: Iulian Barbu <iul@amazon.com>

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -164,7 +164,7 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
     fn start_addr(&self) -> GuestAddress;
 
     /// Get maximum (inclusive) address managed by the region.
-    fn end_addr(&self) -> GuestAddress {
+    fn last_addr(&self) -> GuestAddress {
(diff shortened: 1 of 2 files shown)


bonzini:
Approved, do we need to bump the version since this is an incompatible change?

andreeaflorescu:
@bonzini for the other crates, we bumped the version before publishing to crates.io. What do you think about treating this crate the same and bundle multiple changes when bumping the version?

bonzini:
@andreeaflorescu I agree&mdash;I asked just in case I was forgetting something obvious.

PR #61, opened 2020-01-16
Remove dependency on tempfile

### LINKED ISSUE #23: Remove dependency on tempfile
While tempfile is a nice utility that creates temp files which are automatically deleted when the object is out of scope, it is also a burden in terms of dependency.

Having the tempfile crate as a dependency adds 17 dependencies to the vm-memory crate. I propose to get rid of it and instead have a dummy wrapper that we can use in tests.

serban300:
fixes #23 

Posting this for early review. We need to wait for https://github.com/rust-vmm/vmm-sys-util/pull/70 to be merged and for a new version of vmm-sys-util to be published before actually merging this PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,3 +22,4 @@ features = ["errhandlingapi"]
 [dev-dependencies]
 matches = ">=0"
 tempfile = ">=3.0.2"
+vmm-sys-util = ">= 0.4.0"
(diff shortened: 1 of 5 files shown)


serban300:
A new version of `vmm-sys-util` that adds Windows support for `tempfile` has been published on `crates.io`. So this is ready to be merged.

@bonzini @jiangliu @alexandruag please review !

PR #64, opened 2020-01-27
Enhance GuestMemoryMmap to support memory hotplug

jiangliu:
This PR enhances the GuestMemory trait and introduce a new GuestMemoryMmapAtomic struct to support memory hotplug.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,10 +10,12 @@ edition = "2018"
 default = []
 integer-atomics = []
 backend-mmap = []
+backend-mmap-hotplug = ["arc-swap", "backend-mmap"]
 
 [dependencies]
 libc = ">=0.2.39"
 cast = ">=0"
+arc-swap = { version = ">=0.4.4", optional = true }
 
 [target.'cfg(windows)'.dependencies.winapi]
 version = ">=0.3"
--- a/coverage_config.json
+++ b/coverage_config.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 84.4,
+  "coverage_score": 83.9,
   "exclude_path": "mmap_windows.rs",
-  "crate_features": "backend-mmap"
+  "crate_features": "backend-mmap,backend-mmap-hotplug"
 }
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -31,7 +31,7 @@ use std::convert::From;
 use std::fmt::{self, Display};
 use std::fs::File;
 use std::io::{self, Read, Write};
-use std::ops::{BitAnd, BitOr};
+use std::ops::{BitAnd, BitOr, Deref};
 use std::sync::Arc;
 
 use crate::address::{Address, AddressValue};
@@ -234,6 +234,25 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
     }
 }
 
+/// The guard object returned by GuestMemory::snapshot().
+pub enum GuestMemoryGuard<'a, M: GuestMemory> {
+    /// The guard contains a GuestMemory object.
+    Valued(M),
+    /// The guard contains a reference to a GuestMemory object.
+    Ref(&'a M),
+}
+
+impl<'a, M: GuestMemory> Deref for GuestMemoryGuard<'a, M> {
+    type Target = M;
+
+    fn deref(&self) -> &Self::Target {
+        match self {
+            GuestMemoryGuard::Valued(ref v) => v,
+            GuestMemoryGuard::Ref(r) => *r,
+        }
+    }
+}
+
 /// Represents a container for a collection of GuestMemoryRegion objects.
 ///
 /// The main responsibilities of the GuestMemory trait are:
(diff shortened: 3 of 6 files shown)


jiangliu:
> I think this is not needed; using traits makes it much more complicated to handle lifetimes etc. Instead, we already have a trait that does everything we need, and that trait is `Bytes<GuestAddress>`. So for example:
> 
> ```rust
> pub struct GuestMemoryRwLock<M: GuestMemory> {
>     current: RwLock<M>,
> }
> 
> impl<M: GuestMemory> GuestMemoryRwLock<M> {
>     fn snapshot(&self) -> RwLockReadGuard<M> {
>         self.current.read().unwrap()
>     }
> }
> 
> impl<M: 'static + GuestMemory> Bytes<GuestAddress> for GuestMemoryRwLock<M> {
>     type E = Error;
> 
>     fn write(&self, buf: &[u8], addr: GuestAddress) -> Result<usize> {
>         self.snapshot().write(buf, addr)
>     }
> 
>     fn read(&self, buf: &mut [u8], addr: GuestAddress) -> Result<usize> {
>         self.snapshot().read(buf, addr)
>     }
> ```
> 
> and the same can be done for ArcSwap. We should just pick one of an RCU-like concurrency mechanism (either ArcSwap or crossbeam-epoch, probably?), and make it implement `insert_region`, `remove_region` and `Bytes<GuestAddress>`. I don't see a need to touch either `GuestMemory` or `GuestMemoryMmap`.

This design is simpler, but let's first talk about the design goals:) We may need to make choices before discussing about the implementation.
1) The interfaces exposed to GuestMemory clients, VirtioDevice<T: GuestMemory> vs VirtioDevice<T: Bytes<GuestAddress>>. 
To reuse code, we are planning to change 
```rust
pub trait VirtioDevice {
    /// Activates this device for real usage.
    fn activate(
        &mut self,
        mem: GuestMemory,
        ...
    ) -> ActivateResult;
}
```
to
```rust
pub trait VirtioDevice<M: GuestMemory> {
    /// Activates this device for real usage.
    fn activate(
        &mut self,
        mem: M,
        ...
    ) -> ActivateResult;
}
```
If `\<M: Byte<GuestAddress\>` is used instead of `\<M: GuestMemory\>`, the guest memory clients could only read/write data through guest_mem->read/write()->find_region()->read/write(). And if `\<M: GuestMemory\>` is used, the clients may optimize performance by cacheing the region smartly and skip redundant `guest_mem->read/write()->find_region()`.

2. Whether to hide the Guard<> behind the scene.
```rust
    /// Puts an available descriptor head into the used ring for use by the guest.
    pub fn add_used(&mut self, mem: &GuestMemory, desc_index: u16, len: u32) {
       ... 
        // These writes can't fail as we are guaranteed to be within the descriptor ring.
        mem.write_obj(u32::from(desc_index), used_elem).unwrap();
        mem.write_obj(len as u32, used_elem.unchecked_add(4)).unwrap();
        mem.write_obj(self.next_used.0 as u16, used_ring.unchecked_add(2)).unwrap();
    }
```
vs
```rust
    /// Puts an available descriptor head into the used ring for use by the guest.
    pub fn add_used(&mut self, mem: &GuestMemory, desc_index: u16, len: u32) {
       ... 

        let snapshot = mem.snapshot();

        // These writes can't fail as we are guaranteed to be within the descriptor ring.
        snapshot.write_obj(u32::from(desc_index), used_elem).unwrap();
        snapshot.write_obj(len as u32, used_elem.unchecked_add(4)).unwrap();
        snapshot.write_obj(self.next_used.0 as u16, used_ring.unchecked_add(2)).unwrap();
    }
```
Actually we prefer creating the snapshot at Device::IoEventHandler entrance, to avoid repeat RwLock::read()/ArcSwap::load() operations.

3. The way to protect GuestMemory::{num_regions, find_region, with_region, with_region_mut, map_and_fold} from memory hotplug.
In addition to methods of Byte<GuestAddress>, we also need to protect GuestMemory::{num_regions, find_region, with_region, with_region_mut, map_and_fold} . The most troublesome method is GuestMemory::find_region(), which returns a reference to an entry in the guest_memory object, thus need to handle lifetime issues. Actually it's the only method requires an GuestMemoryGuard<> object, all other methods could be implemented by hiding ArcSwap::load() internally.

So current implementation tries to achieve best performance with a complex snapshot()/GuestMemoryGuard<> abstraction.

jiangliu:
> Note 'underlying' not 'underline'

Thanks, we fix it in next version.

bonzini:
Hmm, I need to study this a lot more... Having both State and GuestMemoryGuard seems redundant to me, and I need to understand how the various States are created.

Don't put too much weight in my review - I marked it as change requested just to avoid that it's rushed in.

bonzini:
BTW, please don't push to rust-vmm/vm-memory itself; use a private fork instead.

jiangliu:
> BTW, please don't push to rust-vmm/vm-memory itself; use a private fork instead.

Sorry, my personal git remote repository has been misconfigured to rust/vm-memory when solving a github authentication issue. I have deleted those two branches created by me and will reopen another PR for this memory hotplug.

bonzini:
I pushed a reworked version of your patches at bonzini/hotplug. The idea is to use the type system (structs and traits) instead of enums and match.

PR #65, opened 2020-01-27
[GuestMemory] Add code examples

serban300:


bonzini:
@andreeaflorescu Do you know what's happening with buildkite/vm-memory-ci?

bonzini:
Hmm actually I missed that the test_map_fold() line is not commented. Not a huge deal but if you can push a fix for that and change the subject to "run the map_and_fold example as a test" it would be even better. With some luck that would wake up the CI too...

andreeaflorescu:
@bonzini It's kinda random. I'll take a look.

serban300:
> Hmm actually I missed that the test_map_fold() line is not commented. Not a huge deal but if you can push a fix for that and change the subject to "run the map_and_fold example as a test" it would be even better. With some luck that would wake up the CI too...

Sure. Done.

PR #68, opened 2020-01-30
Enhance GuestMemory to support memory hotplug 

bonzini:
Since @jiangliu closed #64, and this code is now documented and I think in decent shape, let's create a draft pull request to give it more exposure.

The main difference with #64 is that I am touching the core traits to allow:

* a generic implementation a memory map that wraps another `GuestMemory` M and adds RCU-style mutability (`GuestMemoryAtomic<M>`);

* no panics and no fighting the borrow checker for `find_region`, which makes no sense for a mutable memory map (see the `GuestMemoryMut`/`GuestMemory` split in the first patch)

* an abstraction of "snapshotting" a memory map that is zero-cost when the memory map is immutable (`GuestMemorySnapshottable<M>`). That is, whenever client code (e.g. vm-virtio) adds support for hotplug, it will not impose any penalty on VMMs whose memory map is immutable and passed around as e.g. `&'a GuestMemoryMmap`.

Even though it's very different from #64, it is definitely inspired by it and the concepts are entirely an evolution of what @jiangliu had proposed.

This is definitely pushing my Rust knowledge and confidence to the limits, so I'm looking forward to the reviews!

### DIFF
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -136,7 +136,7 @@ mod tests {
     use super::*;
     use crate::{
         GuestAddress, GuestMemory, GuestMemoryMmap, GuestMemoryRegion, GuestMemoryResult,
-        GuestUsize,
+        GuestRegionMmap, GuestUsize, MmapRegion,
     };
 
     type GuestMemoryMmapAtomic = GuestMemoryAtomic<GuestMemoryMmap>;
@@ -201,5 +201,48 @@ mod tests {
         assert_eq!(mem3.num_regions(), 2);
         assert!(mem3.find_region(GuestAddress(0x1000)).is_some());
         assert!(mem3.find_region(GuestAddress(0x10000)).is_none());
+    }
+
+    #[test]
+    fn test_atomic_hotplug() {
+        let region_size = 0x1000;
+        let regions = vec![
+            (GuestAddress(0x0), region_size),
+            (GuestAddress(0x10_0000), region_size),
+        ];
+        let mut gmm = Arc::new(GuestMemoryMmap::from_ranges(&regions).unwrap());
+        let gm: GuestMemoryAtomic<_> = gmm.clone().into();
+        let mem_orig = gm.memory();
+        assert_eq!(mem_orig.num_regions(), 2);
+
+        {
+            let guard = gm.lock().unwrap();
+            let new_gmm = Arc::make_mut(&mut gmm);
+            let mmap = Arc::new(
+                GuestRegionMmap::new(MmapRegion::new(0x1000).unwrap(), GuestAddress(0x8000))
+                    .unwrap(),
+            );
+            let new_gmm = new_gmm.insert_region(mmap).unwrap();
+            let mmap = Arc::new(
+                GuestRegionMmap::new(MmapRegion::new(0x1000).unwrap(), GuestAddress(0x4000))
+                    .unwrap(),
+            );
+            let new_gmm = new_gmm.insert_region(mmap).unwrap();
+            let mmap = Arc::new(
+                GuestRegionMmap::new(MmapRegion::new(0x1000).unwrap(), GuestAddress(0xc000))
+                    .unwrap(),
+            );
+            let new_gmm = new_gmm.insert_region(mmap).unwrap();
+            let mmap = Arc::new(
+                GuestRegionMmap::new(MmapRegion::new(0x1000).unwrap(), GuestAddress(0xc000))
+                    .unwrap(),
+            );
+            new_gmm.insert_region(mmap).unwrap_err();
+            guard.replace(new_gmm);
+        }
+
+        assert_eq!(mem_orig.num_regions(), 2);
+        let mem = gm.memory();
+        assert_eq!(mem.num_regions(), 5);
     }
 }
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -1,0 +1,205 @@
+// Copyright (C) 2019 Alibaba Cloud Computing. All rights reserved.
+// Copyright (C) 2020 Red Hat, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! A wrapper over an `ArcSwap<GuestMemory>` struct to support RCU-style mutability.
+//!
+//! With the `backend-atomic` feature enabled, simply replacing `GuestMemoryMmap`
+//! with `GuestMemoryAtomic<GuestMemoryMmap>` will enable support for mutable memory maps.
+//! To support mutable memory maps, devices will also need to use
+//! `GuestAddressSpace::memory()` to gain temporary access to guest memory.
+
+extern crate arc_swap;
+
+use arc_swap::{ArcSwap, Guard};
+use std::ops::Deref;
+use std::sync::{Arc, LockResult, Mutex, MutexGuard, PoisonError};
+
+use crate::{GuestAddressSpace, GuestMemory};
+
+/// A fast implementation of a mutable collection of memory regions.
+///
+/// This implementation uses ArcSwap to provide RCU-like snapshotting of the memory map:
+/// every update of the memory map creates a completely new GuestMemory object, and
+/// readers will not be blocked because the copies they retrieved will be collected once
+/// no one can access them anymore.  Under the assumption that updates to the memory map
+/// are rare, this allows a very efficient implementation of the `memory()` method.
+#[derive(Clone, Debug)]
+pub struct GuestMemoryAtomic<M: GuestMemory> {
+    // GuestAddressSpace<M>, which we want to implement, is basically a drop-in
+    // replacement for &M.  Therefore, we need to pass to devices the GuestMemoryAtomic
+    // rather than a reference to it.  To obtain this effect we wrap the actual fields
+    // of GuestMemoryAtomic with an Arc, and derive the Clone trait.  See the
+    // documentation for GuestAddressSpace for an example.
+    inner: Arc<(ArcSwap<M>, Mutex<()>)>,
+}
+
+impl<M: GuestMemory> From<Arc<M>> for GuestMemoryAtomic<M> {
+    /// create a new GuestMemoryAtomic object whose initial contents come from
+    /// the `map` reference counted GuestMemory.
+    fn from(map: Arc<M>) -> Self {
+        let inner = (ArcSwap::new(map), Mutex::new(()));
+        GuestMemoryAtomic {
+            inner: Arc::new(inner),
+        }
+    }
+}
+
+impl<M: GuestMemory> GuestMemoryAtomic<M> {
+    /// create a new GuestMemoryAtomic object whose initial contents come from
+    /// the `map` GuestMemory.
+    pub fn new(map: M) -> Self {
+        Arc::new(map).into()
+    }
+
+    fn load(&self) -> Guard<'static, Arc<M>> {
+        self.inner.0.load()
+    }
+
+    /// Acquires the update mutex for the GuestMemoryAtomic, blocking the current
+    /// thread until it is able to do so.  The returned RAII guard allows for
+    /// scoped unlock of the mutex (that is, the mutex will be unlocked when
+    /// the guard goes out of scope), and optionally also for replacing the
+    /// contents of the GuestMemoryAtomic when the lock is dropped.
+    pub fn lock(&self) -> LockResult<GuestMemoryExclusiveGuard<M>> {
+        match self.inner.1.lock() {
+            Ok(guard) => Ok(GuestMemoryExclusiveGuard {
+                parent: self,
+                _guard: guard,
+            }),
+            Err(err) => Err(PoisonError::new(GuestMemoryExclusiveGuard {
+                parent: self,
+                _guard: err.into_inner(),
+            })),
+        }
+    }
+}
+
+impl<M: GuestMemory> GuestAddressSpace for GuestMemoryAtomic<M> {
+    type T = GuestMemoryLoadGuard<M>;
+    type M = M;
+
+    fn memory(&self) -> Self::T {
+        GuestMemoryLoadGuard { guard: self.load() }
+    }
+}
+
+/// A guard that provides temporary access to a GuestMemoryAtomic.  This
+/// object is returned from the `memory()` method.  It dereference to
+/// a snapshot of the GuestMemory, so it can be used transparently to
+/// access memory.
+#[derive(Debug)]
+pub struct GuestMemoryLoadGuard<M: GuestMemory> {
+    guard: Guard<'static, Arc<M>>,
+}
+
+impl<M: GuestMemory> GuestMemoryLoadGuard<M> {
+    /// Make a clone of the held pointer and returns it.  This is more
+    /// expensive than just using the snapshot, but it allows to hold on
+    /// to the snapshot outside the scope of the guard.  It also allows
+    /// writers to proceed, so it is recommended if the reference must
+    /// be held for a long time (including for caching purposes).
+    pub fn into_inner(self) -> Arc<M> {
+        Guard::into_inner(self.guard)
+    }
+}
+
+impl<M: GuestMemory> Deref for GuestMemoryLoadGuard<M> {
+    type Target = M;
+
+    fn deref(&self) -> &Self::Target {
+        &*self.guard
+    }
+}
+
+/// An RAII implementation of a "scoped lock" for GuestMemoryAtomic.  When
+/// this structure is dropped (falls out of scope) the lock will be unlocked,
+/// possibly after updating the memory map represented by the
+/// GuestMemoryAtomic that created the guard.
+pub struct GuestMemoryExclusiveGuard<'a, M: GuestMemory> {
+    parent: &'a GuestMemoryAtomic<M>,
+    _guard: MutexGuard<'a, ()>,
+}
+
+impl<M: GuestMemory> GuestMemoryExclusiveGuard<'_, M> {
+    /// Replace the memory map in the GuestMemoryAtomic that created the guard
+    /// with the new memory map, `map`.  The lock is then dropped since this
+    /// method consumes the guard.
+    pub fn replace(self, map: M) {
+        self.parent.inner.0.store(Arc::new(map))
+    }
+}
+
+#[cfg(test)]
+#[cfg(feature = "backend-mmap")]
+mod tests {
+    use super::*;
+    use crate::{
+        GuestAddress, GuestMemory, GuestMemoryMmap, GuestMemoryRegion, GuestMemoryResult,
+        GuestUsize,
+    };
+
+    type GuestMemoryMmapAtomic = GuestMemoryAtomic<GuestMemoryMmap>;
+
+    #[test]
+    fn test_atomic_memory() {
+        let region_size = 0x400;
+        let regions = vec![
+            (GuestAddress(0x0), region_size),
+            (GuestAddress(0x1000), region_size),
+        ];
+        let mut iterated_regions = Vec::new();
+        let gmm = GuestMemoryMmap::from_ranges(&regions).unwrap();
+        let gm = GuestMemoryMmapAtomic::new(gmm);
+        let mem = gm.memory();
+
+        let res: GuestMemoryResult<()> = mem.with_regions(|_, region| {
+            assert_eq!(region.len(), region_size as GuestUsize);
+            Ok(())
+        });
+        assert!(res.is_ok());
+        let res: GuestMemoryResult<()> = mem.with_regions_mut(|_, region| {
+            iterated_regions.push((region.start_addr(), region.len() as usize));
+            Ok(())
+        });
+        assert!(res.is_ok());
+        assert_eq!(regions, iterated_regions);
+        assert_eq!(mem.num_regions(), 2);
+        assert!(mem.find_region(GuestAddress(0x1000)).is_some());
+        assert!(mem.find_region(GuestAddress(0x10000)).is_none());
+
+        assert!(regions
+            .iter()
+            .map(|x| (x.0, x.1))
+            .eq(iterated_regions.iter().map(|x| *x)));
+
+        let mem2 = mem.into_inner();
+        let res: GuestMemoryResult<()> = mem2.with_regions(|_, region| {
+            assert_eq!(region.len(), region_size as GuestUsize);
+            Ok(())
+        });
+        assert!(res.is_ok());
+        let res: GuestMemoryResult<()> = mem2.with_regions_mut(|_, _| Ok(()));
+        assert!(res.is_ok());
+        assert_eq!(mem2.num_regions(), 2);
+        assert!(mem2.find_region(GuestAddress(0x1000)).is_some());
+        assert!(mem2.find_region(GuestAddress(0x10000)).is_none());
+
+        assert!(regions
+            .iter()
+            .map(|x| (x.0, x.1))
+            .eq(iterated_regions.iter().map(|x| *x)));
+
+        let mem3 = mem2.memory();
+        let res: GuestMemoryResult<()> = mem3.with_regions(|_, region| {
+            assert_eq!(region.len(), region_size as GuestUsize);
+            Ok(())
+        });
+        assert!(res.is_ok());
+        let res: GuestMemoryResult<()> = mem3.with_regions_mut(|_, _| Ok(()));
+        assert!(res.is_ok());
+        assert_eq!(mem3.num_regions(), 2);
+        assert!(mem3.find_region(GuestAddress(0x1000)).is_some());
+        assert!(mem3.find_region(GuestAddress(0x10000)).is_none());
+    }
+}
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -26,12 +26,17 @@
 //!     - hide the detail of accessing guest's physical address.
 //!     - map a request address to a GuestMemoryRegion object and relay the request to it.
 //!     - handle cases where an access request spanning two or more GuestMemoryRegion objects.
+//!
+//! Whenever a collection of GuestMemoryRegion objects is mutable,
+//! [GuestAddressSpace](trait.GuestAddressSpace.html) should be implemented
+//! for clients to obtain a [GuestMemory] reference or smart pointer.
 
 use std::convert::From;
 use std::fmt::{self, Display};
 use std::fs::File;
 use std::io::{self, Read, Write};
-use std::ops::{BitAnd, BitOr};
+use std::ops::{BitAnd, BitOr, Deref};
+use std::rc::Rc;
 use std::sync::Arc;
 
 use crate::address::{Address, AddressValue};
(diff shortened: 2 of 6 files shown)


src/mmap.rs
```diff
@@ -444,6 +444,62 @@ impl GuestMemoryMmap {
...
+        // It shouldn't be a huge Vec, so prefer simple implementation.
+        for reg in self.regions.iter() {
+            if region.start_addr() >= reg.start_addr() && region.start_addr() <= reg.last_addr() {
+                return Err(Error::MemoryRegionOverlap);
+            } else if reg.start_addr() >= region.start_addr()
+                && reg.start_addr() <= region.last_addr()
+            {
+                return Err(Error::MemoryRegionOverlap);
+            }
+        }
```
serban300: Since some similar logic is also used in `GuestMemoryMmap::from_arc_regions()` I think it would be worth defining a method that checks if 2 regions overlap inside the `GuestMemoryRegion` trait.

rbradford:
Hi @bonzini this looks pretty good and is a more generalised version of the approach I took to implement hotplug in Cloud Hypervisor. I think the next step is probably to see if I can port our code over to use it as that will highlight any issues with this PR or with other bits of the tree (i'm kinda worried about linux-loader and some of our devices.)

bonzini:
Yes, the generalization ensures zero cost for immutable memory maps.

Should I separate the first patch into its own pull request already? Be sure to take a look  at the docs because they explain an associated type trick that can come in handy when porting vm-virtio to this extension.

src/guest_memory.rs
```diff
@@ -302,6 +306,96 @@ pub trait GuestMemoryMut {
...
+/// impl<'a> GuestMemoryBackend<'a> for ImmutableMmap {
+///     type Snapshot = GuestMemoryMmap;
+///     type Memory = &'a GuestMemoryMmap;
+/// }
+/// let mut for_immutable_mmap: VirtioDevice<ImmutableMmap> =
+///     VirtioDevice::new();
+/// let mmap = get_mmap();
+/// for_immutable_mmap.activate(&mmap);
+/// # }
+/// ```
```
jiangliu: In my original design, I used `M: GuestMemory` for both GuestMemoryBackend::Snapshot and GuestMemoryBackend::Memory.
The consideration is that:
1) for GuestMemory clients not caring about performance and only using stateless interfaces such as Bytes, it doesn't need to care about memory hotplug and snapshot().
2) for GuestMemory clients caring about performance or using stateful interfaces such as find_region/num_region/last_addr etc, it need to use the snapshot() method to get a stable GuestMemory snapshot.

With this proposed design, we need to teach all GuestMemory consumers about memory hotplug and snapshot(). But I'm on this side to clearly expose Memory/Snapshot concepts to vm-memory consumers.

And the changes to vm-virtio and virtio device drivers should be acceptable.

jiangliu:
> Yes, the generalization ensures zero cost for immutable memory maps.
> 
> Should I separate the first patch into its own pull request already? Be sure to take a look at the docs because they explain an associated type trick that can come in handy when porting vm-virtio to this extension.

Seems we are almost on the same page, any comments from @alexandruag ?
As both Cloud hypervisor and we are both working on memory hotplug, it would be great to merge all together:)

jiangliu:
We should be great to merge this part before publishing to crates.io

src/guest_memory.rs
```diff
@@ -309,6 +304,134 @@ pub trait GuestMemory {
...
+/// to repetitive types such as `VirtioDevice<GuestMemoryMmap, &GuestMemoryMmap>`.
+///
+/// ```
+/// # #[cfg(feature = "backend-mmap")]
+/// # use vm_memory::GuestMemoryMmap;
+/// # #[cfg(feature = "backend-atomic")]
+/// # use vm_memory::GuestMemoryAtomic;
+/// # use vm_memory::{GuestAddress, GuestMemory, GuestMemorySnapshottable};
+///
+/// # // Perhaps this could be added to vm-memory itself?
```
jiangliu: Definitely it should be included in vm-memory, because it's may be the main interfaces used by consumers such as vm-virtio.

src/mmap.rs
bonzini: Good idea.

src/guest_memory.rs
bonzini: > for GuestMemory clients not caring about performance and only using stateless interfaces such as Bytes, it doesn't need to care about memory hotplug and snapshot().

The main issue is that you could have a client caring about performance and potentially hotplug (`vm-virtio`), but it should not add any cost if the `GuestMemory` user does not need hotplug (like firecracker).

> we need to teach all GuestMemory consumers about memory hotplug and snapshot(). 

Not all of them. You can keep using `GuestMemory` just like now if you don't support hotplug. The need for changes only arises if you want to use `GuestMemoryMut` to support hotplug. Then you need to introduce `snapshot()`.

jiangliu: > > for GuestMemory clients not caring about performance and only using stateless interfaces such as Bytes, it doesn't need to care about memory hotplug and snapshot().
> 
> The main issue is that you could have a client caring about performance and potentially hotplug (`vm-virtio`), but it should not add any cost if the `GuestMemory` user does not need hotplug (like firecracker).
There are two cases here,
1) vmm supports memory hotplug or not. If the vmm doesn't support hotplug, it's easy to handle, current GuestMemory trait just works without code changes.
2) components (such as device drivers) support memory hotplug or not when the vmm supports memory hotplug.
What I'm thinking is about case 2.

> 
> > we need to teach all GuestMemory consumers about memory hotplug and snapshot().
> 
> Not all of them. You can keep using `GuestMemory` just like now if you don't support hotplug. The need for changes only arises if you want to use `GuestMemoryMut` to support hotplug. Then you need to introduce `snapshot()`.

The issue arises when sharing code among VMMs. For VMM specific implementations, we could just choose one of hotplug or non-hotplug interfaces. To share code, such as vm-device/vm-virtio, we need to teach them about hotplug/snapshot().

alexandruag:
> Seems we are almost on the same page, any comments from @alexandruag ?
> As both Cloud hypervisor and we are both working on memory hotplug, it would be great to merge all together:)

Sorry, I'm still trying to warp my head around a couple of things. Will post any comments by Monday EOD.

sboeuf:
@bonzini could you please rebase on top of master? Mainly you're missing the latest changes from https://github.com/rust-vmm/vm-memory/pull/70

src/guest_memory.rs
```diff
@@ -302,6 +306,96 @@ pub trait GuestMemoryMut {
...
+/// for_immutable_mmap.activate(&mmap);
+/// # }
+/// ```
+
+pub trait GuestMemorySnapshottable<M: GuestMemory> {
+    /// The type of the object returned by `snapshot()`.
+    type S: Deref<Target = M>;
+
+    /// Return a consistent snapshot of the memory map.
+    fn snapshot(&self) -> Self::S;
```
alexandruag: The biggest question in my mind right now is whether this method needs to be a first class citizen of the `GuestMemory` interface. I can think of two potential reasons:
- There are use cases where the region set needs to remain immutable over the course of multiple `GuestMemory` method invocations. Like I mentioned in the `GuestMemoryMut` comment, I'm really wondering why that would be necessary, so any examples are greatly appreciated.
- The cost of the `snapshot` operation is significant enough that we want/need to amortize it over multiple inner `GuestMemory` method calls whenever possible.

If either of the above is true, I think this needs to be a part of the main `GuestMemory` interface, because components such as device building blocks that are written without assuming a particular guest memory implementation need to use a uniform interface (I think one of Gerry's comments was also pointing this out). Otherwise, it can be abstracted within a wrapper implementation (such as `GuestMemoryAtomic`) of the regular `GuestMemory` interface.

bonzini: > The cost of the snapshot operation is significant enough that we want/need to amortize it over multiple inner GuestMemory method calls whenever possible.

It's not really the cost of snapshot(), it's the cost of find_region(). And if you want to cache the result of find_region(), you cannot use the do_in_region() approach and you need a guard. snapshot() provides the guard.

> If either of the above is true, I think this needs to be a part of the main GuestMemory interface

This method is a part of the main interface, though it's a bit roundabout---more precisely,  thanks to the generic implementation of `GuestMemorySnapshottable` it is a part of `&'a GuestMemory` and `Arc<GuestMemory>`, for both of which it can simply return `self.

src/guest_memory.rs
```diff
@@ -235,23 +238,11 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
...
-///
-/// The main responsibilities of the GuestMemory trait are:
-/// - hide the detail of accessing guest's physical address.
-/// - map a request address to a GuestMemoryRegion object and relay the request to it.
-/// - handle cases where an access request spanning two or more GuestMemoryRegion objects.
-///
-/// Note: all regions in a GuestMemory object must not intersect with each other.
-pub trait GuestMemory {
+/// All regions in a GuestMemoryMut object must not intersect with each other.
+pub trait GuestMemoryMut {
```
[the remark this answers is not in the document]
bonzini: > both traits allow interior mutability regarding the contents of the guest memory regions, but the difference comes from whether the set of regions can be mutated as well (or so I understand).

Yes, this is correct. Mutability of the contents is kind of implicit in the whole `VolatileMemory` concept, what matters here is the set of regions.

> that can be fixed by making find_region return a guard, or switching back to something like the old do_in_region which is based on closures, so nothing has to be leaked to the outside.

You're not very far from the design that I used here, actually. The guard is the `GuestMemory` returned by a `GuestMemoryMut`'s `snapshot()` method; when the `find_region` method is then invoked on the snapshot itself, the lifetime is limited to that of the snapshot (and therefore the snapshot acts as a guard).

> GuestMemoryMut can only guarantee the same over the course of a single method invocation, but why isn't the latter sufficient?

Basically because it prevents caching. In order to implement full-blown caching, something like a `GuestMemoryMut` observer should be added (also mutuated by QEMU's `MemoryListener` concept), but that would be separate because observers in Rust are tricky.

src/guest_memory.rs
```diff
@@ -302,6 +306,96 @@ pub trait GuestMemoryMut {
...
+///     type Memory = &'a GuestMemoryMmap;
+/// }
+/// let mut for_immutable_mmap: VirtioDevice<ImmutableMmap> =
+///     VirtioDevice::new();
+/// let mmap = get_mmap();
+/// for_immutable_mmap.activate(&mmap);
+/// # }
+/// ```
+
+pub trait GuestMemorySnapshottable<M: GuestMemory> {
```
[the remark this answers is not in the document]
bonzini: What about `GuestAddressSpace`, and renaming the method to `memory_map()`?

src/guest_memory.rs
bonzini: Ok, done.

bonzini:
> Also, I was wondering whether this approach is susceptible to some potential race conditions. For example, if a region is inserted, then an address from the new region is written to an old memory region, and an old snapshot is used to "dereference" the address we'll get an error :(

Yes, this is intended. The behavior of the guest is more or less undefined in that case, so it will be okay as long as the snapshots are used for as little time as possible. Otherwise, you'll get the behavior you mention plus a memory leak. If you do not want this, it's of course possible to implement a `GuestMemoryAtomic` using a Mutex or RwLock.

bonzini:
Regarding naming, in the discussion with Serban I proposed the following replacement:

* GuestMemorySnapshottable -> GuestAddressSpace

* `snapshot()` -> `memory_map()`

src/guest_memory.rs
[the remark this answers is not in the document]
jiangliu: When we are working with Solaris/Linux memory hotplug, we used reference count on pages and observer pattern to reclaim memory on demand. But the whole memory hotplug implementation is really complex, we have struggled with it for years, to fix subsystem/drivers without knowledge about memory hotplug. It's really a hard and long way.

For VMM memory usage patterns, I feel a snapshot mechanism should be a balanced tradeoff between complexity and performance. We only need to maintain region level reference count and address space level snapshots, and changes to GuestMemory consumers to support snapshot should be minor.

For cases which really cares about performance, ArcSwap cache could be used to minimize the   overhead of snapshot, near to zero:)

jiangliu:
> > Also, I was wondering whether this approach is susceptible to some potential race conditions. For example, if a region is inserted, then an address from the new region is written to an old memory region, and an old snapshot is used to "dereference" the address we'll get an error :(
> 
> Yes, this is intended. The behavior of the guest is more or less undefined in that case, so it will be okay as long as the snapshots are used for as little time as possible. Otherwise, you'll get the behavior you mention plus a memory leak. If you do not want this, it's of course possible to implement a `GuestMemoryAtomic` using a Mutex or RwLock.



> > Also, I was wondering whether this approach is susceptible to some potential race conditions. For example, if a region is inserted, then an address from the new region is written to an old memory region, and an old snapshot is used to "dereference" the address we'll get an error :(
> 
> Yes, this is intended. The behavior of the guest is more or less undefined in that case, so it will be okay as long as the snapshots are used for as little time as possible. Otherwise, you'll get the behavior you mention plus a memory leak. If you do not want this, it's of course possible to implement a `GuestMemoryAtomic` using a Mutex or RwLock.

If the memory hotplug controller is implemented correctly, this scenario won't happen. The work flow is something like:
1) Issue memory hotplug command through VMM API
2) VMM prepares new memory resource
3) VMM updates the GuestMemoryMut instance to include the hot-added memory regions with RCU-like mechanism like ArcSwap.
4) VMM waits for a RCU quiesce period to ensure all old snapshots have been released.
5) VMM generates memory hot-add events to guest OS.
6) Guest OS makes use of hot-added memory.
After step 4, all newly taken snapshot should include the hot-added memory.

jiangliu:
> Regarding naming, in the discussion with Serban I proposed the following replacement:
> 
> * GuestMemorySnapshottable -> GuestAddressSpace
> * `snapshot()` -> `memory_map()`

Naming is always hard:)
We have built another abstraction layer over GuestMemory, which already used the notation "AddressSpace", luckily it's not named "GuestAddressSpace".
Actually I have one question here, what's the role of GuestMemory? It's a accessor to access guest memory or it's a resource owner of guest memory. In my point of view, GuestMemory is just an accessor of guest memory, another abstraction of AddressSpace is the resource owner of guest memory. We could generate multiple different GuestMemory objects to access different parts of guest memory. For example, we could split guest memory guest kernel and non-guest-kernel memory, and build one GuestMemory instance for guest kernel and another GuestMemory instance for non-guest-kernel memory.

src/guest_memory.rs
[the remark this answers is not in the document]
jiangliu: > > that can be fixed by making find_region return a guard, or switching back to something like the old do_in_region which is based on closures, so nothing has to be leaked to the outside.
> 
> You're not very far from the design that I used here, actually. The guard is the `GuestMemory` returned by a `GuestMemoryMut`'s `snapshot()` method; when the `find_region` method is then invoked on the snapshot itself, the lifetime is limited to that of the snapshot (and therefore the snapshot acts as a guard).

Yes, the key idea is to return:
1) a reference for non-hotplug-capable guest memory implement to minimize penalty
2) a guard for hotplug-capable guest memory implementation to ensure correctness.
Thanks to rust's lifetime, it's hard to combine a reference and a guard as one type. Paolo and I are trying to solve the lifetime issue with different methods. For C, it's not an issue at all:(

With Paolo's proposal, a possible pattern for device driver may be something like:
```rust
struct DummyDevice<M: GuestMemoryBackend> {
     vm_memory: M::Memory,
}
DummyDevice<M: GuestMemory> {
      pub fn new(vm_memory: M::Memory {
             DummyDevice {
                        vm_memory
             }
      }
      pub fn handler(&self) {
            let snapshot = self.vm_memory.snapshot();
            self.process_request(snapshot);
      }
      fn process_request(&self, snapshot: M:Snapshot) {
      }
}
```
Basically we need to change existing device driver to this mode, but it should be trivial.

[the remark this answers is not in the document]
jiangliu: Too optimistic, we need to carries on an extra lifetime parameter for GuestMemoryBackend:(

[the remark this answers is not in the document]
bonzini: Yes, `'a` is still needed but it's much lighter-weight than a generic type parameter, and especially nested generic type parameters which you have to write as `<A, B<A>>`.

bonzini:
> what's the role of GuestMemory? It's a accessor to access guest memory or it's a resource owner of guest memory. In my point of view, GuestMemory is just an accessor of guest memory, another abstraction of AddressSpace is the resource owner of guest memory

Exactly: snapshotting is the mechanism through which the owner grants (time-limited) access to guest memory. When the implementation of GuestMemoryMut and GuestMemory is separate, then you need the GuestMemorySnapshottable trait (or GuestAddressSpace...) to go from one to the other. When it is the same, instead you use the generic implementations on `&'a M`/`Rc<M>`/`Arc<M>`; then access is granted to the client just by passing a reference (which has a well defined lifetime, possibly `'static`, and therefore doubles as a guard) or by cloning a reference-counted pointer (which keeps the resources alive).

jiangliu:
> > what's the role of GuestMemory? It's a accessor to access guest memory or it's a resource owner of guest memory. In my point of view, GuestMemory is just an accessor of guest memory, another abstraction of AddressSpace is the resource owner of guest memory
> 
> Exactly: snapshotting is the mechanism through which the owner grants (time-limited) access to guest memory. When the implementation of GuestMemoryMut and GuestMemory is separate, then you need the GuestMemorySnapshottable trait (or GuestAddressSpace...) to go from one to the other. When it is the same, instead you use the generic implementations on `&'a M`/`Rc<M>`/`Arc<M>`; then access is granted to the client just by passing a reference (which has a well defined lifetime, possibly `'static`, and therefore doubles as a guard) or by cloning a reference-counted pointer (which keeps the resources alive).

Great, we should be on the same page now. My only concern remaining is the extra lifetime parameter. The original design is heavily shaped by the requirement to avoid extra lifetime parameter.

bonzini:
Is it really important? The lifetime parameter is already present if the client simply stores an `& 'a GuestMemoryMmap`, and in the hotplug case it can be replaced by `'_` as in the MutableMmap example from the documentation.

jiangliu:
> Is it really important? The lifetime parameter is already present if the client simply stores an `& 'a GuestMemoryMmap`, and in the hotplug case it can be replaced by `'_` as in the MutableMmap example from the documentation.

I gave myself some more time to think about the overall abstraction:)
When I was starting to work on the vm-memory crate, I built the abstraction as :
AddressSpace as the resource owner and GuestMemory as the accessor, but the AddressSpace concept has been rejected. Then we have moved the AddressSpace into the vmm part.
Now we are talking the concept of AddressSpace again, I really like the concept to split resource owner from accessor:)

I suggest an abstraction as: AddressSpace -> GuestMemoryHandle -> GuestMemorySnapshot.
AddressSpace belongs to vmm, GuestMemoryHandle and GuestMemory belong to vm-memory.
With this abstraction, vmm is the resource owner. Vmm  creates handles and passes them to guest memory clients such as kernel loader/IO devices. Guest memory clients creates stable snapshot view from handles on demand.
Why introduces a layer of GuestMemoryHandle? We want to limit the portion of guest memory accessible to each guest memory clients. For example, we want to constraint the guest memory accessible by IO subsystem.

I have scanned the code with generalized <M: GuestMemory>, the majority usage of generic type parameter `M`  is `&M`, with several instance of Arc<M>. So to ease the code conversion, how about replacing GuestMemoryBackend with GuestMemoryHandle, and keep GuestMemory as is though it's actually GuestMemorySnapshot.

This abstraction should be flexible enough. It could support all of firecracker, cloud hypervisor and our dragonball VMM.

jiangliu:
> Is it really important? The lifetime parameter is already present if the client simply stores an `& 'a GuestMemoryMmap`, and in the hotplug case it can be replaced by `'_` as in the MutableMmap example from the documentation.

For most guest memory clients with generic type <M: GuestMemory>, they store `Arc\<M\>` instead of `&'a M` to avoid dealing with lifetime parameters.

bonzini:
> I suggest an abstraction as: AddressSpace -> GuestMemoryHandle -> GuestMemorySnapshot.

Essentially what you want is something more flexible than GuestMemoryMmap. The last two layers are essentially GuestMemorySnapshottable (now GuestAddressSpace) and GuestMemory. On top of that you want a hierarchy of memory regions similar to QEMU; you can build it in your VMM and wrap it with GuestMemoryAtomic to get the other two for free. It's orthogonal to this issue, and we can even add it to vm-memory under a separate feature.

I tried moving the associated types into GuestMemorySnapshottable but failed, so the backend trait is separate, but I think the system we have is quite usable.

jiangliu:
> > I suggest an abstraction as: AddressSpace -> GuestMemoryHandle -> GuestMemorySnapshot.
> 
> Essentially what you want is something more flexible than GuestMemoryMmap. The last two layers are essentially GuestMemorySnapshottable (now GuestAddressSpace) and GuestMemory. GuestMemoryBackend is a small wrapper around the former.
> 
> On top of that you want a hierarchy of memory regions similar to QEMU; you can build it in your VMM and wrap it with GuestMemoryAtomic to get the other two for free. It's orthogonal to this issue, and we can even add it to vm-memory under a separate feature.
> 
> I tried moving the associated types into GuestMemorySnapshottable but failed, so the backend trait is separate, but I think the system we have is quite usable.

Yes, I like the abstraction. And it could be easily integrated with our existing code.
One minor suggestion is not to call GuestMemoryBackend as GuestAddressSpace, something like GuestMemoryHandle would be welcomed:)

I also tried another solution to add a lifetime parameter to GuestMemory, the implementation is simple, but the abstraction is not as good as your proposal. 
So I hope we could merge this proposal soon:)

src/guest_memory.rs
```diff
@@ -336,6 +340,140 @@ pub trait GuestMemoryMut {
...
+/// # use vm_memory::{GuestMemory, GuestMemoryBackend, GuestMemoryMmap};
+/// struct MmapBackend;
+/// # #[cfg(feature = "backend-mmap")]
+/// impl<'a> GuestMemoryBackend<'a> for MmapBackend {
+///     type Memory = GuestMemoryMmap;
+///     type AddressSpace = &'a GuestMemoryMmap;
+/// }
+/// ```
+
+pub trait GuestMemoryBackend<'a> {
```
alexandruag: Hmm, why is this trait necessary/helpful in addition to `GuestAddressSpace`?  Is it mostly/mainly a solution to the discussion in #72? I noticed one of Paolo's comments saying some associated type(s) could not be moved in `GuestAddressSpace` so the backend remains separate; what was the desired outcome there?

src/guest_memory.rs
```diff
@@ -336,6 +340,140 @@ pub trait GuestMemoryMut {
...
+///     type AddressSpace = &'a GuestMemoryMmap;
+/// }
+/// let mut for_immutable_mmap: VirtioDevice<ImmutableMmap> =
+///     VirtioDevice::new();
+/// let mmap = get_mmap();
+/// for_immutable_mmap.activate(&mmap);
+/// # }
+/// ```
+
+pub trait GuestAddressSpace<M: GuestMemory> {
```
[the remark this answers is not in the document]
jiangliu: I'm working on this way, and it works with a limitation that GuestMemoryMmap will always used as Arc<GuestMemoryMmap>. I think it's a practically acceptable limitation though:)

> Hi again, thanks for the detailed discussions and explanations. Just posted a couple more questions from my side. I'm not super happy with the additional abstraction TBH, but I think we can build most reusable building blocks in a way that depends on `GuestMemory` alone and is agnostic to `GuestAddressSpace`, so it's going to be relatively straightforward to course correct if there's a better /more decoupled way of doing things at a later point.

Actually we could get rid of GuestMemoryMut now.

src/guest_memory.rs
jiangliu: I have worked out a patch upon Paolo's patches to get rid of GuestMemoryBackend/GuestMemoryBackendMT. GuestMemory and GuestAddressSpace should be enough.
By that way we only add an extra layer GuestAddressSpace over GuestMemory.

jiangliu:
>Does GuestMemoryMut still serve any purpose at this point? Seems the model we're heading >towards is a hierarchy where we have things that implement GuestMemory and then the new >AddressSpace concept on top, that provides Deref<Target=M:GuestMemory> guards which we use >to get GuestMemory "leases".

Yes, we could get rid of GuestMemoryMut now with the GuestAddressSpace abstraction.

jiangliu:
Hi @bonzini , I was happening to work on a patch to solve the @alexandruag 's review comments. I have pushed the patch to https://github.com/jiangliu/vm-memory/commit/d5b9b3d648e39e392f489b933142147e48983cc2 , which is based on your patch set.

As @alexandruag has mentioned, we could revert GuestMemoryMut related changes now, eventually we only need to GuestMemroyAddress abstraction to support memory hotplug.

bonzini:
Looks good! I'll try to merge both @jiangliu's refactoring and @alexandruag's other suggested changes (renaming, removing GuestMemoryMut).

bonzini:
> Note this cause a limitation that GuestMemoryMmap will always be used
as Arc<GuestMemoryMmap>, though it's a pratically acceptible limitaiton.

@jiangliu, can you expand on this? This origin of this limitation is not clear to me. It's not a big limitation since you can always `as_ref()` the `Arc`, but I would like to understand it better.

jiangliu:
> > Note this cause a limitation that GuestMemoryMmap will always be used
> > as Arc, though it's a pratically acceptible limitaiton.
> 
> @jiangliu, can you expand on this? This origin of this limitation is not clear to me. It's not a big limitation since you can always `as_ref()` the `Arc`, but I would like to understand it better.

/// # fn test_1() {
/// // Using `VirtioDevice` with an immutable GuestMemoryMmap:
/// let mut for_immutable_mmap: VirtioDevice<Arc<GuestMemoryMmap>> =
///     VirtioDevice::new();
/// let mmap = Arc::new(get_mmap());
/// for_immutable_mmap.activate(mmap.clone());
/// let mut another: VirtioDevice<Arc<GuestMemoryMmap>> =
///     VirtioDevice::new();
/// another.activate(mmap.clone());
/// # }

The test case fails if we use VirtioDevice<GuestMemoryMmap> instead of VirtioDevice<Arc<GuestMemoryMmap>> because GuestMemoryMmap doesn't implement GuestAddressSpace and it's not easy to implement GuestAddressSpace for GuestMemoryMmap too.
Luckily we have impl<M: GuestMemory> GuestAddressSpace for Arc<M> as Arc<GuestMemoryMmap> works.

bonzini:
Done, this is much nicer. Associated types everywhere, thanks Jiang and Alex for pushing and helping with the simplifications!

bonzini:
@jiangliu Do you have time for a proof of concept of switching vm-virtio to `GuestAddressSpace`? I think we need some kind of real-world testing before committing to the new model.

bonzini:
> The test case fails if we use `VirtioDevice<GuestMemoryMmap>`

Right, you should instead use `VirtioDevice<&GuestMemoryMmap>`. Remember that `GuestAddressSpace` is a replacement for a _reference_ to GuestMemory, not for the GuestMemory itself. So references still work.

jiangliu:
> @jiangliu Do you have time for a proof of concept of switching vm-virtio to `GuestAddressSpace`? I think we need some kind of real-world testing before committing to the new model.

Hi Paolo, I have converted one vm-virtio driver to use the GuestAddressSpace trait, the conversion is strait forward. There are two minor issues left:
1) whether we should mark GuestMemory/GuestAddressSpace as Send/Sync
2) whether we should mark GuestAddressSpace as Clone, we need some experience to choose among Arc<GuestAddressSpace> or GuestAddressSpace: Clone.
But this two minor issues could be handled separately later.

jiangliu:
> > The test case fails if we use `VirtioDevice<GuestMemoryMmap>`
> 
> Right, you should instead use `VirtioDevice<&GuestMemoryMmap>`. Remember that `GuestAddressSpace` is a replacement for a _reference_ to GuestMemory, not for the GuestMemory itself. So references still work.

I tried to make the hotplug/non-hotplug test cases as symmetric as possible, it's an exercise for code reuse:)

> > The test case fails if we use `VirtioDevice<GuestMemoryMmap>`
> 
> Right, you should instead use `VirtioDevice<&GuestMemoryMmap>`. Remember that `GuestAddressSpace` is a replacement for a _reference_ to GuestMemory, not for the GuestMemory itself. So references still work.

I just tried to make the hotplug/non-hotplug test cases as symmetric as possible, it's an exercise for code reuse:)

src/atomic.rs
```diff
@@ -0,0 +1,205 @@
...
+}
+
+impl<M: GuestMemory> GuestMemoryAtomic<M> {
+    /// create a new GuestMemoryAtomic object whose initial contents come from
+    /// the `map` GuestMemory.
+    pub fn new(map: M) -> Self {
+        Arc::new(map).into()
+    }
+
+    fn load(&self) -> Guard<'static, Arc<M>> {
```
sboeuf: Aren't we supposed to expose this one as public?

jiangliu: Method load() will be hide behind GuestAddressSpace::memory(). External clients do not need direct access to the load method.

bonzini: Right, this method only hides the ugly `self.inner.0.`. `memory()` returns the correct wrapper for `arc_swap::Guard`.

sboeuf: Yep I realized that after a few min :)

sboeuf:
@jiangliu @bonzini [here](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/762) is the code updated in Cloud-Hypervisor, based on this PR.

bonzini:
@jiangliu @sboeuf can you confirm it's just a testing method? Perhaps making it `pub(crate)` is better.

sboeuf:
@bonzini 
>can you confirm it's just a testing method? Perhaps making it pub(crate) is better.

Which method are you referring to?

bonzini:
Sorry, wrong pull request.

sboeuf:
Just curious, what are we waiting for to get this PR merged now that it seems everybody agreed on the design?

bonzini:
Code review from maintainers I suppose. :)

sboeuf:
@alexandruag @serban300 @andreeaflorescu PTAL :)

rbradford:
@sboeuf needs a rebase before it can be merged

PR #71, opened 2020-02-07
fix clippy warnings

serban300:


serban300:
> I think this need not use `_`. a_vec is accessed on the next line. Even better would be to implement iter() on VolatileArrayRef. For now I would leave this aside.

I missed this. Fixed.

src/mmap.rs
```diff
@@ -377,13 +377,19 @@ pub struct GuestMemoryMmap {
     regions: Vec<Arc<GuestRegionMmap>>,
 }
 
-impl GuestMemoryMmap {
-    /// Creates an empty `GuestMemoryMmap` instance.
-    pub fn new() -> Self {
+impl Default for GuestMemoryMmap {
```
jiangliu: How about #[derive(Default)]? I tried "#[derive(Default)]" today and seems it works.

serban300: Better. Done.

PR #72, opened 2020-02-07
Mark GuestMemory as Send + Sync

jiangliu:
Instances of GuestMemory will almost be accessed from multiple vCPU
threads and IO threads, so it should be Send + Sync. On the other hand,
it helps to simplify the conversion from
struct VirtioDeviceA {
	mem: GuestMemoryMmap,
}
to
struct VirtioDeviceA<M: GuestMemory> {
	mem: M,
}

Otherwise we need to add too much Send + Sync markers.

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -249,7 +249,7 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
 /// - handle cases where an access request spanning two or more GuestMemoryRegion objects.
 ///
 /// Note: the regions inside a [`GuestMemory`](trait.GuestMemory.html) object must not overlap.
-pub trait GuestMemory {
+pub trait GuestMemory: Send + Sync {
     /// Type of objects hosted by the address space.
     type R: GuestMemoryRegion;
 


bonzini:
Why is this useful? I can imagine a simple-minded implementation of hotplug using `RefCell<Rc<GuestMemoryMmap>>` rather than `ArcSwap<Arc<>>`, and Rust would tell us that it is not thread-safe _if that actually matters_.

Do you have examples of cases where clients had to specify `Send` and `Sync` bounds manually?

jiangliu:
> Why is this useful? I can imagine a simple-minded implementation of hotplug using `RefCell<Rc<GuestMemoryMmap>>` rather than `ArcSwap<Arc<>>`, and Rust would tell us that it is not thread-safe _if that actually matters_.
> 
> Do you have examples of cases where clients had to specify `Send` and `Sync` bounds manually?

It's not directly related to memory hotplug. Previously GuestMemory is a concrete type, now it's a trait. We encountered this issue when converting existing code to use vm-memory crate by using generic parameter as:
```rust
pub struct VirtioDeviceConfig<M: GuestMemory> {
    /// Guest memory accessor.
    pub mem: Arc<M>,
    ...
}
```
 To make VirtioDeviceConfig as Send + Sync, GuestMemory need to be Send + Sync, otherwise we will repeat "<M: GuestMemory + Send + Sync>" in many places.

bonzini:
> To make VirtioDeviceConfig as Send + Sync, GuestMemory need to be Send + Sync, otherwise we will repeat "<M: GuestMemory + Send + Sync>" in many places.

Can you explain why VirtioDeviceConfig needs to be Send+Sync? Is it because you're relying on ioeventfd to process I/O in a separate thread?

jiangliu:
> > To make VirtioDeviceConfig as Send + Sync, GuestMemory need to be Send + Sync, otherwise we will repeat "<M: GuestMemory + Send + Sync>" in many places.
> 
> Can you explain why VirtioDeviceConfig needs to be Send+Sync? Is it because you're relying on ioeventfd to process I/O in a separate thread?

Yes, we use IO worker threads to run epoll loops, so need to send device contexts to those worker threads.

bonzini:
I think we should instead use the trait that I had in the doc for the memory hotplug:

    trait GuestMemoryBackend<'a> {
       type Snapshot: GuestMemory;
       type Memory: 'a + GuestMemorySnapshottable<Self::Snapshot>;
    }

and add a Send+Sync variant:

    trait GuestMemoryBackendMT<'a> {
       type Snapshot: GuestMemory + Send + Sync;
       type Memory: 'a + GuestMemorySnapshottable<Self::Snapshot> + Send + Sync;
    }

alexandruag:
Hmm, seems like this may no longer apply after the changes in #68. Let's see where that goes first.

bonzini:
Could this be solved with a marker trait like

    trait GuestAddressSpaceMT {}
    impl<M> GuestAddressSpaceMT for M where M: GuestAddressSpace + Send + Sync {}

jiangliu:
> Could this be solved with a marker trait like
> 
> ```
> trait GuestAddressSpaceMT {}
> impl<M> GuestAddressSpaceMT for M where M: GuestAddressSpace + Send + Sync {}
> ```

Yes, we are exploring a similar tricky with:
pub trait PciSystemContext: Sync + Send + Clone {
    type I: InterruptManager + Send + Sync + Clone;
    type D: IoManagerContext + Send + Sync + Clone;

    fn get_device_manager_context(&self) -> Self::D;
    fn get_interrupt_manager(&self) -> Self::I;
}

which does help to simplify the dependency chain.

bonzini:
Closing for now. With #68 being merged and including a `GuestAddressSpace` implementation for `Rc<T>` I think we're considering non-thread-safe `GuestMemory` to be a valid choice.

PR #75, opened 2020-02-17
improve fds_overlap()

serban300:
- compare dev and inode instead of fd, in order to ensure uniqueness
- use less comparisons

### DIFF
--- a/coverage_config.json
+++ b/coverage_config.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 84.5,
+  "coverage_score": 84.2,
   "exclude_path": "mmap_windows.rs",
   "crate_features": "backend-mmap"
 }
--- a/src/mmap_unix.rs
+++ b/src/mmap_unix.rs
@@ -13,6 +13,7 @@
 use std::error;
 use std::fmt;
 use std::io;
+use std::os::unix::fs::MetadataExt;
 use std::os::unix::io::AsRawFd;
 use std::ptr::null_mut;
 use std::result;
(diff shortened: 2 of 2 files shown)


bonzini:
@jiangliu @sboeuf can you confirm that `fds_overlap` is just a testing method? Perhaps making it `pub(crate)` is better.

sboeuf:
@bonzini 
>can you confirm that fds_overlap is just a testing method? Perhaps making it pub(crate) is better.

I can confirm that only vm-memory uses this method for testing purpose. I'm just not sure if that's something the `mmap.rs` could rely on to perform some further checks when creating the regions.

Let's ask @alexandruag as I think he introduced this in the first place.

alexandruag:
Hi, the `fds_overlap` is there as a helper method in case it made sense to anyone to do the extra checks. If there are no performance/other concerns, I would consider including it as a mandatory verification going forward. If people are ok with that, the original (present) form allowed going around the restriction by `dup`-ing or aliasing `fd`s whenever overlap was desired for some reason; looks this is no longer possible with the new approach.

jiangliu:
> Hi, the `fds_overlap` is there as a helper method in case it made sense to anyone to do the extra checks. If there are no performance/other concerns, I would consider including it as a mandatory verification going forward. If people are ok with that, the original (present) form allowed going around the restriction by `dup`-ing or aliasing `fd`s whenever overlap was desired for some reason; looks this is no longer possible with the new approach.

It would be great to keep it as a helper without enforcing it on.

serban300:
Sorry for leaving this opened for so long, I completely forgot about this PR. I will close it since allowing to duplicate `fd`s is the desired behaviour.

Issue #62, opened 2020-01-21
Add guard pages to mmaped memory regions


serban300:
In theory it is possible to take a pointer to a mmaped memory region and read/write outside of the region's bounds.

We could prevent this behavior by adding a guard page at the beginning and at the end of each region.

serban300:
Fixed in Firecracker

PR #77, opened 2020-02-20
Surround memory mappings with guard pages

serban300:
implements #62 

In theory it is possible to take a pointer to a `MmapedRegion` and read/write outside of the region's bounds. We can try to prevent this behavior by surrounding the underlying mapped memory with guard pages.

Posting this as a draft in order to understand if this feature is of any interest to the community and to the maintainers of the crate.

### DIFF
--- a/src/mmap_unix.rs
+++ b/src/mmap_unix.rs
@@ -74,8 +74,12 @@ pub type Result<T> = result::Result<T, Error>;
 /// space size of the process.
 #[derive(Debug)]
 pub struct MmapRegion {
-    addr: *mut u8,
-    size: usize,
+    start_addr: *mut libc::c_void,
+    total_size: usize,
+
+    data_addr: *mut libc::c_void,
+    data_size: usize,
+
     file_offset: Option<FileOffset>,
     prot: i32,
     flags: i32,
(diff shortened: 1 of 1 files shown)


serban300:
@bonzini @jiangliu @alexandruag 

Sorry, just a quick ping. Can you take a look on this proposal ?

Thanks !

jiangliu:
> @bonzini @jiangliu @alexandruag
> 
> Sorry, just a quick ping. Can you take a look on this proposal ?
> 
> Thanks !

Will these red zones cause trouble to huge page mappings?

serban300:
> Will these red zones cause trouble to huge page mappings?

@jiangliu I don't think huge page mappings can lead to corner cases here. So I can't think of any trouble. I hope I'm not missing anything.

serban300:
Hi @bonzini, @jiangliu, @alexandruag,

When you have a few minutes to spare could you help me with a quick high-level opinion here ?

I would just like to understand if there is any interest in this feature or if you think that it would bring enough value compared to the added complexity in order to pursue it. This PR is only a draft, but if there's any interest I can spend some more time and try to clean up the implementation. If not, I can just close it.

Thanks,
Serban

glitzflitz:
Have you considered the case of jumping over the guard pages? Although the possibility is less due to correct bound checking but there were vulnerabilities related to this in linux kernel. Even though stack based arrays aren't widely used in rust but I'm not sure if rust has equivalent of GCC's `"-fstack-check"` option just for the extra precaution.
https://www.qualys.com/2017/06/19/stack-clash/stack-clash.txt
https://lkml.org/lkml/2017/6/22/345
One option is to enable configuring the number of pages that are used as guards.

src/mmap_unix.rs
```diff
@@ -154,25 +158,103 @@ impl MmapRegion {
...
+    ///
+    /// Returns a boolean indicating whether the operation succeeded.
+    ///
+    pub fn try_guard(&mut self) -> bool {
+        // Check if the `MmapRegion` is already guarded
+        if self.start_addr != self.data_addr && self.total_size != self.data_size {
+            return false;
+        }
+
+        let page_size = match unsafe { libc::sysconf(libc::_SC_PAGESIZE) } {
```
jiangliu, line 198: Should we align the redzone to hugepage size(2M/1G)?
Otherwise it may conflicts with THP.

jiangliu:
Could you please help add some unit tests for the new interface?

serban300:
@glitzflitz thanks for the comment. It's true that this protection doesn't cover all the scenarios. It's always possible to jump over the guard pages. But it makes it less likely to break out of the memory region especially by accident. I am not very familiar with `-fstack-check`. But these memory regions are allocated on the heap. So I'm not sure if it would help.

@jiangliu thank you for the review. As soon as I can I will take another look on this draft PR, try to simplify it if possible, and also address your comments.

alexandruag:
Hi, apologies for the delayed response. I'll have a better look at the code a bit later, but right now it seems like a less invasive option is to leverage the `build_raw` constructor for `MmapRegion` introduced in #109 (not published yet, but we can do that soon) + externally managed mappings. 

My current preference is to see what can be done in terms of simplications/performance improvements before changing how other things work (for example, if it makes sense to restrict the host addresses of the mapped regions to get contiguous areas, that might simplify using guard pages as well). Does the `build_raw`-based approach sound reasonable for the time being?

serban300:
@alexandruag Thanks for the suggestion. If I understand correctly `build_raw` looks like a good fit. I'll dive a bit deeper when I have time.

serban300:
Hey ! As suggested by @alexandruag after a new version of vm-memory is released, I will use `build_raw` in order to create custom mmap regions in Firecracker surrounded by guard pages. I will close this PR for the moment.

PR #78, opened 2020-03-03
fixed the crate manifest

andreeaflorescu:
The crate description is a required field of the manifest. This is
needed for publishing the crates on crates.io.

Added keywords as well to make this crate easy to discover.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,8 @@
 [package]
 name = "vm-memory"
 version = "0.1.0"
+description = "Safe abstractions for accessing the VM physical memory"
+keywords = ["memory", "virtual machine"]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,6 +1,8 @@
 [package]
 name = "vm-memory"
 version = "0.1.0"
+description = "Safe abstractions for accessing the VM physical memory"
+keywords = ["memory, "virtual machine"]
```
aghecenco:
```suggestion
keywords = ["memory", "virtual machine"]
```

andreeaflorescu: Auch, nice catch :))) I would've need a PR to fix a PR that fixes a broken manifest :)) sounds familiar?

Issue #76, opened 2020-02-19
Add `get_slice` and `as_volatile_slice` to GuestMemoryRegion


bonzini:
These would return a `Result<>`, similar to `get_host_address`, but the resulting object would be range-checked and allow to access guest memory without unsafe code.

PR #81, opened 2020-03-05
Add "get_slice" and "as_volatile_slice" methods to GuestMemoryRegion

michael2012z:
Fixes https://github.com/rust-vmm/vm-memory/issues/76.

I'm not experienced in vm-memory crate, this is my first commit here. Hope I didn't make too many mistakes. :) 

Ping @bonzini , could you take a look, is it the thing expected to be added in https://github.com/rust-vmm/vm-memory/issues/76 ? Please let me know if anything wrong.

BTW, does it make sense to add them in `GuestMemory` as well?

Signed-off-by: Michael Zhao <michael.zhao@arm.com>

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -246,6 +246,45 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
     unsafe fn as_mut_slice(&self) -> Option<&mut [u8]> {
         None
     }
+
+    /// Returns a [`VolatileSlice`](struct.VolatileSlice.html) of `count` bytes starting at
+    /// `offset`.
+    #[allow(unused_variables)]
+    fn get_slice(
+        &self,
+        offset: MemoryRegionAddress,
+        count: usize,
+    ) -> Result<volatile_memory::VolatileSlice> {
+        Err(Error::HostAddressNotAvailable)
+    }
+
+    /// Gets a slice of memory for the entire region that supports volatile access.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[cfg(feature = "backend-mmap")]
+    /// # use vm_memory::{GuestAddress, MmapRegion, GuestRegionMmap, GuestMemoryRegion};
+    /// # #[cfg(feature = "backend-mmap")]
+    /// # use vm_memory::volatile_memory::{VolatileMemory, VolatileSlice, VolatileRef};
+    ///
+    /// # #[cfg(feature = "backend-mmap")]
+    /// # fn test_as_volatile_slice() {
+    ///     let region =
+    ///         GuestRegionMmap::new(MmapRegion::new(0x400).unwrap(), GuestAddress(0x0))
+    ///         .unwrap();
+    ///     let slice = region.as_volatile_slice().unwrap();
+    ///     let v = 42u32;
+    ///     let r = slice.get_ref::<u32>(0x200).unwrap();
+    ///     r.store(v);
+    ///     assert_eq!(r.load(), v);
+    /// # }
+    /// # #[cfg(feature = "backend-mmap")]
+    /// # test_as_volatile_slice();
+    /// ```
+    fn as_volatile_slice(&self) -> Result<volatile_memory::VolatileSlice> {
+        self.get_slice(MemoryRegionAddress(0), self.len() as usize)
+    }
 }
 
 /// GuestAddressSpace provides a way to retrieve a GuestMemory object.
@@ -562,6 +601,18 @@ pub trait GuestMemory {
         self.to_region_addr(addr)
             .ok_or_else(|| Error::InvalidGuestAddress(addr))
             .and_then(|(r, addr)| r.get_host_address(addr))
+    }
+
+    /// Returns a [`VolatileSlice`](struct.VolatileSlice.html) of `count` bytes starting at
+    /// `addr`.
+    fn get_slice(
+        &self,
+        addr: GuestAddress,
+        count: usize,
+    ) -> Result<volatile_memory::VolatileSlice> {
+        self.to_region_addr(addr)
+            .ok_or_else(|| Error::InvalidGuestAddress(addr))
+            .and_then(|(r, addr)| r.get_slice(addr, count))
     }
 }
 
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -24,7 +24,7 @@ use crate::address::Address;
 use crate::guest_memory::{
     self, FileOffset, GuestAddress, GuestMemory, GuestMemoryRegion, GuestUsize, MemoryRegionAddress,
 };
-use crate::volatile_memory::VolatileMemory;
+use crate::volatile_memory::{VolatileMemory, VolatileSlice};
 use crate::Bytes;
 
 #[cfg(unix)]
@@ -165,6 +165,7 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
     fn write(&self, buf: &[u8], addr: MemoryRegionAddress) -> guest_memory::Result<usize> {
         let maddr = addr.raw_value() as usize;
         self.as_volatile_slice()
+            .unwrap()
             .write(buf, maddr)
             .map_err(Into::into)
     }
@@ -183,6 +184,7 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
     fn read(&self, buf: &mut [u8], addr: MemoryRegionAddress) -> guest_memory::Result<usize> {
         let maddr = addr.raw_value() as usize;
         self.as_volatile_slice()
+            .unwrap()
             .read(buf, maddr)
             .map_err(Into::into)
     }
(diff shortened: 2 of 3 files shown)


src/guest_memory.rs
```diff
@@ -246,6 +246,17 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
...
+    /// Returns a [`VolatileSlice`](struct.VolatileSlice.html) of `count` bytes starting at
+    /// `offset`.
+    fn get_slice(
+        &self,
+        offset: usize,
+        count: usize,
+    ) -> volatile_memory::Result<volatile_memory::VolatileSlice>;
+
+    /// Gets a slice of memory for the entire region that supports volatile access.
+    fn as_volatile_slice(&self) -> volatile_memory::VolatileSlice;
```
bonzini: Using `self.len()` you can implement this in GuestMemoryRegion directly.

src/guest_memory.rs
```diff
@@ -246,6 +246,17 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
     unsafe fn as_mut_slice(&self) -> Option<&mut [u8]> {
         None
     }
+
+    /// Returns a [`VolatileSlice`](struct.VolatileSlice.html) of `count` bytes starting at
+    /// `offset`.
+    fn get_slice(
```
bonzini: The default implementation could return HostAddressNotAvailable, like get_host_address.

michael2012z: There is a problem that way, `get_slice` needs to return `volatile_memory::Result`, while `HostAddressNotAvailable` doesn't match. We can return something like `Err(volatile_memory::Error::OutOfBounds { addr: offset })` instead, but the meaning is not precise. 

So, how about keeping it without default implementation?

bonzini: I think you should not return `volatile_memory::Result` but `guest_memory::Result`; I missed it in my original review. There is already a `From<>` implementation, so you can just write

```rust
    let slice = self.mapping.get_slice(offset, count)?
    Ok(slice)
```

and the `?` operator will take care of the conversion (unfortuantely you cannot just write `self.mapping.get_slice(offset, count).into()`, I think).

michael2012z: Thank you for detailed explanation and the example. Implementing `From` make the Result handling among different traits easy and graceful!
I've updated the code.

michael2012z:
Hi, @jiangliu , @alexandruag , could you help review this PR? Thank you.

bonzini:
@michael2012z please rebase because the coverage percentage might change, then @jiangliu can review

michael2012z:
Thanks. The PR was rebased.

src/guest_memory.rs
```diff
@@ -246,6 +246,22 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
...
+        &self,
+        _offset: MemoryRegionAddress,
+        _count: usize,
+    ) -> Result<volatile_memory::VolatileSlice> {
+        Err(Error::HostAddressNotAvailable)
+    }
+
+    /// Gets a slice of memory for the entire region that supports volatile access.
+    fn as_volatile_slice(&self) -> volatile_memory::VolatileSlice {
+        self.get_slice(MemoryRegionAddress(0), self.len() as usize)
```
andreeaflorescu: Not sure you can have this implementation in the trait because this unwrap being safe depends on the implementation of `get_slice`. Is there any other contract which makes this unwrap safe?

src/guest_memory.rs
```diff
@@ -563,6 +579,18 @@ pub trait GuestMemory {
...
+    /// Returns a [`VolatileSlice`](struct.VolatileSlice.html) of `count` bytes starting at
+    /// `addr`.
+    fn get_slice(
+        &self,
+        addr: GuestAddress,
+        count: usize,
+    ) -> Result<volatile_memory::VolatileSlice> {
+        self.find_region(addr)
+            .ok_or_else(|| Error::InvalidGuestAddress(addr))
+            .and_then(|r| r.get_slice(r.to_region_addr(addr).unwrap(), count))
```
andreeaflorescu: Please add a comment to specify why is this unwrap safe.

src/guest_memory.rs
```diff
@@ -246,6 +246,22 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
...
+    fn get_slice(
+        &self,
+        _offset: MemoryRegionAddress,
+        _count: usize,
+    ) -> Result<volatile_memory::VolatileSlice> {
+        Err(Error::HostAddressNotAvailable)
+    }
+
+    /// Gets a slice of memory for the entire region that supports volatile access.
+    fn as_volatile_slice(&self) -> volatile_memory::VolatileSlice {
```
andreeaflorescu: Please also add a doc comment to explain how you can use this functionality and why. Otherwise people might just miss it. Same for the other functions you are adding.

If it's not an overkill, I would also recommend adding an example.

src/guest_memory.rs
bonzini: You're right, the unwrap needs to be removed.

michael2012z: To remove the unwrap, do you mean changing the return type to `Result<volatile_memory::VolatileSlice>` ?

bonzini: Yes.

src/guest_memory.rs
bonzini: You  can actually use `GuestMemory::to_region_addr`, like

```rust
    self.to_region_addr(addr)
        .ok_or_else(|| Error::InvalidGuestAddress(addr))
        .and_then(|(r, addr)| r.get_slice(addr, count))
```

src/guest_memory.rs
bonzini: I would actually recommend adding _only_ an example, for example creating a VolatileRef<u32> from the VolatileSlice and storing 42 into it.

michael2012z: Cool! I will write this example to reveal the world's top secret.

src/mmap.rs
```diff
@@ -165,6 +165,7 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
     fn write(&self, buf: &[u8], addr: MemoryRegionAddress) -> guest_memory::Result<usize> {
         let maddr = addr.raw_value() as usize;
         self.as_volatile_slice()
+            .unwrap()
```
michael2012z: Hi, @bonzini , @andreeaflorescu . Please pay attention to the code update here.

In the code base (without this PR), reference of `GuestRegionMmap` has had `get_slice(`) and `as_volatile_slice()`, because `GuestRegionMmap` is de-referenced to `MmapRegion` (https://github.com/rust-vmm/vm-memory/blob/master/src/mmap.rs#L147).

In my first commit, I added `GuestMemoryRegion::as_volatile_slice()`, so what's called here became the new added function. There was no problem because the signature of the 2 `as_volatile_slice()`'s is identical. But when I tried to modify the return type of `GuestMemoryRegion::as_volatile_slice()`, the code here didn't build. That's why I modified this line.

bonzini: Yes, this is okay. The GuestRegionMmap knows that its as_volatile_slice implementation cannot fail.

src/guest_memory.rs
```diff
@@ -258,9 +258,32 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
...
+    ///         .unwrap();
+    ///     let slice = region.as_volatile_slice().unwrap();
+    ///     let v = 42u32;
+    ///     let ref_w = slice.get_ref::<u32>(0x200).unwrap();
+    ///     ref_w.store(v);
+    ///     let ref_r = slice.get_ref::<u32>(0x200).unwrap();
+    ///     assert_eq!(ref_r.load(), v);
+    /// }
+    /// # #[cfg(feature = "backend-mmap")]
+    /// test_as_volatile_slice();
```
bonzini: This line and the `}` line before need to be `#`-marked.

src/guest_memory.rs
```diff
@@ -258,9 +258,32 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
...
+    /// # #[cfg(feature = "backend-mmap")]
+    /// # fn test_as_volatile_slice() {
+    ///     let region =
+    ///         GuestRegionMmap::new(MmapRegion::new(0x400).unwrap(), GuestAddress(0x0))
+    ///         .unwrap();
+    ///     let slice = region.as_volatile_slice().unwrap();
+    ///     let v = 42u32;
+    ///     let ref_w = slice.get_ref::<u32>(0x200).unwrap();
+    ///     ref_w.store(v);
+    ///     let ref_r = slice.get_ref::<u32>(0x200).unwrap();
```
bonzini: You can use a single `get_ref` call for both storing and loading.

src/guest_memory.rs
```diff
@@ -246,6 +246,44 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
     unsafe fn as_mut_slice(&self) -> Option<&mut [u8]> {
         None
     }
+
+    /// Returns a [`VolatileSlice`](struct.VolatileSlice.html) of `count` bytes starting at
+    /// `offset`.
+    fn get_slice(
+        &self,
+        _offset: MemoryRegionAddress,
```
andreeaflorescu: nit: It is better to `#[allow(unused_variables)]` instead of prefixing parameters with `_` because these will be part of the documentation and it looks weird on docs.rs.

michael2012z: Yes, `_` makes weird variable names in docs, should be changed.
Please approve again. Thank you.

PR #82, opened 2020-03-09
Make result of GuestAddressSpace::memory() cloneable

bonzini:
I needed this in order to update vm-virtio for GuestAddressSpace support. Unfortunately this requires support in the arc-swap crate too.


bonzini:
CI expected to fail until https://github.com/vorner/arc-swap/pull/25 is pushed to crates.io.

PR #83, opened 2020-03-11
Provide access to Bytes<> for all ByteValued types

bonzini:
This allows using Bytes methods to read and write ByteValued types. See vmm-sys-util/issues#77 for the discussion.


bonzini:
@aghecenco Implementing `VolatileMemory` allowed writes even if you had only `&self` which is a bit too scary. I instead added an `as_bytes()` method directly on `ByteValued`, where the `&mut self` hides completely the volatility part.

PR #85, opened 2020-03-12
Update CI and enable coverage test on Aarch64

michael2012z:
Although vm-memory is not architecture specific, it is better to align
with the latest CI and have ARM coverage tested.

Signed-off-by: Michael Zhao <michael.zhao@arm.com>

### DIFF
--- a/coverage_config_aarch64.json
+++ b/coverage_config_aarch64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 84.8,
+  "coverage_score": 82.4,
   "exclude_path": "mmap_windows.rs",
   "crate_features": "backend-mmap,backend-atomic"
 }
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,0 +1,5 @@
+{
+  "coverage_score": 84.7,
+  "exclude_path": "mmap_windows.rs",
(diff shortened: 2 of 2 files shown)


coverage_config_aarch64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 84.8,
+  "coverage_score": 82.4,
```
andreeaflorescu: The big difference between aarch64 and x86_64 is very weird since the code is not at all different.

michael2012z: Also interesting is, after updating to new CI, X86 coverage dropped 0.1%.

andreeaflorescu: That is probably because of the Rust version update.

michael2012z: I compared the report generated on x86 and aarch64. It looks like there are some mistakes in Kcov report for the latter one.

michael2012z: Is this possible?

andreeaflorescu: Yeah, I was expecting that because in other crates we also get 1% difference at least even though the same code applies to both of them.

andreeaflorescu:
The rust-vmm-ci was updated as part of the security event #98 . Closing this.

PR #91, opened 2020-03-28
handle ErrorKind::Interrupted when doing file IOs

### LINKED ISSUE #90: Temparory EINTR error condition breaks vm_memory logic
According to rust doc for Read::read():
If this function encounters any form of I/O or other error, an error
variant will be returned. If an error is returned then it must be
guaranteed that no bytes were read.

An error of the [`ErrorKind::Interrupted`] kind is non-fatal and the
read operation should be retried if there is nothing else to do.

So we should silently ignore [`ErrorKind::Interrupted`] and retry,
otherwise it will false alarms/wrong results.

jiangliu:
Fix #90 

Quote from rust doc for Read::read():
If this function encounters any form of I/O or other error, an error
variant will be returned. If an error is returned then it must be
guaranteed that no bytes were read.

An error of the [`ErrorKind::Interrupted`] kind is non-fatal and the
read operation should be retried if there is nothing else to do.

So we should silently ignore [`ErrorKind::Interrupted`] and retry,
otherwise it will false alarms/wrong results.

### DIFF
--- a/coverage_config.json
+++ b/coverage_config.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 85.7,
+  "coverage_score": 84.9,
   "exclude_path": "mmap_windows.rs",
   "crate_features": "backend-mmap,backend-atomic"
 }
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -743,15 +743,27 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
                 // This is safe cause `start` and `len` are within the `region`.
                 let start = caddr.raw_value() as usize;
                 let end = start + len;
-                let bytes_read = src.read(&mut dst[start..end]).map_err(Error::IOError)?;
-                Ok(bytes_read)
+                loop {
+                    match src.read(&mut dst[start..end]) {
+                        Ok(n) => break Ok(n),
+                        Err(ref e) if e.kind() == std::io::ErrorKind::Interrupted => continue,
+                        Err(e) => break Err(Error::IOError(e)),
+                    }
+                }
             } else {
                 let len = std::cmp::min(len, MAX_ACCESS_CHUNK);
                 let mut buf = vec![0u8; len].into_boxed_slice();
-                let bytes_read = src.read(&mut buf[..]).map_err(Error::IOError)?;
-                let bytes_written = region.write(&buf[0..bytes_read], caddr)?;
-                assert_eq!(bytes_written, bytes_read);
-                Ok(bytes_read)
+                loop {
+                    match src.read(&mut buf[..]) {
+                        Ok(bytes_read) => {
+                            let bytes_written = region.write(&buf[0..bytes_read], caddr)?;
+                            assert_eq!(bytes_written, bytes_read);
+                            break Ok(bytes_read);
+                        }
+                        Err(ref e) if e.kind() == std::io::ErrorKind::Interrupted => continue,
+                        Err(e) => break Err(Error::IOError(e)),
+                    }
+                }
             }
         })
     }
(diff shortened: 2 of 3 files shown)


bonzini:
Looks good to me. I think that break with value is considered more idiomatic Rust than return in code like this, however.

jiangliu:
> Looks good to me. I think that break with value is considered more idiomatic Rust than return in code like this, however.

Changed to use "break with value":)

alexandruag:
Changes look good. WDYT about making things prettier by defining and using some helper functions (`read_uninterrupted`/`write_uniterrupted`)? Seems there's potential for code reuse; in this crate alone, the newly added blocks kinda repeat a couple of times. We can go even further and add those methods as part of traits automatically implemented for `T:Read/Write`.

jiangliu:
> Changes look good. WDYT about making things prettier by defining and using some helper functions (`read_uninterrupted`/`write_uniterrupted`)? Seems there's potential for code reuse; in this crate alone, the newly added blocks kinda repeat a couple of times. We can go even further and add those methods as part of traits automatically implemented for `T:Read/Write`.

The code is actually a generic pattern, and it's a common case to  ignore the interrupted error condition by accident. The question is where to host it, 
1) repeat it in each crate?
2) implement it as generic helper in vmm-sys-util?

bonzini:
The best place would be an associated function in IOError (taking a closure). I think for vm-memory Jiang's patch is okay.

alexandruag:
I think the best place would ideally be the std lib :D Anyway I kinda liked the auto traits idea, and quickly prototyped something [here](https://github.com/alexandruag/vm-memory/commit/f5783c3dc480723e79b0dd40780cb12d398d872b). If you are ok with it as well, we can keep it like that for now and move it someplace else later (I'll also add some tests). Otherwise I guess it's fine to keep the current PR as is, I just thought the changes are a little too verbose.

bonzini:
> I think the best place would ideally be the std lib :D

Well yeah that's what i meant by "an associated function in IOError" (actually `io::Error`).  But we could also write our own helper function (without a trait)

```rust
    // fn<T> uninterrupted(f: FnMut() -> io::Result<T>) -> io::Result<T> { ... }
    uninterrupted(|| src.read(dst))
        .map_err(Error::IOError)
```

jiangliu:
> I think the best place would ideally be the std lib :D Anyway I kinda liked the auto traits idea, and quickly prototyped something [here](https://github.com/alexandruag/vm-memory/commit/f5783c3dc480723e79b0dd40780cb12d398d872b). If you are ok with it as well, we can keep it like that for now and move it someplace else later (I'll also add some tests). Otherwise I guess it's fine to keep the current PR as is, I just thought the changes are a little too verbose.

Great. Then we could merge this PR first, then a following-up PR from you?

alexandruag:
Sure! Regarding Paolo's comment, I think things look a bit cleaner if we don't work with closures, and simply have a function (inside or outside a trait; the latter would take the reader/writer as a parameter) that behaves like `read/write`, but absorbs the `EINTR`s. I'll try to do a PR as well, and keep the helpers local just so we don't add another dependency. I also wonder why something like this isn't in the standard library already, because it seems like a common use case :-s

jiangliu:
> Sure! Regarding Paolo's comment, I think things look a bit cleaner if we don't work with closures, and simply have a function (inside or outside a trait; the latter would take the reader/writer as a parameter) that behaves like `read/write`, but absorbs the `EINTR`s. I'll try to do a PR as well, and keep the helpers local just so we don't add another dependency. I also wonder why something like this isn't in the standard library already, because it seems like a common use case :-s

Thanks for handling this.
It's really a error prone design, the std lib should be extended:)

PR #103, opened 2020-07-16
Add basic performance tests

lauralt:
This PR adds some basic performance tests for read/write latencies on GuestMemoryMmap.


alexandruag:
Hi folks, does anyone mind if we merge this? The tests are completely separated from the rest of the code, and I think they are a good starting point to do some basic performance comparisons (manually for now). After more tweaking and finding a formula with an acceptable overall duration per run, we'd like have something in the CI that detects regressions automatically.

jiangliu:
Nice job!
It's so happy to see the benchmark result:)

Issue #102, opened 2020-07-15
Iron out the GuestMemory interface


alexandruag:
We should clarify some aspects of the `GuestMemory` interface, and a good starting point is around atomicity semantics. As Gerry mentioned in [this](https://github.com/rust-vmm/vm-memory/issues/100#issuecomment-657947883) comment, there are a couple of ways to do an atomic access, but I think all of them have some downsides right now. 

`get_atomic_ref/array_ref` are a bit unwieldy, but more important, they're only available when the address falls within a guest region that's backed by host memory. This means that users of generic `M: GuestMemory` objects must either add a fallback code path, or rely exclusively on `read/write_obj`.

The main issue with `read/write_obj` is they don't have well-defined semantics at the interface level. One simple contract we could go for is *"`read/write_obj` provide atomic accesses for certain sizes at aligned memory locations"* (with the caveat that different archs provide atomicity guarantees up to different widths). A more verbose (but clearer and overall safer IMO) alternative is to expand the interface around `read/write_obj` to something like the following pseudocode:

```rust
// A type (easily convertible to/from GuestAddress) which guarantees the
// inner value is aligned with respect to T.
pub struct AlignedGuestAddress<T>(/*...*/);

// Just a marker.
pub unsafe trait AtomicAccess: ByteValued { }

impl GuestMemory {
    // ...

    // Always atomic!
    fn read_atomic<T: AtomicAccess>(addr: AlignedGuestAddress<T>) -> Result<T>;

    // The next two methods are not guaranteed to be atomic.
    fn read_obj<T: ByteValued>(addr: AlignedGuestAddress<T>) -> Result<T>;

    fn read_obj_unaligned<T: ByteValued>(addr: GuestAddress) -> Result<T>;

    // Similar for write ...
}

```

I think adding something like `AlignedGuestAddress` is a useful abstraction, because many implementations (i.e. device model building blocks) provide/demand alignment guarantees for certain addresses, and it makes sense to explicitly leverage the additional information or model the extra requirement. It also helps with making operation semantics more explicit.

WDYT?

PR #104, opened 2020-07-19
Add explicit atomic (and aligned) operations

alexandruag:
Related to #102

Hi,

This PR adds four new methods (`read_atomic`, `write_atomic`, `read_aligned`, and `write_aligned`) to the `GuestMemory` and `GuestMemoryRegion` interfaces. Here are some details:

- The proposal is mainly about adding `read_atomic` and `write_atomic` as part of the `GuestMemory` interface in order to provide a more straightforward (and always available) mechanism for performing atomic accesses with explicit semantics.

- The `Aligned<Addr, T>` abstraction appears interesting and useful to represent additional alignment information and/or requirements about an address. It needs some more methods to improve the ergonomics of casting + offsetting, together with examples, but for the time being I am really curious if others also think this is worth pursuing.

- For now, the new methods are added directly to `GuestMemory` and `GuestMemoryRegion` (as opposed to being part of the `Bytes` interface) because `Bytes` provides no guarantees about the alignment of the container (whereas both `GuestMemory` and `GuestMemoryRegion` are expected to be aligned to page boundaries -- maybe we should mention this explicitly btw). There are a couple of ways to reconcile the interfaces going forward, but I think what matters for now is to have a well-defined and straightforward way of performing atomic accesses based on a generic `M: GuestMemory` object.

- I've also ran a couple of synthetic tests, and the improvement for atomic reads/writes seems quite promising (i.e. > 2x). I'll try to do some more after I manage to get a better setup going.

### DIFF
--- a/src/align.rs
+++ b/src/align.rs
@@ -2,6 +2,7 @@
 //! of addresses by leveraging the type system.
 
 use std::convert::TryFrom;
+use std::fmt::{self, Debug, Formatter};
 use std::marker::PhantomData;
 use std::mem::align_of;
 use std::result;
@@ -50,6 +51,12 @@ impl<Addr: Clone, T> Clone for Aligned<Addr, T> {
 }
 
 impl<Addr: Copy, T> Copy for Aligned<Addr, T> {}
+
+impl<Addr: Debug, T> Debug for Aligned<Addr, T> {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
+        write!(f, "Aligned({:?}, {})", self.addr, align_of::<T>())
+    }
+}
 
 impl<Addr, T> Aligned<Addr, T> {
     /// Instantiate a new `Aligned` value without checking the alignment.
@@ -120,6 +127,83 @@ impl<Addr: Address, T> Aligned<Addr, T> {
             Ok(unsafe { Aligned::new(addr) })
         } else {
             Aligned::from_addr(addr)
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::GuestAddress;
+
+    use super::*;
+
+    #[test]
+    fn test_check_addr_aligned() {
+        assert!(check_addr_aligned::<_, u64>(GuestAddress(0x8)).is_ok());
+        assert_eq!(
+            check_addr_aligned::<_, u64>(GuestAddress(0x107)).unwrap_err(),
+            AlignmentError::Misaligned
+        );
+        assert_eq!(
+            check_addr_aligned::<_, u64>(GuestAddress(std::u64::MAX)).unwrap_err(),
+            AlignmentError::Overflow
+        );
+    }
+
+    #[test]
+    fn test_aligned() {
+        let guest_addr = GuestAddress(0x100);
+        let odd_guest_addr = GuestAddress(0x101);
+
+        let a = unsafe { Aligned::<_, u32>::new(guest_addr) };
+        assert_eq!(a.addr, guest_addr);
+        assert_eq!(a.into(), guest_addr);
+
+        {
+            let b = a.clone();
+            assert_eq!(a.addr, b.addr);
+        }
+
+        {
+            let b = unsafe { a.reinterpret::<i32>() };
+            assert_eq!(a.addr, b.addr);
+        }
+
+        {
+            let b = Aligned::<_, u32>::from_addr(guest_addr).unwrap();
+            assert_eq!(b.addr, guest_addr);
+
+            assert_eq!(
+                Aligned::<_, u32>::from_addr(odd_guest_addr).unwrap_err(),
+                AlignmentError::Misaligned
+            );
+        }
+
+        {
+            let b = a.cast::<u16>().unwrap();
+            assert_eq!(a.addr, b.addr);
+            // This is ok because, even though `a` is an `Aligned<GuestAddress, u32>`, the
+            // inner `addr` value is aligned w.r.t. `u64` as well.
+            assert!(a.cast::<u64>().is_ok());
+
+            // Let's also try out an `Aligned<GuestAddress, u32>` that's no longer
+            // aligned for `u64`.
+            let c = Aligned::<_, u32>::from_addr(GuestAddress(0x4)).unwrap();
+            assert_eq!(c.cast::<u64>().unwrap_err(), AlignmentError::Misaligned);
+        }
+
+        {
+            let u32_offset = 4;
+            let b = a.offset::<u32>(u32_offset).unwrap();
+            assert_eq!(b.addr, a.addr.unchecked_add(u32_offset));
+
+            assert_eq!(
+                a.offset::<u64>(u32_offset).unwrap_err(),
+                AlignmentError::Misaligned
+            );
+
+            let c = b.offset::<u64>(u32_offset).unwrap();
+            assert_eq!(c.addr, b.addr.unchecked_add(u32_offset));
         }
     }
 }
--- a/src/align.rs
+++ b/src/align.rs
@@ -1,0 +1,113 @@
+//! Provides abstractions (often effectively zero-cost) for modeling the alignment
+//! of addresses by leveraging the type system.
+
+use std::convert::TryFrom;
+use std::marker::PhantomData;
+use std::mem::align_of;
+use std::result;
+
+use crate::Address;
+
+/// Errors related to operations which involve aligned addresses.
+#[derive(Debug, PartialEq)]
+pub enum AlignmentError {
+    /// Misalignment was detected during a conversion.
+    Misaligned,
+    /// An overflow occurred while computing address values.
+    Overflow,
+}
+
+// Check whether `addr` is aligned with respect to `T`.
+fn check_addr_aligned<Addr: Address, T>(addr: Addr) -> result::Result<(), AlignmentError> {
+    // The value returned from `align_of` should fit within an `u8`
+    // for the foreseeable future.
+    let align = u8::try_from(align_of::<T>()).expect("Unexpected align_of value.");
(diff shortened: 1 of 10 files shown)


bonzini:
What is the specific reason for removing the automatic implementation of Bytes? That is, which methods are meant to be overridden by different GuestMemory implementation?

Regarding read/write_atomic, I would prefer to have an ordering argument and use atomics instead of volatile accesses, but the overall concept is great. I think we should add an Error for misaligned containers so that the methods can be added directly to Bytes. Can you also document  how read/write_atomic compare to VolatileAtomicRef? Is it just different ergonomics just like VolatileRef vs.read/write_obj?

alexandruag:
Hi Paolo, thanks for the comment! I just want to try out a couple more ideas (including the things you mentioned) and then I'll update the PR and provide a complete answer.

alexandruag:
Hi again, the PR is updated. I've added an `atomic_ref` method to `Bytes`, as well as `{read|write}_{atomic|aligned}`. To answer the previous questions:

- I think the automatic `Bytes` implementation for `GuestMemory` removes too much flexibility from custom implementations, but it's not really in the scope of this PR to remove it, so I've dropped that commit (it will return in a different PR :D).

- I've added an `Ordering` argument to `read/write_atomic`, and a default implementation that leverages `atomic_ref`. The presence of all these methods within the main `Bytes` interface is mainly a quality of life improvement (and the current implementation seems to make things faster). I think having the atomic operations clearly defined and easily accessible makes things safer and easier to understand when writing stuff like the virtio queue/device code, etc.

- `atomic_ref` and `VolatileMemory::get_atomic_ref` are actually not totally generic, in the sense that they require the access to happen within a memory region backed by host memory, because they return an actual reference. If this ever becomes an actual limitation, we can replace the reference with a reference-like abstraction. I know it's kinda awkward to have them both right now, but I think we should wait a bit more to see what other changes happen before we're happy with the interface.

- It's worth noting that `read/write_obj` (and similar) methods provide more than just better ergonomics when compared to `VolatileRef`, because the latter is only usable when the objects do not cross region boundaries (which makes it quite awkward to use; it also suffers from the "must be backed by host memory" restriction). Going forward, it would be nice if the `Array` and `Ref` abstractions hide away cross region accesses, and the `Bytes` methods can use them and only provide better ergonomics. 

- I wanted to update the PR as soon as possible to get more feedback so the `Aligned<A, T>` usability improvements are still on their way :( If ppl are ok with the PR, I'll add more documentation and the missing tests as well.

src/align.rs
```diff
@@ -0,0 +1,86 @@
...
+        Aligned {
+            addr: self.addr,
+            phantom: PhantomData,
+        }
+    }
+}
+
+impl<Addr: Address, T> Aligned<Addr, T> {
+    /// Attempt to create an `Aligned` value based on `addr`.
+    pub fn from_addr(addr: Addr) -> result::Result<Self, AlignmentError> {
```
bonzini, line 73: Should these two functions be implementations of `std::convert::TryFrom` instead?

src/guest_memory.rs
```diff
@@ -117,11 +119,33 @@ impl Display for Error {
 pub struct GuestAddress(pub u64);
 impl_address_ops!(GuestAddress, u64);
 
+/// Wraps a `GuestAddress` that's known to be aligned with respect to `T`.
+pub type AlignedGuestAddress<T> = Aligned<GuestAddress, T>;
+
+impl<T> std::convert::TryFrom<GuestAddress> for AlignedGuestAddress<T> {
```
bonzini, line 147: Implementing `impl<Addr, T> TryFrom<Addr> for Aligned<Addr, T>` should make these two more specific implementations unnecessary.

src/guest_memory.rs
```diff
@@ -117,11 +119,33 @@ impl Display for Error {
 pub struct GuestAddress(pub u64);
 impl_address_ops!(GuestAddress, u64);
 
+/// Wraps a `GuestAddress` that's known to be aligned with respect to `T`.
+pub type AlignedGuestAddress<T> = Aligned<GuestAddress, T>;
```
bonzini, line 145: Are these two types useful?

src/align.rs
```diff
@@ -18,12 +18,24 @@ pub enum AlignmentError {
 }
 
 /// An address that's aligned with respect to `T`.
-#[derive(Clone, Copy)]
 pub struct Aligned<Addr, T> {
     addr: Addr,
     phantom: PhantomData<T>,
 }
 
+// Implementing `Clone` and `Copy` manually because deriving them did not work well
```
bonzini, line 26: Would it work if you use a `PhantomData<*mut const T>` instead?

src/guest_memory.rs
```diff
@@ -115,9 +123,23 @@ impl Display for Error {
 /// On ARM64, a 32-bit hypervisor may be used to support a 64-bit guest. For simplicity,
 /// `u64` is used to store the the raw value no matter if the guest a 32-bit or 64-bit virtual
 /// machine.
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Ord, PartialOrd)]
 pub struct GuestAddress(pub u64);
+
+impl From<u64> for GuestAddress {
```
bonzini, line 129: I think these should be implemented automatically by `impl_address_ops!`

src/guest_memory.rs
```diff
@@ -115,9 +123,23 @@ impl Display for Error {
...
+    fn from(value: u64) -> Self {
+        GuestAddress(value)
+    }
+}
+
+impl Into<u64> for GuestAddress {
```
bonzini, line 135: And Into should probably never be implemented, instead you should implement `From<GuestAddress> for u64`.

src/guest_memory.rs
```diff
@@ -131,9 +153,23 @@ impl<T> std::convert::TryFrom<GuestAddress> for AlignedGuestAddress<T> {
 }
 
 /// Represents an offset inside a region.
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Ord, PartialOrd)]
 pub struct MemoryRegionAddress(pub u64);
+
+impl From<u64> for MemoryRegionAddress {
```
bonzini, line 159: Same here.

alexandruag:
Hi, I rebased & pushed two changes: using integer arithmetic + cast in `GuestRegionMmap::check_ptr` instead of `ptr::offset` (some more details in a comment there), and added an `offset<U>(self, value: Addr::V)` method to `Aligned<Addr, T>`, which is both an offset and a (potential) cast. 

I also tried to use some trait + bounds constructs to have Rust "know" which alignments are ok to cast into which. Although it works, things get a bit too convoluted so I've kept this simple approach where both `cast` and `offset` return a `Result`, and one way to avoid having too many `unwraps` (which often can be optimized away because they can't trigger) is to use `?` and a `From<AlignmentError>` implementation for the  error type returned by the calling function. I'm going to try and use the atomic methods on top of some of the existing code in rust-vmm/PRs to get a better feel for things from a convenience perspective.

alexandruag:
Hi, I've added unit tests as well in a new commit, and removed [RFC] from the PR name. Looking forward to any (re) reviews :D

src/align.rs
```diff
@@ -0,0 +1,206 @@
...
+
+        {
+            let u32_offset = 4;
+            let b = a.offset::<u32>(u32_offset).unwrap();
+            assert_eq!(b.addr, a.addr.unchecked_add(u32_offset));
+
+            assert_eq!(
+                a.offset::<u64>(u32_offset).unwrap_err(),
+                AlignmentError::Misaligned
+            );
```
jiangliu, line 203: How about adding another case:
```
      c = a.offset::<u64>(u32_offset).unwrap();
      assert_eq!(c.addr, b.addr.unchecked_add(u32_offset));
```

src/bytes.rs
```diff
@@ -193,6 +267,90 @@ pub trait Bytes<A> {
     /// Part of the data may have been copied nevertheless.
     fn read_slice(&self, buf: &mut [u8], addr: A) -> Result<(), Self::E>;
 
+    /// Returns a reference that allows atomic operations for `T` at the specified address.
```
jiangliu, line 270: I'm hesitating whether the atomic related interfaces should be hosted by a dedicated trait.
Per my understanding, we have three types of memory access modes,
1) read/write in byte stream mode.
2) read/write in naturally aligned mode (word/double word/quad word).
3) read/write/swap/compare_and_exchange in naturally aligned atomic mode with memory ordering.

Bytes::{read/write/read_slice/write_slice/read_from/write_to/read_exact_from/write_all_to} should access memory in byte stream mode.
But Bytes::{read_obj/write_obj} should access memory in naturally aligned mode.
So should we move read_obj/write_obj/atomic_ref/write_atomic/read_atomic out of the Bytes trait?

alexandruag: I think both the name and some of the doc comments (i.e. the one about *Candidates which may implement this trait include:...*) are a bit misleading for `Bytes` (in my mind I've always used something along the lines of `RandomAccessMemory`). Your overview of access types sounds reasonable, and what's forcing us right now to explicitly turn operations like `read_obj` into `memcpy`s is cross region access (I'm working on a proposal to address that, and will post something later this week).

To answer your question, I'm not totally sure what the best place for these operations is in the long term; having them in `Bytes` seems like the best option for now. I think we still need a bit more time to figure out how the stable `vm-memory` interface looks like, and we should address some of the existing concerns first.

alexandruag:
Closing for now in favor of #114. Thanks for the discussion so far everyone!

PR #109, opened 2020-08-10
mmap_unix: add support for externally managed mappings

slp:
Add the unsafe MmapRegion::build_raw method to allow users to instance
MmapRegion with externally managed mappings. Add also a field to
indicate whether the mapping is internal or external, so we don't try
to munmap it on MmapRegion::Drop.

Signed-off-by: Sergio Lopez <slp@redhat.com>

### DIFF
--- a/src/mmap_unix.rs
+++ b/src/mmap_unix.rs
@@ -28,6 +28,10 @@ use crate::volatile_memory::{self, compute_offset, VolatileMemory, VolatileSlice
 pub enum Error {
     /// The specified file offset and length cause overflow when added.
     InvalidOffsetLength,
+    /// The specified pointer to the mapping is not page-aligned.
+    InvalidPointer,
+    /// The specified size for the region is not a multiple of the page size.
+    InvalidSize,
     /// The forbidden `MAP_FIXED` flag was specified.
     MapFixed,
     /// Mappings using the same fd overlap in terms of file offset and length.
@@ -44,6 +48,14 @@ impl fmt::Display for Error {
             Error::InvalidOffsetLength => write!(
                 f,
                 "The specified file offset and length cause overflow when added"
+            ),
+            Error::InvalidPointer => write!(
+                f,
+                "The specified pointer to the mapping is not page-aligned",
+            ),
+            Error::InvalidSize => write!(
+                f,
+                "The specified size for the region is not a multiple of the page size",
             ),
             Error::MapFixed => write!(f, "The forbidden `MAP_FIXED` flag was specified"),
             Error::MappingOverlap => write!(
@@ -79,6 +91,7 @@ pub struct MmapRegion {
     file_offset: Option<FileOffset>,
     prot: i32,
     flags: i32,
+    owned: bool,
 }
 
 // Send and Sync aren't automatically inherited for the raw address pointer.
@@ -160,6 +173,48 @@ impl MmapRegion {
             file_offset,
             prot,
             flags,
+            owned: true,
+        })
+    }
+
+    /// Creates a MmapRegion instance for an externally managed mapping.
+    ///
+    /// This method is intended to be used exclusively in situations in which the mapping backing
+    /// the region is provided by an entity outside the control of the caller (e.g. the dynamic
+    /// linker).
+    ///
+    /// # Arguments
+    /// * `addr` - Pointer to the start of the mapping. Must be page-aligned.
+    /// * `size` - The size of the memory region in bytes. Must be a multiple of the page size.
+    /// * `prot` - Must correspond to the memory protection attributes of the existing mapping.
+    /// * `flags` - Must correspond to the flags that were passed to `mmap` for the creation of
+    ///             the existing mapping.
+    ///
+    /// # Safety
+    ///
+    /// To use this safely, the caller must guarantee that `addr` and `size` define a region within
+    /// a valid mapping that is already present in the process.
+    pub unsafe fn build_raw(addr: *mut u8, size: usize, prot: i32, flags: i32) -> Result<Self> {
+        // Safe because this call just returns the page size and doesn't have any side effects.
+        let page_size = libc::sysconf(libc::_SC_PAGESIZE) as usize;
+
+        // Check that the pointer to the mapping is page-aligned.
+        if (addr as usize) & (page_size - 1) != 0 {
+            return Err(Error::InvalidPointer);
+        }
+
+        // Check that the size is a multiple of the page size.
+        if size & (page_size - 1) != 0 {
+            return Err(Error::InvalidSize);
+        }
+
+        Ok(Self {
+            addr,
+            size,
+            file_offset: None,
+            prot,
+            flags,
+            owned: false,
         })
     }
 
@@ -188,6 +243,11 @@ impl MmapRegion {
     /// Returns the value of the `flags` parameter passed to `mmap` when mapping this region.
     pub fn flags(&self) -> i32 {
         self.flags
+    }
+
+    /// Returns `true` if the mapping is owned by this `MmapRegion` instance.
+    pub fn owned(&self) -> bool {
+        self.owned
     }
 
     /// Checks whether this region and `other` are backed by overlapping
@@ -252,8 +312,10 @@ impl Drop for MmapRegion {
     fn drop(&mut self) {
         // This is safe because we mmap the area at addr ourselves, and nobody
         // else is holding a reference to it.
-        unsafe {
-            libc::munmap(self.addr as *mut libc::c_void, self.size);
+        if self.owned {
+            unsafe {
+                libc::munmap(self.addr as *mut libc::c_void, self.size);
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


alexandruag:
Hi! If possible, can you please share more details regarding the use case(s) for this change? Some of the usage examples I'm trying to think of can be accomplished via alternative means (for example, mapping the same `memfd` in multiple places).

slp:
> Hi! If possible, can you please share more details regarding the use case(s) for this change? Some of the usage examples I'm trying to think of can be accomplished via alternative means (for example, mapping the same `memfd` in multiple places).

@alexandruag In [libkrun](https://github.com/libkrun/libkrun) we need to create a MmapRegion for a section coming from a dynamically linked library that has been mapped by the linker. The file backing the library may no longer be accessible due to context changes (namespace manipulation).

I acknowledge that our use case may be a bit particular, but I think the change we need in ```vm-memory``` is so small and simple that shouldn't become a maintenance burden.

src/mmap_unix.rs
```diff
@@ -79,6 +79,7 @@ pub struct MmapRegion {
     file_offset: Option<FileOffset>,
     prot: i32,
     flags: i32,
+    external: bool,
```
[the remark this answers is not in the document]
slp: Sounds good to me.

slp:
> Thanks for the explanation! `libkrun` looks very interesting! Just left a minor comment, and it seems you also need to update the coverage value in `coverage_config_x86_64` (more details on the CI results page).

@alexandruag Thanks! I've updated the PR to replace ```external``` with ```owned``` and fix the coverage value (wasn't sure who should update it, thanks for the clarification).

alexandruag:
Looks good! I just realized there are two points I forgot to mention in my previous comment, sorry about that :( 

While I don't think this is mentioned anywhere yet, we do effectively rely on regions having certain alignment and size restrictions. I guess it will be simplest to state at some point (and check) that a region is aligned to a page boundary and its size is a multiple of the page size. Does that hamper your use case in any way?

Also, I forgot we have a changelog now :-s Can you please add an `Added` entry in `CHANGELOG.md` for the `[Unreleased]` version?

jiangliu:
> Looks good! I just realized there are two points I forgot to mention in my previous comment, sorry about that :(
> 
> While I don't think this is mentioned anywhere yet, we do effectively rely on regions having certain alignment and size restrictions. I guess it will be simplest to state at some point (and check) that a region is aligned to a page boundary and its size is a multiple of the page size. Does that hamper your use case in any way?
We do have some implicit assumption that the base address of a region is PAGE aligned. And on Linux, the mmap() syscall ensures that file offset and returned address are page aligned. Not sure about the behavior mmap on Windows.
So we should ensure the region base and length are PAGE aligned.

slp:
> While I don't think this is mentioned anywhere yet, we do effectively rely on regions having certain alignment and size restrictions. I guess it will be simplest to state at some point (and check) that a region is aligned to a page boundary and its size is a multiple of the page size. Does that hamper your use case in any way?

Sounds good, and shouldn't be a problem for libkrun. I've just added those checks and extended the test to ensure it also fails when expected. Could you please take a look?
 
> Also, I forgot we have a changelog now :-s Can you please add an `Added` entry in `CHANGELOG.md` for the `[Unreleased]` version?

Sure, no problem. I see all the existing entries refer to an issue, do I need to create one covering the goal of this PR?

slp:
Hm... coverage has dropped by 0.30% now... will take a look at it tomorrow.

slp:
> Hm... coverage has dropped by 0.30% now... will take a look at it tomorrow.

Okay, so the coverage drop was caused by the newly added `Errors`, so I've added a `test_display_error` test and coverage is now up to 86.0% (from 85.2%).

alexandruag:
Thanks for the changes! Things look good. We'll document the alignment & such requirements after we polish the interface a bit more, and by then we should also figure out what the exact model & semantics are (regions being aligned looks mandatory, but the others might be relaxed).

I think it's fine to use a reference to the PR itself in the change log as well, but just in case this contradicts some style guidelines we're trying to establish, do you mind if we wait just a bit more to see if others have a different opinion?

slp:
> I think it's fine to use a reference to the PR itself in the change log as well, but just in case this contradicts some style guidelines we're trying to establish, do you mind if we wait just a bit more to see if others have a different opinion?

No problem, we can wait a bit to see what's the consensus about that. Thanks!

jiangliu (approved):

src/mmap_unix.rs
```diff
@@ -160,6 +176,47 @@ impl MmapRegion {
...
+    /// * `prot` - The desired memory protection of the mapping.
+    /// * `flags` - This argument determines whether updates to the mapping are visible to other
+    ///             processes mapping the same region, and whether updates are carried through to
+    ///             the underlying file.
+    ///
+    /// # Safety
+    ///
+    /// To use this safely, the caller must guarantee that `addr` and `size` define a region within
+    /// a valid mapping that is already present in the process.
+    pub unsafe fn build_raw(addr: *mut u8, size: usize, prot: i32, flags: i32) -> Result<Self> {
```
andreeaflorescu, line 197: Can you also add an `Example` section? It will help people find where to get defines for `prot` and `flags` from.
The example can be something like:
```
let addr = 0;
let size = unsafe { libc::sysconf(libc::_SC_PAGESIZE) as usize };
let prot = libc::PROT_READ | libc::PROT_WRITE;
let flags = libc::MAP_NORESERVE | libc::MAP_PRIVATE;
let r = unsafe { MmapRegion::build_raw(addr as *mut u8, size, prot, flags).unwrap() };
```

src/mmap_unix.rs
```diff
@@ -160,6 +176,47 @@ impl MmapRegion {
             file_offset,
             prot,
             flags,
+            owned: true,
+        })
+    }
+
+    /// Creates a MmapRegion instance for an externally managed mapping.
```
andreeaflorescu, line 183: I think it would be useful for other people that look at the interface of MmapRegion to also to give an example of situation in which people can choose to use this function as opposed to the other builders. This helps people that look at the vm-memory interface for the first time to decide when to use one over the other.

andreeaflorescu (commented):
Regarding the changelog, since there is no associated issue with this PR, we can either create one, or link the PR. As @alexandruag said, we don';t really have any policy regarding this so far.

Also, sorry for jumping in with the review so late. Feel free to consider my comments nits and address them in a subsequent PR. They are mostly related to documenting the added functionality so that external users have an easier time :D

src/mmap_unix.rs
```diff
@@ -24,10 +24,15 @@ use crate::mmap::{check_file_offset, AsSlice};
 use crate::volatile_memory::{self, compute_offset, VolatileMemory, VolatileSlice};
 
 /// Error conditions that may arise when creating a new `MmapRegion` object.
+#[allow(missing_docs)]
```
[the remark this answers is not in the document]
slp: That's to avoid a complain for the anonymous structs added to `Error` for the `InvalidPointer` and `InvalidSize` variants. We already have something similar in `volatile_memory::Error`.

On the other hand, I've added that just for the coverage test. If we remove that test, we can get rid of this.

src/mmap_unix.rs
```diff
@@ -278,6 +342,19 @@ mod tests {
         }
     }
 
+    #[test]
```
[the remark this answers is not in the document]
slp: Sure, if decreasing the coverage it's fine, we can drop it.

src/mmap_unix.rs
slp: Hm... actually, I'd expect `build_raw` to be used only in situations where the user doesn't have direct control of the mapping, which implies that `prot` and `flags` should correspond to the actual properties of the already existing mapping.

On a second look, I see the method's documentation for `prot` and `flags` is not clear about this. What do you think if we clarify that instead of adding an example?

src/mmap_unix.rs
slp: Sounds like a good idea to me. I'm going to add it.

src/mmap_unix.rs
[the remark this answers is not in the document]
andreeaflorescu: That'd be nice.

src/mmap_unix.rs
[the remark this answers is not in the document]
andreeaflorescu: Coverage should be there just so that we get a better idea when we forgot to add tests. I am pretty against coverage tests, but I can understand how it can be rather a subjective view. I would drop it if you're also fine with that.

src/mmap_unix.rs
andreeaflorescu: Thanks!

src/mmap_unix.rs
andreeaflorescu: Sounds good.

slp:
@andreeaflorescu I've rebased the commit and addressed your comments. PTAL, thanks!

jiangliu (approved):

slp:
I've just noticed that I forgot to add the entry to `CHANGELOG.md`. Fixed now.

jiangliu (approved):

andreeaflorescu (approved):

PR #114, opened 2020-09-21
[RFC] vm-memory refactoring proposal

alexandruag:
Hi! This RFC proposes changes to `vm-memory` with the purpose of improving performance, avoiding duplication of logic that accesses memory contents, and simplifying/eliminating some of the crate contents. For now, the PR mostly proposes changes and several alternative abstractions; more commits that remove existing things will follow.

The proposal is based on the simplifying assumption that guest memory regions adjacent in terms of guest physical addresses can be contiguous in process virtual memory as well. This mean we can aggregate multiple neighbouring regions into a single continuous range, which removes the need to  implement special handling for cross-region accesses. The PR adds a `GuestRange` helper object and checks that guest regions provided to `GuestMemoryMmap` constructors satisfy the condition (otherwise an error is generated). There's no logic yet which helps create such memory objects (will be added later; there's an example in the benches code that uses a simplistic approach, and others are possible). Also, for use cases that do not require `GuestMemoryMmap` objects with adjacent regions to be created, folks wanting to try things out can simply `[patch]` their `vm-memory` dependency in `Cargo.toml` with commit f2f4db7a to see if there's any noticeable difference.

The PR also introduces an `access` module, which leverages the previous assumption to provide a single `Bytes` implementation for all entities which expose the interface, based on much simpler helper traits. It also proposes `ArrayRef` and `Ref` as simpler alternatives to `VolatileRef`, `VolatileArrayRef`, and `VolatileSlice`. Finally, it also adds `Aligned` and atomic access logic based on parts of #104. I think the aligned part is useful, but at the same time it's orthogonal to everything else and can be dropped if ppl don't think it brings any value. More details are available in commit messages and individual comments.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,6 @@ autobenches = false
 
 [features]
 default = []
-integer-atomics = []
 backend-mmap = []
 backend-atomic = ["arc-swap"]
 
--- a/benches/main.rs
+++ b/benches/main.rs
@@ -18,7 +18,7 @@ pub fn criterion_benchmark(_c: &mut Criterion) {
 
 criterion_group! {
     name = benches;
-    config = Criterion::default().sample_size(200).measurement_time(std::time::Duration::from_secs(50));
+    config = Criterion::default().sample_size(200).measurement_time(std::time::Duration::from_secs(30));
     targets = criterion_benchmark
 }
 
--- a/benches/mmap/mod.rs
+++ b/benches/mmap/mod.rs
@@ -13,7 +13,10 @@ use std::path::Path;
 
 use criterion::{black_box, Criterion};
 
-use vm_memory::{ByteValued, Bytes, GuestAddress, GuestMemory, GuestMemoryMmap};
+use vm_memory::{
+    Address, ByteValued, Bytes, GuestAddress, GuestMemory, GuestMemoryMmap, GuestRegionMmap,
+    MmapRegion,
+};
 
 const REGION_SIZE: u64 = 0x8000_0000;
 const REGIONS_COUNT: u64 = 8;
@@ -62,12 +65,36 @@ impl AccessKind {
 }
 
 pub fn benchmark_for_mmap(c: &mut Criterion) {
-    let mut regions: Vec<(GuestAddress, usize)> = Vec::new();
-    for i in 0..REGIONS_COUNT {
-        regions.push((GuestAddress(i * REGION_SIZE), REGION_SIZE as usize));
-    }
+    // Since we've changed the `GuestMemoryMmap` constructor to require regions which are
+    // contiguous w.r.t. GPAs to do the same for HVAs, we need to manually ensure this holds for
+    // our region configuration (until we provide functionality for building, and not just checking
+    // in `vm-memory`. We're using a quick approach where we allocate one large region, which is
+    // then used to build subregions based on `MmapRegion::build_raw`. 
+    let helper_region = MmapRegion::new((REGION_SIZE * REGIONS_COUNT) as usize).unwrap();
+    let memory = GuestMemoryMmap::from_regions(
+        (0..REGIONS_COUNT as usize)
+            .into_iter()
+            .map(|idx| {
+                // The following are safe because this is just for testing/illustration purposes.
+                let addr = unsafe {
+                    helper_region
+                        .as_ptr()
+                        .offset(idx as isize * REGION_SIZE as isize)
+                };
 
-    let memory = GuestMemoryMmap::from_ranges(regions.as_slice()).unwrap();
+                let mmap_region =
+                    unsafe { MmapRegion::build_raw(addr, REGION_SIZE as usize, 0, 0) }.unwrap();
+
+                GuestRegionMmap::new(
+                    mmap_region,
+                    GuestAddress(0).unchecked_add(REGION_SIZE * idx as u64),
+                )
+                .unwrap()
+            })
+            .collect(),
+    )
+    .unwrap();
+
     // Just a sanity check.
     assert_eq!(
         memory.last_addr(),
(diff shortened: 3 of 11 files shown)


bonzini:
> The proposal is based on the simplifying assumption that guest memory regions adjacent in terms of guest physical addresses can be contiguous in process virtual memory as well. 

Do you mean "must be contiguous"? This unfortunately is not possible.

The simplest case is two PCI BARs that are programmed by the guest to be adjacent in terms of guest physical addresses.  Those BARs likely will come from two completely separate mmaps, and won't be contiguous. Also because the BAR addresses are decided by the guest, the guest can shuffle them around as much as it wants.

Memory hot-plug is similar, if the hotplug interface allows the guest to specify a guest physical address of its choice.

That said, I do like a lot:

* the aligned `Ref` and `ArrayRef`. Those are a no-brainer. Perhaps we can rename the existing `VolatileRef` and `VolatileArrayRef` to `UnalignedRef` and `UnalignedArrayRef`, with the old names kept for backwards compatibility?

* the automatic bytes implementation, though I am worried that the blanket `impl<A> Bytes for C where C: CheckAccess<A>` is too limiting; IIUC it prevents anyone else from adding their own blanket implementation of Bytes for one particular address type.

alexandruag:
Hi Paolo! Thanks so much for the swift reply! For some reason I was thinking things such as your examples are not actually undermining the viability of the approach. For example, in the hotplug case, we can remap on the host side to keep the correspondence going (does the guest ever overlap guest physical regions?). I'll get in the tank about this and return with an answer or even more questions :-s

I think it's reasonable not to avoid breaking backwards compatibility at this point, especially since those abstractions were kinda broken by cross-region access. Whichever way we go, converging on a stable interface is prob most important right now. I'm not sure whether your concern about the auto implementation applies outside the crate (because `AutoBytes` is not exported), but will def investigate that at least to understand better what's going on.

bonzini:
> in the hotplug case, we can remap on the host side to keep the correspondence going

You could use memfd I suppose instead of anonymous mmap I guess. But it's not easy, for example remapping on the host side needs to be done carefully to avoid race conditions against accesses in the guest. There could be address space fragmentation concerns too.

> AutoBytes is not exported

Oh I missed that. But perhaps it _should_ be exported. :)

alexandruag:
Hmm, seems there are so many possible settings that it's better to begin with a description of how things may work out, and then touch upon the concerns as well. If `GuestRegionMmap` objects are only created via `mmap` (which was strictly the case before adding `MmapRegion::build_raw`), the continuity assumption from the PR is straightforward from a construction perspective. For example, we can start by reserving a sufficiently large virtual memory range using an anon `mmap` with `PROT_NONE`, and then have fine grained control over the virtual addresses of individual regions with `mmap` + `MAP_FIXED` inside the range. Fragmentation problems cannot be ruled out, but I imagined most use cases involve starting with a set of regions, and then "growing at one end" or something similar.

Going back to Paolo's examples, the PCI BAR scenario is problematic only if we need to create new regions on reaction when BAR addresses get shuffled. Is that actually the case? Memory hotplug where the guest arbitrarily picks the GPA is tricky, but that's also true for the current approach to an extent because a region won't work until the appropriate configuration gets propagated.

That being said, if the set of use cases ppl envision for `GuestMemoryMmap` is incompatible with the previous assumption we can close this. I'll prepare another PR that brings over atomic access support to `Bytes`, and applies various simplifications at the interface level only (for example, we can temporarily remove `VolatileRef` & others from the `GuestMemory` traits because they don't abstract away cross-region access or regions not backed by host-memory, but keep them as part of the backend implementation that people already use).

This allows us to continue with a minimal stable interface (which can be confidently used for generic consumers such as `vm-virtio`, without affecting any specific backend), and we can bring back more things as we fix them moving forward. How does that sound?

bonzini:
> Going back to Paolo's examples, the PCI BAR scenario is problematic only if we need to create new regions on reaction when BAR addresses get shuffled

Seems like it's the opposite: because we want to `preserve` the contents of the regions when BAR addresses get shuffled, generally speaking we cannot do an `munmap` and remap inside the large region with `MAP_FIXED`.

Another problem is aliased regions, i.e. multiple regions backed by the same memory. One example is (at boot on PC) `0xe0000` which aliases the last 128K of the ROM (i.e. it duplicates the range `0xfffe0000` to `0xffffffff`). Unfortunately, at `0xdffff` you have low memory while at `0xfffdffff` you have the rest of the ROM; and likewise `0x100000` and `0x100000000` point to different memory.

You can see this in a table if you run QEMU with `qemu-system-x86_64 -m 4G -monitor stdio -S` and type `info mtree -f`. You'll get something like

```
...
 Root memory region: system
  0000000000000000-000000000009ffff (prio 0, ram): pc.ram kvm
  00000000000a0000-00000000000bffff (prio 1, i/o): vga-lowmem
  00000000000c0000-00000000000dffff (prio 1, rom): pc.rom kvm
  00000000000e0000-00000000000fffff (prio 0, rom): pc.bios @0000000000020000 kvm
  0000000000100000-00000000bfffffff (prio 0, ram): pc.ram @0000000000100000 kvm
  00000000fec00000-00000000fec00fff (prio 0, i/o): kvm-ioapic
  00000000fed00000-00000000fed003ff (prio 0, i/o): hpet
  00000000fee00000-00000000feefffff (prio 4096, i/o): kvm-apic-msi
  00000000fffc0000-00000000ffffffff (prio 0, rom): pc.bios kvm
  0000000100000000-000000013fffffff (prio 0, ram): pc.ram @00000000c0000000 kvm
...
```

Even ignoring the "i/o" regions, which aren't handled by vm-memory in Firecracker or cloud-hypervisor, there are problematic  consecutive regions all over the place.

alexandruag:
Thanks for the examples Paolo! I'll def dive deeper into these areas going forward. Closing this PR. I would like to try and proceed with the plan outlined above around simplifying the backend-agnostic interfaces for now. I posted #115 as a first step, and would be great if you can have a look.

PR #115, opened 2020-09-22
Add methods for atomic access to Bytes

alexandruag:
This PR adds the `load` and `store` methods to the `Bytes` interface, which can be used to perform atomic accesses based on a specified `Ordering` argument. It's a simplified variant of #104, which does not include the abstractions for aligned accesses.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,6 @@ autobenches = false
 
 [features]
 default = []
-integer-atomics = []
 backend-mmap = []
 backend-atomic = ["arc-swap"]
 
--- a/src/atomic_integer.rs
+++ b/src/atomic_integer.rs
@@ -1,0 +1,90 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+use std::sync::atomic::Ordering;
+
+/// Objects that implement this trait must consist exclusively of atomic types
+/// from [`std::sync::atomic`](https://doc.rust-lang.org/std/sync/atomic/), except for
+/// [`AtomicPtr<T>`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicPtr.html) and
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -12,7 +12,6 @@ autobenches = false
 
 [features]
 default = []
-integer-atomics = []
```
jiangliu, line 15: This seems like a breaking change, should we bump the minor version number for next release?

src/atomic_integer.rs
```diff
@@ -0,0 +1,86 @@
+use std::sync::atomic::Ordering;
```
jiangliu, line 1: Could you please help to add a license header?

src/atomic_integer.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// Objects that implement this trait must consist exclusively of atomic types
+/// from [`std::sync::atomic`](https://doc.rust-lang.org/std/sync/atomic/), except for
+/// [`AtomicPtr<T>`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicPtr.html).
```
jiangliu, line 5: How about AtomicBool?

alexandruag: `AtomicBool` is dangerous in this context (i.e. we're essentially casting some memory from the guest to a type), because values other than `0` and `1` for a bool value may lead to undefined behaviour.

bonzini:
Looks good to me!

src/atomic_integer.rs
jiangliu: So it may be better to refine the comments  as "except for [`AtomicPtr<T>`] and AtomicBool<T>".

jiangliu:
LGTM

src/atomic_integer.rs
alexandruag: That's totally right. There was no mention of `AtomicBool` initially, but I've added *except for `AtomicPtr` and `AtomicBool` ...* as part of the commit which expands the `AtomicInteger` trait. The message from the commit that removes the `AtomicInteger` for `AtomicBool` impl also offers a bit more details abou the why.

jiangliu (approved):

aghecenco (approved):

Issue #100, opened 2020-06-03
Performance degradation after fixing #95


andreeaflorescu:
When implementing the fix for #95, we introduced a performance degradation on (some?) glibc builds. On Firecracker with iperf3 and we observed a performance degradation of 5% for glibc builds.
On Cloud Hypervisor, the performance degradation is significantly worse. Observed impact is up to 50%. More details here: https://github.com/cloud-hypervisor/cloud-hypervisor/issues/1258

Opening this issue so that we can decide on the next steps for fixing the performance degradation.

My 2-cents: I wouldn't want to introduce this fix only for x86_64 musl builds & aarch64 glibc & musl builds because we cannot know for sure what glibc versions are people using out there, hence we cannot know for sure that glibc is doing the right thing (i.e. optimizing memcpy at a higher granularity than 1 byte).

I would say that the underlying problem which is the reason for the performance degradation & the bug, is that we lose type information about the object that needs to written/read. I would rather like us to work on improving the interface so that type information doesn't need to be sort-of inferred (by checking alignments & reading/writing in the largest possible chunks). I would need to do some experiments before having a solution here.

Another thing that I believe is of paramount importance at this moment is to add performance testing to vm-memory. Pretty much every other function in vm-memory is on the critical performance path. We should make sure to not introduce regression here as we continue the development.

CC: @sboeuf @rbradford @sameo @alexandruag @serban300 @bonzini

bonzini:
> we lose type information about the object that needs to written/read

Why is that causing performance degradation? The underlying object is likely `&[u8]`, but *because* the memcpy-like code does not care about the underlying type it is still able to copy in 64-bit chunks. The problem is simply that the glibc memcpy is more sophisticated.

> Pretty much every other function in vm-memory is on the critical performance path. We should make sure to not introduce regression here as we continue the development.

Agreed. Though a better implementation of virtio-net could avoid the copies by doing `recv` and `send` directly into memory.

rbradford:
@bonzini We're looking at using `.read_from()` and `.write_to()` but that still involve at least one "vm-memory copy" (see cloud-hypervisor/cloud-hypervisor#1265)

rbradford:
> @bonzini We're looking at using `.read_from()` and `.write_to()` but that still involve at least one "vm-memory copy" (see [cloud-hypervisor/cloud-hypervisor#1265](https://github.com/cloud-hypervisor/cloud-hypervisor/issues/1265))

[...]

PR #117, opened 2020-10-02
volatile_memory: Only use volatile copy for small objects

rbradford:
Where small objects are those objects that are less then the native data
width for the platform. This ensure that volatile and alignment safe
read/writes are used when updating structures that are sensitive to this
such as virtio devices where the spec requires writes to be atomic.

Fixes: cloud-hypervisor/cloud-hypervisor#1258
Fixes: #100

Signed-off-by: Rob Bradford <robert.bradford@intel.com>

### DIFF
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 85.4,
+  "coverage_score": 85.5,
   "exclude_path": "mmap_windows.rs",
   "crate_features": "backend-mmap,backend-atomic"
 }
--- a/src/volatile_memory.rs
+++ b/src/volatile_memory.rs
@@ -1082,7 +1082,7 @@ mod copy_slice_impl {
         }
     }
 
-    pub(super) fn copy_slice(dst: &mut [u8], src: &[u8]) -> usize {
+    fn copy_slice_volatile(dst: &mut [u8], src: &[u8]) -> usize {
         let total = min(src.len(), dst.len());
         let mut left = total;
 
@@ -1107,6 +1107,17 @@ mod copy_slice_impl {
         copy_aligned_slice(4);
         copy_aligned_slice(2);
         copy_aligned_slice(1);
+
+        total
+    }
+
+    pub(super) fn copy_slice(dst: &mut [u8], src: &[u8]) -> usize {
+        let total = min(src.len(), dst.len());
+        if total <= size_of::<usize>() {
+            copy_slice_volatile(dst, src);
+        } else {
+            dst[..total].copy_from_slice(&src[..total]);
+        }
 
         total
     }


rbradford:
Consensus in #100 seems to be that this is the right approach. It would be nice if we could drop our fork of vm-memory.

jiangliu:
Now the vm-memory create has atomic interfaces for atomic access. Is it suitable to split existing vm-memory interfaces into byte stream access and object access? 
Current the read_obj()/write_obj() could be classified as object access, and all other existing interfaces of GuestMemory are byte stream accesses. So instead of heuristic algorithm, we may use:
1) copy_slice() for read_obj()/write_obj() with correctly aligned address
2) revert to original slice.write(buf) for other access methods.

rbradford:
> Now the vm-memory create has atomic interfaces for atomic access. Is it suitable to split existing vm-memory interfaces into byte stream access and object access?
> Current the read_obj()/write_obj() could be classified as object access, and all other existing interfaces of GuestMemory are byte stream accesses. So instead of heuristic algorithm, we may use:
> 
>     1. copy_slice() for read_obj()/write_obj() with correctly aligned address
>
>     2. revert to original slice.write(buf) for other access methods.

Sorry @jiangliu I don't really follow what you are saying. Could you elaborate more, giving some concrete examples? I think you're saying that we need to check everywhere that the code is being used to ensure we're using the correct API?

jiangliu:
> > Now the vm-memory create has atomic interfaces for atomic access. Is it suitable to split existing vm-memory interfaces into byte stream access and object access?
> > Current the read_obj()/write_obj() could be classified as object access, and all other existing interfaces of GuestMemory are byte stream accesses. So instead of heuristic algorithm, we may use:
> > ```
> > 1. copy_slice() for read_obj()/write_obj() with correctly aligned address
> > 
> > 2. revert to original slice.write(buf) for other access methods.
> > ```
> 
> Sorry @jiangliu I don't really follow what you are saying. Could you elaborate more, giving some concrete examples? I think you're saying that we need to check everywhere that the code is being used to ensure we're using the correct API?

Sorry for my poor english. What I means is
1) revert #95
2) only use copy_slice()(introduced by #95 ) for read_obj()/write_obj().

rbradford:
> > > Now the vm-memory create has atomic interfaces for atomic access. Is it suitable to split existing vm-memory interfaces into byte stream access and object access?
> > > Current the read_obj()/write_obj() could be classified as object access, and all other existing interfaces of GuestMemory are byte stream accesses. So instead of heuristic algorithm, we may use:
> > > ```
> > > 1. copy_slice() for read_obj()/write_obj() with correctly aligned address
> > > 
> > > 2. revert to original slice.write(buf) for other access methods.
> > > ```
> > 
> > 
> > Sorry @jiangliu I don't really follow what you are saying. Could you elaborate more, giving some concrete examples? I think you're saying that we need to check everywhere that the code is being used to ensure we're using the correct API?
> 
> Sorry for my poor english. What I means is
> 
>     1. revert #95
> 
>     2. only use copy_slice()(introduced by #95 ) for read_obj()/write_obj().

Ah, cool, I get it. I think that might solve our problem as we use the `.write_slice()/.read_slice()` entry points on our critical paths. Are you going to propose a patch?

jiangliu:
> > > > Now the vm-memory create has atomic interfaces for atomic access. Is it suitable to split existing vm-memory interfaces into byte stream access and object access?
> > > > Current the read_obj()/write_obj() could be classified as object access, and all other existing interfaces of GuestMemory are byte stream accesses. So instead of heuristic algorithm, we may use:
> > > > ```
> > > > 1. copy_slice() for read_obj()/write_obj() with correctly aligned address
> > > > 
> > > > 2. revert to original slice.write(buf) for other access methods.
> > > > ```
> > > 
> > > 
> > > Sorry @jiangliu I don't really follow what you are saying. Could you elaborate more, giving some concrete examples? I think you're saying that we need to check everywhere that the code is being used to ensure we're using the correct API?
> > 
> > 
> > Sorry for my poor english. What I means is
> > ```
> > 1. revert #95
> > 
> > 2. only use copy_slice()(introduced by #95 ) for read_obj()/write_obj().
> > ```
> 
> Ah, cool, I get it. I think that might solve our problem as we use the `.write_slice()/.read_slice()` entry points on our critical paths. Are you going to propose a patch?

We are on Chinese National Holiday this week, so it would be great if you could help on this:)

alexandruag:
Hi, sorry for the late review. The changes look good! #95 was necessary because we were using `read/write_obj` as if they have atomic access semantics (which is required in several places, such as virtio queue handling). The recently published `v0.3.0` adds the `load` and `store` methods to `Bytes`, which are guaranteed to perform atomic accesses (with a user-specified ordering). 

At this point we can go back to a world where `read/write_obj` do not have to provide any guarantees w.r.t. concurrent access, which simplifies the code and revers the performance impact. However, we have to ensure (to the extent possible) that `vm-memory` users switch to the new methods for accesses that need to be atomic first. We can use this patch while the transition is going on. Does that sound reasonable?

alexandruag:
Hi again! Should we merge this? Also, what do you think about reverting #95 entirely at some point in the future? (including for `read/write_obj` since they are not actually supposed to be used for atomic accesses; we need to clarify that as part of the interface documentation as well)

jiangliu:
It sounds reasonable to merge this PR as a transient solution to avoid regressions. And eventually we should revert to `copy_from_slice`.

jiangliu (approved):

slp:
I was wondering when would be the right time to revert #95 ? Perhaps we can use the recent jump to 0.3.x as an opportunity to do it now? Users needing the old behavior can stay at 0.2.2.

alexandruag:
I think that reverting should take place as part of a subsequent series (i.e. 0.4.x), so that ppl can use 0.3.x to transition to the new atomic methods while the old `read/write_obj` behaviour remains in place as well. Afterwards, the main concerns are that consumers which still rely on the old semantics should be careful not to inadvertently upgrade to the newer versions, and we'll prob have to backport stuff for a while. I guess we should start by engaging the known customers of `vm-memory` (and also open an issue about it) to see how fast we can move forward. I'll start doing that (+ double check once more and merge this PR if there are no further comments) by tomorrow.

slp:
@alexandruag Sounds like a sensible strategy to me.

alexandruag (approved):

Issue #118, opened 2020-10-05
Support MmapRegion with huge pages


iulianbarbu:
**Feature request**

Add support for huge pages guest memory. For a specific use case, please consult [this issue](https://github.com/firecracker-microvm/firecracker/issues/2139).

alexandruag:
Hi! Just to add a bit more context, `MmapRegion` objects that are backed by huge pages can be created using the `MmapRegion::build` constructor with a `flags` value that includes `MAP_HUGETLB` (there's also the option to use transparent huge pages, for example as discussed around https://github.com/rust-vmm/vm-memory/pull/113). Using an externally managed mapping via `MmapRegion::build_raw` is also possible. What's currently missing is a way to request huge page support directly via the high level `GuestMemoryMmap` constructors, which is something we're thinking about going forward.

EmeraldShift:
@iulianbarbu 

Hi! @doracgp and I are students at the University of Texas at Austin, and we are interested in virtualization and hypervisor technology. We'd like to tackle this issue. Can it be assigned to us?

To be sure, what does this issue entail, beyond presenting a clear interface to request huge pages, backed by these low level functions as @alexandruag describes?

We are also interested in using such a new API downstream in https://github.com/firecracker-microvm/firecracker/issues/2139.

rbradford:
You'd want to use `memfd_create()` to create the huge page fs rather than fiddling about with opening a `hugetlbfs` filesystem mount. This is how we do it in Cloud Hypervisor.

jiangliu:
There's no perfect solution for huge page yet, both hugetlbfs and memfd have limitations. According to our experience, memfd is preferred.

EmeraldShift:
> You'd want to use `memfd_create()` to create the huge page file descriptor rather than fiddling about with opening a `hugetlbfs` filesystem mount. This is how we do it in Cloud Hypervisor.

Is the intention here fully encapsulate the use of memfd and present an "anonymous huge-page-backed" (and maybe also "file huge-page-backed"?) constructor for `GuestMemoryMmap`? It looks like a way to go about this would be to either (1) add a bool `huge` to `from_ranges` indicating the desired page size, or (2) add a new constructor method which indicates a request for huge-page backing.

As far as alignment requirements, are those the responsibility of the VMM, or of this crate? I'm not seeing eager checks on `GuestAddress` alignment, but I might just be missing it.

doracgp:
[...]

PR #120, opened 2020-11-06
Add is_hugetlbfs() to GuestMemoryRegion

teawater:
Virtio-balloon can release the unused host memory to decrease the memory usage of the VMM.
 Release normal pages and hugetlbfs pages require different operations. (madvise MADV_DONTNEED and fallocate64 FALLOC_FL_PUNCH_HOLE)

This commit add Add is_hugetlbfs() to GuestMemoryRegion to help VMM decide if this is a hugetlbfs address or not.

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -289,6 +289,26 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
     fn as_volatile_slice(&self) -> Result<volatile_memory::VolatileSlice> {
         self.get_slice(MemoryRegionAddress(0), self.len() as usize)
     }
+
+    /// Show if the region is based on the `HugeTLBFS`.
+    /// Returns Some(true) if the region is backed by hugetlbfs.
+    /// None represents that no information is available.
+    ///
+    /// # Examples (uses the `backend-mmap` feature)
+    ///
+    /// ```
+    /// # #[cfg(feature = "backend-mmap")]
+    /// # {
+    /// #   use vm_memory::{GuestAddress, GuestMemory, GuestMemoryMmap, GuestRegionMmap};
+    ///     let addr = GuestAddress(0x1000);
+    ///     let mem = GuestMemoryMmap::from_ranges(&[(addr, 0x1000)]).unwrap();
+    ///     let r = mem.find_region(addr).unwrap();
+    ///     assert_eq!(r.is_hugetlbfs(), None);
+    /// # }
+    /// ```
+    fn is_hugetlbfs(&self) -> Option<bool> {
+        None
+    }
 }
 
 /// `GuestAddressSpace` provides a way to retrieve a `GuestMemory` object.
@@ -1198,5 +1218,14 @@ mod tests {
         let bad_addr = addr.unchecked_add(0x1000);
 
         crate::bytes::tests::check_atomic_accesses(mem, addr, bad_addr);
+    }
+
+    #[cfg(feature = "backend-mmap")]
+    #[test]
+    fn test_guest_memory_mmap_is_hugetlbfs() {
+        let addr = GuestAddress(0x1000);
+        let mem = GuestMemoryMmap::from_ranges(&[(addr, 0x1000)]).unwrap();
+        let r = mem.find_region(addr).unwrap();
+        assert_eq!(r.is_hugetlbfs(), None);
     }
 }
--- a/src/mmap_unix.rs
+++ b/src/mmap_unix.rs
@@ -90,6 +90,7 @@ pub struct MmapRegion {
     prot: i32,
     flags: i32,
     owned: bool,
+    hugetlbfs: Option<bool>,
 }
 
 // Send and Sync aren't automatically inherited for the raw address pointer.
@@ -172,6 +173,7 @@ impl MmapRegion {
             prot,
             flags,
             owned: true,
+            hugetlbfs: None,
         })
     }
 
@@ -213,6 +215,7 @@ impl MmapRegion {
             prot,
             flags,
             owned: false,
+            hugetlbfs: None,
         })
     }
 
@@ -271,6 +274,16 @@ impl MmapRegion {
             }
         }
         false
+    }
+
+    /// Set the hugetlbfs of the region
+    pub fn set_hugetlbfs(&mut self, hugetlbfs: bool) {
+        self.hugetlbfs = Some(hugetlbfs)
+    }
+
+    /// Returns `true` if the region is hugetlbfs
+    pub fn is_hugetlbfs(&self) -> Option<bool> {
+        self.hugetlbfs
     }
 }
 
@@ -352,6 +365,45 @@ mod tests {
             r.flags(),
             libc::MAP_ANONYMOUS | libc::MAP_NORESERVE | libc::MAP_PRIVATE
         );
+    }
+
+    #[test]
+    fn test_mmap_region_set_hugetlbfs() {
+        assert!(MmapRegion::new(0).is_err());
+
+        let size = 4096;
+
+        let r = MmapRegion::new(size).unwrap();
+        assert_eq!(r.size(), size);
+        assert!(r.file_offset().is_none());
+        assert_eq!(r.prot(), libc::PROT_READ | libc::PROT_WRITE);
+        assert_eq!(
+            r.flags(),
+            libc::MAP_ANONYMOUS | libc::MAP_NORESERVE | libc::MAP_PRIVATE
+        );
+        assert_eq!(r.is_hugetlbfs(), None);
+
+        let mut r = MmapRegion::new(size).unwrap();
+        r.set_hugetlbfs(false);
+        assert_eq!(r.size(), size);
+        assert!(r.file_offset().is_none());
+        assert_eq!(r.prot(), libc::PROT_READ | libc::PROT_WRITE);
+        assert_eq!(
+            r.flags(),
+            libc::MAP_ANONYMOUS | libc::MAP_NORESERVE | libc::MAP_PRIVATE
+        );
+        assert_eq!(r.is_hugetlbfs(), Some(false));
+
+        let mut r = MmapRegion::new(size).unwrap();
+        r.set_hugetlbfs(true);
+        assert_eq!(r.size(), size);
+        assert!(r.file_offset().is_none());
+        assert_eq!(r.prot(), libc::PROT_READ | libc::PROT_WRITE);
+        assert_eq!(
+            r.flags(),
+            libc::MAP_ANONYMOUS | libc::MAP_NORESERVE | libc::MAP_PRIVATE
+        );
+        assert_eq!(r.is_hugetlbfs(), Some(true));
     }
 
     #[test]
--- a/src/mmap_windows.rs
+++ b/src/mmap_windows.rs
@@ -175,6 +175,11 @@ impl MmapRegion {
     pub fn file_offset(&self) -> Option<&FileOffset> {
         self.file_offset.as_ref()
     }
+
+    /// This information is not available on Windows platforms.
+    pub fn is_hugetlbfs(&self) -> Option<bool> {
+        None
+    }
 }
 
 impl AsSlice for MmapRegion {
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -456,6 +456,10 @@ impl GuestMemoryRegion for GuestRegionMmap {
         let slice = self.mapping.get_slice(offset.raw_value() as usize, count)?;
         Ok(slice)
     }
+
+    fn is_hugetlbfs(&self) -> Option<bool> {
+        self.mapping.is_hugetlbfs()
+    }
 }
 
 /// [`GuestMemory`](trait.GuestMemory.html) implementation that mmaps the guest's memory


src/guest_memory.rs
```diff
@@ -288,6 +288,11 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
     fn as_volatile_slice(&self) -> Result<volatile_memory::VolatileSlice> {
         self.get_slice(MemoryRegionAddress(0), self.len() as usize)
     }
+
+    /// Returns true if the region is hugepages
+    fn is_hugepages(&self) -> bool {
```
jiangliu, line 293: Does hugepage here mean Hugetlbfs? Or  it also includes Transparent Huge Pages?

teawater: Yes, hugepage means Hugetlbfs.
THPs can be release as normal pages.

jiangliu:
Maybe we lack of some method to attaching some label information to memory regions.  For the `is_hugepage` case, we just store/retrieve a flag associated with a region, and the flag doesn't affect vm-memory behavior. So it would be great to introduce  a new common mechanism to attach label data to regions.

teawater:
> Maybe we lack of some method to attaching some label information to memory regions. For the `is_hugepage` case, we just store/retrieve a flag associated with a region, and the flag doesn't affect vm-memory behavior. So it would be great to introduce a new common mechanism to attach label data to regions.

What about add a new u64 flags to MmapRegion to store the label of a region?

src/guest_memory.rs
jiangliu: Hugepage is a little ambiguous, should we clearly state that it's hugetlbfs?

teawater:
@jiangliu Pushed a new version according to your comments.

alexandruag:
Hi! Are you using something like `memfds` for hugepage-backed areas? (that would explain the `fallocate`). Having a way to express the presence of hugepages at the `vm-memory` interface level (i.e. `GuestMemory` and/or `GuestMemoryRegion`) looks interesting I'm actually not sure what the best approach is (have to think more :D).

However, if you know you're using `MmapRegion`/`GuestRegionMmap` it looks like you can inspect the `flags` member to determine whether `MAP_HUGETLB` is set (I'm not 100% sure how this interacts with file mappings; would be nice if someone readily has more info here). Also, even at a higher level,`MADV_FREE` can be applied to any (anonymous) mapping, `FALLOC_FL_PUNCH_HOLE` to file mappings (if supported by the fs), and a distinction can be made based on the presence or absence of `GuestMemoryRegion::file_offset`. Is any of this helpful for the time being?

teawater:
> Hi! Are you using something like `memfds` for hugepage-backed areas? (that would explain the `fallocate`). Having a way to express the presence of hugepages at the `vm-memory` interface level (i.e. `GuestMemory` and/or `GuestMemoryRegion`) looks interesting I'm actually not sure what the best approach is (have to think more :D).
> 

Yes, I am working on virtio-balloon free page reporting of cloud-hypervisor.  Its areas without backing file are based on memfds.  The areas will setup with MFD_HUGETLB if need.

> However, if you know you're using `MmapRegion`/`GuestRegionMmap` it looks like you can inspect the `flags` member to determine whether `MAP_HUGETLB` is set (I'm not 100% sure how this interacts with file mappings; would be nice if someone readily has more info here). 

Thanks!  I will use it instead of "set_hugetlbfs" in the next version.

> Also, even at a higher level,`MADV_FREE` can be applied to any (anonymous) mapping, `FALLOC_FL_PUNCH_HOLE` to file mappings (if supported by the fs), and a distinction can be made based on the presence or absence of `GuestMemoryRegion::file_offset`. Is any of this helpful for the time being?

I would like to use MADV_FREE as a option but not default. 
The reason is:
1. MADV_FREE is lazy free.
2. MADV_FREE is just work for anon page.

Best,
Hui

jiangliu:
> > Hi! Are you using something like `memfds` for hugepage-backed areas? (that would explain the `fallocate`). Having a way to express the presence of hugepages at the `vm-memory` interface level (i.e. `GuestMemory` and/or `GuestMemoryRegion`) looks interesting I'm actually not sure what the best approach is (have to think more :D).
> 
> Yes, I am working on virtio-balloon free page reporting of cloud-hypervisor. Its areas without backing file are based on memfds. The areas will setup with MFD_HUGETLB if need.
In addition to mmap(MFD_HUGETLB), there are other ways to use hugetlbfs backend. So detecting the MFD_HUGETLB flag may not be concise.

> 
> > However, if you know you're using `MmapRegion`/`GuestRegionMmap` it looks like you can inspect the `flags` member to determine whether `MAP_HUGETLB` is set (I'm not 100% sure how this interacts with file mappings; would be nice if someone readily has more info here).
> 
> Thanks! I will use it instead of "set_hugetlbfs" in the next version.
> 
> > Also, even at a higher level,`MADV_FREE` can be applied to any (anonymous) mapping, `FALLOC_FL_PUNCH_HOLE` to file mappings (if supported by the fs), and a distinction can be made based on the presence or absence of `GuestMemoryRegion::file_offset`. Is any of this helpful for the time being?
> 
> I would like to use MADV_FREE as a option but not default.
> The reason is:
> 
> 1. MADV_FREE is lazy free.
> 2. MADV_FREE is just work for anon page.
> 
> Best,
> Hui

alexandruag:
It's true there are multiple ways to leverage huge pages in general, but specifically within `MmapRegion`, `flags` should catch all of them, right? For example, can we `mmap` a file backed by hugepages without setting that flag?

@teawater, just wondering, why is lazy free worse? Also, with respect to point number 2, you'd use `DONT_NEED` for anon only as well, since it looks like `FALLOC_FL_PUNCH_HOLE` has to be used for `mmap`-ed fds.

That being said, having a high level api to provide page size information sounds interesting. For example I'm wondering whether more information than a simple boolean flag is useful/necessary (i.e. the actual page size). Also, looks like Windows supports something similar but they're called large pages. I would def like to think about his more. @teawater is your use case blocked the existing `vm-memory` implementation not supporting that information? There's also the possibility of managing an external (with respect to `vm_memory`) mapping of addresses/regions to huge page information. Can something like that work?

teawater:
> It's true there are multiple ways to leverage huge pages in general, but specifically within `MmapRegion`, `flags` should catch all of them, right? For example, can we `mmap` a file backed by hugepages without setting that flag?

Yes.
And memfd with MFD_HUGETLB doesn't need it either.

> 
> @teawater, just wondering, why is lazy free worse? 

lazy free is not suitable for many environments because the pages will be released when memory pressure.
But memory pressure may cause some performance problems.
Many systems try to avoid going into memory stressed states.
That is why I think MADV_FREE should not be the default operation.

> Also, with respect to point number 2, you'd use `DONT_NEED` for anon only as well, since it looks like `FALLOC_FL_PUNCH_HOLE` has to be used for `mmap`-ed fds.
> 

Thanks for your reminding.

> That being said, having a high level api to provide page size information sounds interesting. For example I'm wondering whether more information than a simple boolean flag is useful/necessary (i.e. the actual page size). Also, looks like Windows supports something similar but they're called large pages. I would def like to think about his more. @teawater is your use case blocked the existing `vm-memory` implementation not supporting that information? There's also the possibility of managing an external (with respect to `vm_memory`) mapping of addresses/regions to huge page information. Can something like that work?

I think it is not about page size but page type.
Linux kernel provides different release pages(keep vma) API for hugetlb pages and others.

alexandruag:
Thanks for the details! I'll have a better look at these things and return to the thread afterwards. Meanwhile, it would be great if other folks chime in as well.

EmeraldShift:
> Maybe we lack of some method to attaching some label information to memory regions. For the `is_hugepage` case, we just store/retrieve a flag associated with a region, and the flag doesn't affect vm-memory behavior. So it would be great to introduce a new common mechanism to attach label data to regions.

This makes sense, but for this specific issue I'm not understanding the purpose behind having `set_hugetlbfs`, independent of the region constructor.  In my mind, it would make more sense for a flag/option to be passed in on region construction that specifies the desired backing behavior (base pages, THP, memfd/hugetlbfs, etc.). That way, the flag (and `is_hugetlbfs`) would always be consistent with the region's behavior, and we could combine this PR's API (an indicator of the region's hugepage status) with #118 to provide a configurable hugepage API to VMMs.

jiangliu:
> > Maybe we lack of some method to attaching some label information to memory regions. For the `is_hugepage` case, we just store/retrieve a flag associated with a region, and the flag doesn't affect vm-memory behavior. So it would be great to introduce a new common mechanism to attach label data to regions.
> 
> This makes sense, but for this specific issue I'm not understanding the purpose behind having `set_hugetlbfs`, independent of the region constructor. In my mind, it would make more sense for a flag/option to be passed in on region construction that specifies the desired backing behavior (base pages, THP, memfd/hugetlbfs, etc.). That way, the flag (and `is_hugetlbfs`) would always be consistent with the region's behavior, and we could combine this PR's API (an indicator of the region's hugepage status) with #118 to provide a configurable hugepage API to VMMs.

Our current solution is to create file descriptors by the vmm and build memory regions by using the MmapRegion::build() interface. In other words, we have moved the dirty work into vmm:)

alexandruag:
Hi again! Seems like the current iteration of the PR is the most straightforward thing we can do right now to provide the desired functionality. Like @EmeraldShift mentioned, the `set` method looks a bit out of place, but it's prob the best option until we come up with a new iteration of the interfaces. 

Just to clarify the semantics here a bit, if a region is backed by transparent huge pages, the `is_hugepages` method as defined here should return `false` right? Would be nice to have a clearer description of the expected behaviour as part of the documentation. Also, does anyone know more about  any potential correspondence between huge pages on Linux and large pages on Windows? Should we only present this functionality on platforms where it's relevant (i.e. via conditional compilation options)?

teawater:
> Hi again! Seems like the current iteration of the PR is the most straightforward thing we can do right now to provide the desired functionality. Like @EmeraldShift mentioned, the `set` method looks a bit out of place, but it's prob the best option until we come up with a new iteration of the interfaces.
> 
> Just to clarify the semantics here a bit, if a region is backed by transparent huge pages, the `is_hugepages` method as defined here should return `false` right? 
Yes.

> Would be nice to have a clearer description of the expected behaviour as part of the documentation.
What about add the description to https://github.com/rust-vmm/vm-memory/blob/master/DESIGN.md#backend-implementation-based-on-mmap ?

> Also, does anyone know more about any potential correspondence between huge pages on Linux and large pages on Windows? Should we only present this functionality on platforms where it's relevant (i.e. via conditional compilation options)?

alexandruag:
Sounds good!

teawater:
@alexandruag Added a new commit for doc.

DESIGN.md
```diff
@@ -122,6 +122,18 @@ let buf = &mut [0u8; 5];
 let result = guest_memory_mmap.write(buf, addr);
 ```
 
+One of the responsibilities of `GuestRegionMmap` is to provide an API
+`is_hugetlbfs` to show if the region is based on the HugeTLBFS.
+Linux releases normal pages and HugeTLBFS pages require different
```
jiangliu, line 127: require -> requiring?

src/guest_memory.rs
```diff
@@ -288,6 +288,11 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
     fn as_volatile_slice(&self) -> Result<volatile_memory::VolatileSlice> {
         self.get_slice(MemoryRegionAddress(0), self.len() as usize)
     }
+
+    /// Returns true if the region is hugetlbfs
```
jiangliu, line 292: Returns true if the region is backed by hugetlbfs?

DESIGN.md
teawater: Fixed.

src/guest_memory.rs
teawater: Updated.

jiangliu (dismissed):

aghecenco:
Hi @teawater, we're having some trouble with the CI, so builds might be failing or hanging. We' re sorry for the inconvenience, we'll update you here once it's back up and running.

teawater:
> Hi @teawater, we're having some trouble with the CI, so builds might be failing or hanging. We' re sorry for the inconvenience, we'll update you here once it's back up and running.

@aghecenco OK.  Thanks!

DESIGN.md
```diff
@@ -122,6 +122,18 @@ let buf = &mut [0u8; 5];
 let result = guest_memory_mmap.write(buf, addr);
 ```
 
+One of the responsibilities of `GuestRegionMmap` is to provide an API
```
alexandruag, line 125: On second thought, it's probably best to move something like this description and example to the doc comments  of the `is_hugetlbfs` method from the `GuestMemory` interface, where it gets better visibility.

src/guest_memory.rs
```diff
@@ -288,6 +288,11 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
     fn as_volatile_slice(&self) -> Result<volatile_memory::VolatileSlice> {
         self.get_slice(MemoryRegionAddress(0), self.len() as usize)
     }
+
+    /// Returns true if the region is backed by hugetlbfs
+    fn is_hugetlbfs(&self) -> bool {
```
alexandruag, line 293: Hmm, would it actually make sense to return an `Option<bool>`, where `None` represents that no information is available? In our particular situation (where `is_hugetlbfs` has to be set via the `set_hugetlbfs` accesor), this would also be a natural fit since the field starts out as `None`, and only becomes `true` or `false` if it's explicitly set. Just wondering what other folks think about this approach.

alexandruag (commented):
Hi everyone and apologies for the delay in circling back. Left two more comments that I recently thought about; would be great to hear your opinion on them. After we have a conclusion, we should also add a changelog entry for the additions and decide whether we want the `is_hugetlbs` method to be conditionally compiled/included depending on the OS, and we're good to go.

DESIGN.md
teawater: Updated.

src/guest_memory.rs
teawater: Updated.

src/mmap_unix.rs
```diff
@@ -172,6 +173,7 @@ impl MmapRegion {
             prot,
             flags,
             owned: true,
+            hugetlbfs: false,
```
alexandruag, line 176: Can we have the `hugetlbfs` field of `MmapRegion` as an `Option<bool>` which is `None` initially? (since there's no available information right now, until we explicitly set it with `set_hugetlbfs`)

jiangliu (dismissed):

src/guest_memory.rs
```diff
@@ -289,6 +289,30 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
...
     }
+
+    /// Show if the region is based on the `HugeTLBFS`.
+    /// Returns Some(true) if the region is backed by hugetlbfs.
+    /// None represents that no information is available.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[cfg(feature = "backend-mmap")]
```
andreeaflorescu, line 300: We just merged a PR that updated all the code examples that needed features. Can you please update this example to follow that template?

Changes that would be needed:
- specify in the header that this example is based on the the mmap feature. Here is an example: https://github.com/rust-vmm/vm-memory/blob/576a40678687411907c9782fc620e2dcb3c3aebc/src/guest_memory.rs#L266
- declare the `[cfg(feature = "backend-mmap")]` only one at the begginging of the example, and get rid of the `test_guest_memory_mmap_is_hugetlbfs` function. The code can sit directly in the example.

src/mmap_windows.rs
```diff
@@ -175,6 +175,11 @@ impl MmapRegion {
     pub fn file_offset(&self) -> Option<&FileOffset> {
         self.file_offset.as_ref()
     }
+
+    /// Windows doesn't have hugetlbfs.
```
andreeaflorescu, line 179: Can we rephrase this a bit? Right now it sounds more like a comment than documentation for the public interface. How about something like: "This information is not available on Windows platforms."?

src/guest_memory.rs
teawater: Updated.

src/mmap_windows.rs
teawater: Updated.

src/guest_memory.rs
andreeaflorescu: The `test_guest_memory_mmap_is_hugetlbs` function is still there. To get rid of it, and also declare the `#[cfg(feature = "backend-mmap")]`, we would need to start a code block (and have it all conditionally compiled only for that feature).
```rust
    /// # {
    /// # use vm_memory::{GuestAddress, GuestMemory, GuestMemoryMmap, GuestRegionMmap};
    ///
    /// let addr = GuestAddress(0x1000);
    /// let mem = GuestMemoryMmap::from_ranges(&[(addr, 0x1000)]).unwrap();
    /// let r = mem.find_region(addr).unwrap();
    /// assert_eq!(r.is_hugetlbfs(), None);
    /// # }
```

With this approach, you no longer need the function `test_guest_memory_mmap_is_hugetlbfs`, and all the cfg macros.

jiangliu (approved):

src/guest_memory.rs
teawater: Updated.

src/guest_memory.rs
```diff
@@ -1199,4 +1219,13 @@ mod tests {
 
         crate::bytes::tests::check_atomic_accesses(mem, addr, bad_addr);
     }
+
+    #[cfg(feature = "backend-mmap")]
+    #[test]
+    fn test_guest_memory_mmap_is_hugetlbfs() {
```
elmarco, line 1225: Duplicates the test in comment. (as pointed out before)

elmarco:
Why not add a `is_hugetlbs()` to `GuestRegionMmap` and query the underlying `MmapRegion` flags?

src/guest_memory.rs
andreeaflorescu: `cargo fmt` fails because of alignment. The code needs to be aligned as in my previous comment. We can fix it in a subsequent PR.

src/guest_memory.rs
```diff
@@ -289,6 +289,26 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
...
+    /// # #[cfg(feature = "backend-mmap")]
+    /// # {
+    /// #   use vm_memory::{GuestAddress, GuestMemory, GuestMemoryMmap, GuestRegionMmap};
+    ///     let addr = GuestAddress(0x1000);
+    ///     let mem = GuestMemoryMmap::from_ranges(&[(addr, 0x1000)]).unwrap();
+    ///     let r = mem.find_region(addr).unwrap();
+    ///     assert_eq!(r.is_hugetlbfs(), None);
+    /// # }
+    /// ```
+    fn is_hugetlbfs(&self) -> Option<bool> {
```
andreeaflorescu, line 309: Sorry for the late comment, but can we actually just use conditional compilation here and make it available only on Linux, instead of having a blank implementation for Windows?

andreeaflorescu:
> Why not add a `is_hugetlbs()` to `GuestRegionMmap` and query the underlying `MmapRegion` flags?

I think it might makes sense to have it part of the interface in case this information is needed by crates that do not take a dependency on the mmap backend feature.

elmarco:
> > Why not add a `is_hugetlbs()` to `GuestRegionMmap` and query the underlying `MmapRegion` flags?
> 
> I think it might makes sense to have it part of the interface in case this information is needed by crates that do not take a dependency on the mmap backend feature.

I see. But can't we remove the need for set_hugetlbfs() and query the underlying memory instead? (via some trait)

andreeaflorescu:
> I see. But can't we remove the need for set_hugetlbfs() and query the underlying memory instead? (via some trait)

This is an interesting point, that would make the implementation nicer. I do not really know how could we achieve this, so any ideas would be welcome. @bonzini @jiangliu @alexandruag do you have a suggestion?

jiangliu:
> > I see. But can't we remove the need for set_hugetlbfs() and query the underlying memory instead? (via some trait)
> 
> This is an interesting point, that would make the implementation nicer. I do not really know how could we achieve this, so any ideas would be welcome. @bonzini @jiangliu @alexandruag do you have a suggestion?

I don't know any easy way to figure out whether a fd is backed by hugetlbfs:(
We could get file name from /proc/fd/[fdnum], then check the filename. But that's too complicated.

elmarco:
> > > I see. But can't we remove the need for set_hugetlbfs() and query the underlying memory instead? (via some trait)
> > 
> > 
> > This is an interesting point, that would make the implementation nicer. I do not really know how could we achieve this, so any ideas would be welcome. @bonzini @jiangliu @alexandruag do you have a suggestion?
> 
> I don't know any easy way to figure out whether a fd is backed by hugetlbfs:(
> We could get file name from /proc/fd/[fdnum], then check the filename. But that's too complicated.

I was thinking of simply relying on the flags that were given to mmap().

jiangliu:
> > > > I see. But can't we remove the need for set_hugetlbfs() and query the underlying memory instead? (via some trait)
> > > 
> > > 
> > > This is an interesting point, that would make the implementation nicer. I do not really know how could we achieve this, so any ideas would be welcome. @bonzini @jiangliu @alexandruag do you have a suggestion?
> > 
> > 
> > I don't know any easy way to figure out whether a fd is backed by hugetlbfs:(
> > We could get file name from /proc/fd/[fdnum], then check the filename. But that's too complicated.
> 
> I was thinking of simply relying on the flags that were given to mmap().

It works for anonymous memory backed by hugettlbfs by detecting the `MAP_HUGETLB` flag, but it doesn't work when passing a prepared hugetlbfs fd to MmapRegion::build().

andreeaflorescu (approved):
Since this is a long standing PR, let's just merge it, and submit a PR with the required changes afterwards.

jiangliu (approved):

PR #125, opened 2020-12-01
Add dirty bitmap tracking abstractions

alexandruag:
This PR contains an initial proposal to tackle the problem of dirty page/region tracking within a `GuestMemory` object, which happens at the level of each individual region here. We add the `Bitmap` interface, the `BitmapSlice` abstraction (which is passed to objects such as `VolatileSlice`s to enable tracking local accesses in the context of the outer `GuestMemoryRegion`), and begin to add the necessary invocations of tracking functionality  around write accesses. Here are some details and open questions:

- The PR also removes several things (for example, the `VolatileMemory` impls for `MmapRegion` and `&mut [u8]`). I did a quick investigation and it looked like these implementations where not specifically used outside the crate, and I think this also simplifies things a bit. They can be added back and made compatible with the current changes if necessary.

- An open question is what to do (w.r.t. tracking) with methods that return mutable slices or references (the bigger question is whether we should actually have them at all, but that's outside of the current scope :D). My preference is to add the lack of tracking to the safety contract of methods like `as_mut_slice` (i.e. the caller must manually mark the areas that get written to), and to remove (either completely or just from the public interface) methods such as `get_atomic_ref` and `aligned_as_mut` from `VolatileMemory`. One alternative is to treat them like `as_mut_slice`, but this kinda means we're increasing the number of moving parts that have to be taken into account, and most of their use cases are already covered by the `Bytes` interface (there's just `load` and `store` for atomic ops right now, but we can add the rest as well if required). WDYT?

- A potentially significant (and breaking) change is the additional generic type parameter for `Volatile{Slice,Ref,ArrayRef}`. Its purpose is turning this into a zero-cost abstraction (as far as accesses are concerned), and to allow different tracking implementations to be used. The change can become unwieldy if there are a lot of fields/variable definitions that explicitly mention the type, and one way around this is to create a type alias instead of replacing everything. `GuestMemoryMmap` and `GuestRegionMmap` also get an additional type parameter which represents the bitmap implementation in use.

Looking forward to any comments!  If this makes sense, I'll start adding the scaffolding required to test things for generic`Bytes` implementations as well as specific mutable accesses that are enabled by the various `vm-memory` primitives.

### DIFF
--- a/src/bitmap/backend/atomic_bitmap.rs
+++ b/src/bitmap/backend/atomic_bitmap.rs
@@ -147,10 +147,12 @@ impl NewBitmap for AtomicBitmap {
 
 #[cfg(test)]
 mod tests {
+    use super::*;
+
+    use crate::bitmap::tests::test_bitmap;
+
     #[test]
     fn test_bitmap_basic() {
-        use super::AtomicBitmap;
-
         // Test that bitmap size is properly rounded up.
         let a = AtomicBitmap::new(1025, 128);
         assert_eq!(a.len(), 9);
@@ -175,7 +177,6 @@ mod tests {
 
     #[test]
     fn test_bitmap_out_of_range() {
-        use super::AtomicBitmap;
         let b = AtomicBitmap::new(1024, 1);
         // Set a partial range that goes beyond the end of the bitmap
         b.set_addr_range(768, 512);
@@ -183,5 +184,11 @@ mod tests {
         // The bitmap is never set beyond its end.
         assert!(!b.is_addr_set(1024));
         assert!(!b.is_addr_set(1152));
+    }
+
+    #[test]
+    fn test_bitmap_impl() {
+        let b = AtomicBitmap::new(0x2000, 128);
+        test_bitmap(&b);
     }
 }
--- a/src/bitmap/backend/atomic_bitmap.rs
+++ b/src/bitmap/backend/atomic_bitmap.rs
@@ -1,0 +1,187 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+//! Bitmap backend implementation based on atomic integers.
+
+use std::sync::atomic::{AtomicU64, Ordering};
+
+use crate::bitmap::{Bitmap, RefSlice, WithBitmapSlice};
+
+#[cfg(feature = "backend-mmap")]
+use crate::mmap::NewBitmap;
+
+/// `AtomicBitmap` implements a simple bit map on the page level with test and set operations.
+/// It is page-size aware, so it converts addresses to page numbers before setting or clearing
+/// the bits.
+#[derive(Debug)]
+pub struct AtomicBitmap {
+    map: Vec<AtomicU64>,
+    size: usize,
+    page_size: usize,
+}
+
+#[allow(clippy::len_without_is_empty)]
+impl AtomicBitmap {
+    /// Create a new bitmap of `byte_size`, with one bit per page. This is effectively
+    /// rounded up, and we get a new vector of the next multiple of 64 bigger than `bit_size`.
+    pub fn new(byte_size: usize, page_size: usize) -> Self {
+        let mut num_pages = byte_size / page_size;
+        if byte_size % page_size > 0 {
+            num_pages += 1;
+        }
+
+        // Adding one entry element more just in case `num_pages` is not a multiple of `64`.
+        let map_size = num_pages / 64 + 1;
+        let map: Vec<AtomicU64> = (0..map_size).map(|_| AtomicU64::new(0)).collect();
+
+        AtomicBitmap {
+            map,
+            size: num_pages,
+            page_size,
+        }
+    }
+
+    /// Is bit `n` set? Bits outside the range of the bitmap are always unset.
+    pub fn is_bit_set(&self, index: usize) -> bool {
+        if index < self.size {
+            (self.map[index >> 6].load(Ordering::Acquire) & (1 << (index & 63))) != 0
+        } else {
+            // Out-of-range bits are always unset.
+            false
+        }
+    }
+
+    /// Is the bit corresponding to address `addr` set?
+    pub fn is_addr_set(&self, addr: usize) -> bool {
+        self.is_bit_set(addr / self.page_size)
+    }
+
+    /// Set a range of `len` bytes starting at `start_addr`. The first bit set in the bitmap
+    /// is for the page corresponding to `start_addr`, and the last bit that we set corresponds
+    /// to address `start_addr + len - 1`.
+    pub fn set_addr_range(&self, start_addr: usize, len: usize) {
+        // Return early in the unlikely event that `len == 0` so the `len - 1` computation
+        // below does not underflow.
+        if len == 0 {
+            return;
+        }
+
+        let first_bit = start_addr / self.page_size;
+        // Handle input ranges where `start_addr + len - 1` would otherwise overflow an `usize`
+        // by ignoring pages at invalid addresses.
+        let last_bit = start_addr.saturating_add(len - 1) / self.page_size;
+        for n in first_bit..=last_bit {
+            if n >= self.size {
+                // Attempts to set bits beyond the end of the bitmap are simply ignored.
+                break;
+            }
+            self.map[n >> 6].fetch_or(1 << (n & 63), Ordering::SeqCst);
+        }
+    }
+
+    /// Get the length of the bitmap in bits (i.e. in how many pages it can represent).
+    pub fn len(&self) -> usize {
+        self.size
+    }
+
+    /// Reset all bitmap bits to 0.
+    pub fn reset(&self) {
+        for it in self.map.iter() {
+            it.store(0, Ordering::Release);
+        }
+    }
+}
+
+impl Clone for AtomicBitmap {
+    fn clone(&self) -> Self {
+        let map = self
+            .map
+            .iter()
+            .map(|i| i.load(Ordering::Acquire))
+            .map(AtomicU64::new)
+            .collect();
+        AtomicBitmap {
+            map,
+            size: self.size,
+            page_size: self.page_size,
+        }
+    }
+}
+
+impl<'a> WithBitmapSlice<'a> for AtomicBitmap {
+    type S = RefSlice<'a, Self>;
+}
+
+impl Bitmap for AtomicBitmap {
+    fn mark_dirty(&self, offset: usize, len: usize) {
+        self.set_addr_range(offset, len)
+    }
+
+    fn dirty_at(&self, offset: usize) -> bool {
+        self.is_addr_set(offset)
+    }
+
+    fn slice_at(&self, offset: usize) -> <Self as WithBitmapSlice>::S {
+        RefSlice::new(self, offset)
+    }
+}
+
+impl Default for AtomicBitmap {
+    fn default() -> Self {
+        AtomicBitmap::new(0, 0x1000)
+    }
+}
+
+#[cfg(feature = "backend-mmap")]
+impl NewBitmap for AtomicBitmap {
+    fn with_len(len: usize) -> Self {
+        use std::convert::TryFrom;
+
+        // There's no unsafe potential in calling this function.
+        let page_size = unsafe { libc::sysconf(libc::_SC_PAGE_SIZE) };
+        // The `unwrap` is safe to use because the above call should always succeed on the
+        // supported platforms, and the size of a page will always fit within a `usize`.
+        AtomicBitmap::new(len, usize::try_from(page_size).unwrap())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_bitmap_basic() {
+        use super::AtomicBitmap;
+
+        // Test that bitmap size is properly rounded up.
+        let a = AtomicBitmap::new(1025, 128);
+        assert_eq!(a.len(), 9);
+
+        let b = AtomicBitmap::new(1024, 128);
+        assert_eq!(b.len(), 8);
+        b.set_addr_range(128, 129);
+        assert!(!b.is_addr_set(0));
+        assert!(b.is_addr_set(128));
+        assert!(b.is_addr_set(256));
+        assert!(!b.is_addr_set(384));
+
+        let copy_b = b.clone();
+        assert!(copy_b.is_addr_set(256));
+        assert!(!copy_b.is_addr_set(384));
+
+        b.reset();
+        assert!(!b.is_addr_set(128));
+        assert!(!b.is_addr_set(256));
+        assert!(!b.is_addr_set(384));
+    }
+
+    #[test]
+    fn test_bitmap_out_of_range() {
+        use super::AtomicBitmap;
+        let b = AtomicBitmap::new(1024, 1);
+        // Set a partial range that goes beyond the end of the bitmap
+        b.set_addr_range(768, 512);
+        assert!(b.is_addr_set(768));
+        // The bitmap is never set beyond its end.
+        assert!(!b.is_addr_set(1024));
+        assert!(!b.is_addr_set(1152));
+    }
+}
--- a/src/bitmap/backend/atomic_bitmap.rs
+++ b/src/bitmap/backend/atomic_bitmap.rs
@@ -84,6 +84,14 @@ impl AtomicBitmap {
         self.size
     }
 
+    /// Atomically get and reset the dirty page bitmap.
+    pub fn get_and_reset(&self) -> Vec<u64> {
+        self.map
+            .iter()
+            .map(|u| u.fetch_and(0, Ordering::SeqCst))
+            .collect()
+    }
+
     /// Reset all bitmap bits to 0.
     pub fn reset(&self) {
         for it in self.map.iter() {
@@ -193,6 +201,16 @@ mod tests {
         assert!(!b.is_addr_set(128));
         assert!(!b.is_addr_set(256));
         assert!(!b.is_addr_set(384));
+
+        b.set_addr_range(128, 129);
+        let v = b.get_and_reset();
+
+        assert!(!b.is_addr_set(128));
+        assert!(!b.is_addr_set(256));
+        assert!(!b.is_addr_set(384));
+
+        assert_eq!(v.len(), 1);
+        assert_eq!(v[0], 0b110);
     }
 
     #[test]
--- a/src/bitmap/backend/atomic_bitmap.rs
+++ b/src/bitmap/backend/atomic_bitmap.rs
@@ -137,8 +137,28 @@ impl NewBitmap for AtomicBitmap {
     fn with_len(len: usize) -> Self {
         use std::convert::TryFrom;
 
-        // There's no unsafe potential in calling this function.
-        let page_size = unsafe { libc::sysconf(libc::_SC_PAGE_SIZE) };
+        let page_size;
+
+        #[cfg(unix)]
+        {
+            // There's no unsafe potential in calling this function.
+            page_size = unsafe { libc::sysconf(libc::_SC_PAGE_SIZE) };
+        }
+
+        #[cfg(windows)]
+        {
+            use winapi::um::sysinfoapi::{GetSystemInfo, LPSYSTEM_INFO, SYSTEM_INFO};
+
+            // It's safe to initialize this object from a zeroed memory region.
+            let mut sysinfo: SYSTEM_INFO = unsafe { std::mem::zeroed() };
+
+            // It's safe to call this method as the pointer is based on the address
+            // of the previously initialized `sysinfo` object.
+            unsafe { GetSystemInfo(&mut sysinfo as LPSYSTEM_INFO) };
+
+            page_size = sysinfo.dwPageSize;
+        }
+
         // The `unwrap` is safe to use because the above call should always succeed on the
         // supported platforms, and the size of a page will always fit within a `usize`.
         AtomicBitmap::new(len, usize::try_from(page_size).unwrap())
--- a/src/bitmap/mod.rs
+++ b/src/bitmap/mod.rs
@@ -6,9 +6,15 @@
 //! `GuestMemoryRegion` object, and the resulting bitmaps can then be aggregated to build the
 //! global view for an entire `GuestMemory` object.
 
+#[cfg(any(test, feature = "backend-bitmap"))]
+mod backend;
+
 use std::fmt::Debug;
 
 use crate::{GuestMemory, GuestMemoryRegion};
+
+#[cfg(any(test, feature = "backend-bitmap"))]
+pub use backend::RefSlice;
 
 /// Trait implemented by types that support creating `BitmapSlice` objects.
 pub trait WithBitmapSlice<'a> {
--- a/src/bitmap/mod.rs
+++ b/src/bitmap/mod.rs
@@ -14,7 +14,7 @@ use std::fmt::Debug;
 use crate::{GuestMemory, GuestMemoryRegion};
 
 #[cfg(any(test, feature = "backend-bitmap"))]
-pub use backend::RefSlice;
+pub use backend::{AtomicBitmap, RefSlice};
 
 /// Trait implemented by types that support creating `BitmapSlice` objects.
 pub trait WithBitmapSlice<'a> {
--- a/src/bitmap/mod.rs
+++ b/src/bitmap/mod.rs
@@ -109,3 +109,311 @@ pub type BS<'a, B> = <B as WithBitmapSlice<'a>>::S;
 /// Helper type alias for referring to the `BitmapSlice` concrete type associated with
 /// the memory regions of an object `M: GuestMemory`.
 pub type MS<'a, M> = BS<'a, <<M as GuestMemory>::R as GuestMemoryRegion>::B>;
+
+#[cfg(test)]
+pub(crate) mod tests {
+    use super::*;
+
+    use std::io::Cursor;
+    use std::marker::PhantomData;
+    use std::mem::size_of_val;
+    use std::result::Result;
+    use std::sync::atomic::Ordering;
+
+    use crate::{Bytes, VolatileMemory};
+    #[cfg(feature = "backend-mmap")]
+    use crate::{GuestAddress, MemoryRegionAddress};
+
+    // Helper method to check whether a specified range is clean.
+    pub fn range_is_clean<B: Bitmap>(b: &B, start: usize, len: usize) -> bool {
+        (start..start + len).all(|offset| !b.dirty_at(offset))
+    }
+
+    // Helper method to check whether a specified range is dirty.
+    pub fn range_is_dirty<B: Bitmap>(b: &B, start: usize, len: usize) -> bool {
+        (start..start + len).all(|offset| b.dirty_at(offset))
+    }
+
+    pub fn check_range<B: Bitmap>(b: &B, start: usize, len: usize, clean: bool) -> bool {
+        if clean {
+            range_is_clean(b, start, len)
+        } else {
+            range_is_dirty(b, start, len)
+        }
+    }
+
+    // Helper method that tests a generic `B: Bitmap` implementation. It assumes `b` covers
+    // an area of length at least 0x2000.
+    pub fn test_bitmap<B: Bitmap>(b: &B) {
+        let len = 0x2000;
+        let dirty_offset = 0x1000;
+        let dirty_len = 0x100;
+
+        // Some basic checks.
+        let s = b.slice_at(dirty_offset);
+
+        assert!(range_is_clean(b, 0, len));
+        assert!(range_is_clean(&s, 0, dirty_len));
+
+        b.mark_dirty(dirty_offset, dirty_len);
+        assert!(range_is_dirty(b, dirty_offset, dirty_len));
+        assert!(range_is_dirty(&s, 0, dirty_len));
+    }
+
+    #[derive(Debug)]
+    pub enum TestAccessError {
+        RangeCleanCheck,
+        RangeDirtyCheck,
+    }
+
+    // A helper object that implements auxiliary operations for testing `Bytes` implementations
+    // in the context of dirty bitmap tracking.
+    struct BytesHelper<F, G, M> {
+        check_range_fn: F,
+        address_fn: G,
+        phantom: PhantomData<*const M>,
+    }
+
+    // `F` represents a closure the checks whether a specified range associated with the `Bytes`
+    // object that's being tested is marked as dirty or not (depending on the value of the last
+    // parameter). It has the following parameters:
+    // - A reference to a `Bytes` implementations that's subject to testing.
+    // - The offset of the range.
+    // - The length of the range.
+    // - Whether we are checking if the range is clean (when `true`) or marked as dirty.
+    //
+    // `G` represents a closure that translates an offset into an address value that's
+    // relevant for the `Bytes` implementation being tested.
+    impl<F, G, M, A> BytesHelper<F, G, M>
+    where
+        F: Fn(&M, usize, usize, bool) -> bool,
+        G: Fn(usize) -> A,
+        M: Bytes<A>,
+    {
+        fn check_range(&self, m: &M, start: usize, len: usize, clean: bool) -> bool {
+            (self.check_range_fn)(m, start, len, clean)
+        }
+
+        fn address(&self, offset: usize) -> A {
+            (self.address_fn)(offset)
+        }
+
+        fn test_access<Op>(
+            &self,
+            bytes: &M,
+            dirty_offset: usize,
+            dirty_len: usize,
+            op: Op,
+        ) -> Result<(), TestAccessError>
+        where
+            Op: Fn(&M, A),
+        {
+            if !self.check_range(bytes, dirty_offset, dirty_len, true) {
+                return Err(TestAccessError::RangeCleanCheck);
+            }
+
+            op(bytes, self.address(dirty_offset));
+
+            if !self.check_range(bytes, dirty_offset, dirty_len, false) {
+                return Err(TestAccessError::RangeDirtyCheck);
+            }
+
+            Ok(())
+        }
+    }
+
+    // `F` and `G` stand for the same closure types as described in the `BytesHelper` comment.
+    // The `step` parameter represents the offset that's added the the current address after
+    // performing each access. It provides finer grained control when testing tracking
+    // implementations that aggregate entire ranges for accounting purposes (for example, doing
+    // tracking at the page level).
+    pub fn test_bytes<F, G, M, A>(bytes: &M, check_range_fn: F, address_fn: G, step: usize)
+    where
+        F: Fn(&M, usize, usize, bool) -> bool,
+        G: Fn(usize) -> A,
+        A: Copy,
+        M: Bytes<A>,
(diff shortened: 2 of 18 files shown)


rbradford:
@alexandruag We have live migration working in Cloud Hypervisor now taking advantage of the KVM region tracking. I did wonder if we could be in a situation where updates from the VMM side i.e. via GuestMemory might be an issue. The fact you've created this PR makes me think you're seeing this in practice.

rbradford:
@alexandruag It looks good. But don't need a way to atomically clear the bitmap? Like how the `KVM_GET_DIRTY_LOG` ioctl does so that for live migration you can only transfer the needed pages.

alexandruag:
Hi Rob, thanks for the quick reply! Yeah, during previous experiments we noticed the KVM dirty log did not catch some of the write accesses performed by the VMM (which makes sense if stuff like PML is used). Regarding your second question, the `Bitmap` interface here is mainly trying to define the methods required by `GuestMemory` backends to record tracking information. Concrete `Bitmap` implementations will most likely have  additional functionality like the one you mention, and we can def augment the interface to support those as well; I just wanted to start by looking at what it takes to interact with `GuestMemory` implementations.

src/bitmap.rs
```diff
@@ -0,0 +1,85 @@
...
+//! length to a given offset. In particular, this is used to track write accesses within a
+//! `GuestMemoryRegion` object, and the resulting bitmaps can then be aggregated to build the
+//! global view for the entire `GuestMemory`.
+
+use std::fmt::{self, Debug};
+
+/// Common bitmap operations.
+// This is just an initial proposal; we'll add other methods if they need to be part of the main
+// interface as well.
+pub trait Bitmap: Default {
```
jiangliu, line 11: How about naming it as DirtyMap?

src/bitmap.rs
```diff
@@ -0,0 +1,85 @@
...
+    // For now this is mostly useful to preserve compatibility with the old `GuestMemoryMmap` (and
+    // potentially other backends) constructors, which are not bitmap-aware. We can remove this
+    // from the interface (and apply changes elsewhere) if it's too restrictive and/or configuring
+    // the bitmap post-creation (via `GuestMemoryRegion::bitmap_mut`) gets unwieldy.
+    /// Create a new bitmap of the given size.
+    fn new(size: usize) -> Self;
```
jiangliu, line 17: We have already have Default for Bitmap, which should create a default Bitmap in disabled state. Later we could update the Bitmap instance associated with a region to enable/disable the dirty logging. So the new() method may be redudant.

src/bitmap.rs
```diff
@@ -0,0 +1,85 @@
...
+
+/// A no-op `Bitmap` implementation that can be provided to backends that do not actually
+/// require the tracking functionality.
+impl Bitmap for () {
+    fn new(_size: usize) -> Self {
+        ()
+    }
+
+    fn mark_dirty(&self, _offset: usize, _len: usize) {}
+}
```
jiangliu, line 31: How about adding "impl<T: Bitmap> Bitmap for Option<T> "for easily disable/enable the dirty logging?

src/bitmap.rs
```diff
@@ -0,0 +1,85 @@
...
+
+/// Represents a slice into a `Bitmap` object, starting at `base_offset`.
+pub struct BitmapSlice<'a, B> {
+    inner: &'a B,
+    base_offset: usize,
+}
+
+impl<'a, B: Bitmap> BitmapSlice<'a, B> {
```
jiangliu, line 39: With Bitmap::new() removed, we could impl Bitmap for BitmapSlice, which should be convenient:)

src/bitmap.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// Mark the memory range specified by the given `offset` (relative to the base offset of
+    /// the slice) and `len` as dirtied.
+    pub fn mark_dirty(&self, offset: usize, len: usize) {
+        // The `Bitmap` operations are supposed to accompany guest memory accesses defined by the
+        // same parameters (i.e. offset & length), so we use simple wrapping arithmetic instead of
+        // performing additional checks. If an overflow would occur, we simply end up marking some
+        // other region as dirty (which is just a false positive) instead of a region that could
+        // not have been accessed to begin with.
+        self.inner
+            .mark_dirty(self.base_offset.wrapping_add(offset), len)
```
jiangliu, line 57: a normal Add instead of wrapping_add() here should be OK, right?
After all we don't support address wrap around for a region.

src/mmap_unix.rs
```diff
@@ -259,8 +258,8 @@ impl MmapRegion {
                 if f_off1.file().as_raw_fd() == f_off2.file().as_raw_fd() {
                     let s1 = f_off1.start();
                     let s2 = f_off2.start();
-                    let l1 = self.len() as u64;
-                    let l2 = other.len() as u64;
+                    let l1 = self.size() as u64;
```
jiangliu, line 261: Could you please help to split out the change of self.len() to self.size()?

src/mmap_unix.rs
```diff
@@ -289,23 +288,6 @@ impl AsSlice for MmapRegion {
     }
 }
 
-impl VolatileMemory for MmapRegion {
```
jiangliu, line 292: This and related changes to simplify current implementation deserves a dedicated PR so we could do it step by step:)

jiangliu:
Seems it's a chance to get rid of some unused interfaces, we could add them back later when they are really used.
I think the overall design to extend VolatileXxx to assist dirty track should be ok, it's much more safe to let the user of Volatile interfaces do dirty tracking. 
For unsafe interfaces returns *mut u8, it should be ok to let the user track dirty status.

alexandruag:
Thanks for the comments Gerry, I'll start addressing them soon.

src/volatile_memory.rs
```diff
@@ -618,14 +630,14 @@ impl Bytes<usize> for VolatileSlice<'_> {
...
-        }
-        Ok(())
+
+        // It is safe to overwrite the volatile memory. Accessing the guest memory as a mutable
+        // slice is OK because nothing assumes another thread won't change what is loaded. We also
+        // manually update the dirty bitmap below.
+        let dst = unsafe { &mut self.as_mut_slice()[addr..end] };
+
+        self.bitmap.mark_dirty(addr, count);
+        src.read_exact(dst).map_err(Error::IOError)
```
luminitavoicu, line 640: In case the `read_exact` function fails here, we still mark the pages as dirty. Wouldn't it be best to update the bitmap only in case the write succeeds?
```suggestion
        src.read_exact(dst).map_err(Error::IOError)?;
        self.bitmap.mark_dirty(addr, count)
```

src/bitmap.rs
```diff
@@ -0,0 +1,85 @@
...
+impl Bitmap for () {
+    fn new(_size: usize) -> Self {
+        ()
+    }
+
+    fn mark_dirty(&self, _offset: usize, _len: usize) {}
+}
+
+/// Represents a slice into a `Bitmap` object, starting at `base_offset`.
+pub struct BitmapSlice<'a, B> {
```
aghecenco, line 34: Why doesn't a `BitmapSlice` have a size?

src/bitmap.rs
```diff
@@ -0,0 +1,85 @@
...
+pub trait Bitmap: Default {
+    // For now this is mostly useful to preserve compatibility with the old `GuestMemoryMmap` (and
+    // potentially other backends) constructors, which are not bitmap-aware. We can remove this
+    // from the interface (and apply changes elsewhere) if it's too restrictive and/or configuring
+    // the bitmap post-creation (via `GuestMemoryRegion::bitmap_mut`) gets unwieldy.
+    /// Create a new bitmap of the given size.
+    fn new(size: usize) -> Self;
+
+    /// Mark the memory range specified by the given `offset` and `len` as dirtied.
+    fn mark_dirty(&self, offset: usize, len: usize);
```
aghecenco, line 20: What would a `impl Bitmap for Foo` look like - more precisely, what would `Foo` contain to represent the bitmap? Mult be something with interior mutability so you can get away with `&self` here - are you thinking direct pointer manipulation? Splitting the bitmap representation into 64-bit chunks and using `Atomic`s? Something else?

src/volatile_memory.rs
alexandruag: I've added the `mark_dirty` first since even if the memory operation fails, I think we can still have partial writes to guest memory, and they are implicitly covered if the entire range was marked already. It is possible to generate false positives this way, but I think they are not significant errors in the context of tracking dirty memory areas.

alexandruag (commented):
Hi everyone, thanks for the comments! I've updated the PR, and the biggest change is that I used HRTBs (drawing inspiration from @Daniel-Aaron-Bloom's [PR](https://github.com/rust-vmm/vm-memory/pull/130) that was merged a while ago) to turn the tracking into a zero-cost abstraction (no additional operations are performed or space required when a trivial `Bitmap` implementation is used). This does make some signatures a bit more complex, but I've also added some type aliases to help with that.

Instead of the `Bitmap` trait having constructors itself, I've defined a `NewBitmap` trait meant to be used by `GuestRegionMmap` to instantiate bitmap objects without having to change its current interface (i.e. such that bitmaps are moved over at construction time). Another thing I'm considering is to rename `GuestMemoryMmap` and `GuestRegionMmap` to something else, and use the original names for type aliases that imply using a trivial bitmap implementation by default. This should minimize the amount of changes required by `vm-memory` consumers that do not need the new functionality.

The next steps are to fix the unit tests, add some generic tests that can be used with any `Bitmap` implementation, and provide a backend implementation as well. I'll update this PR with the tests as soon as they're done, and I'm looking forward to any feedback regarding the changes so far.

rbradford:
@alexandruag Would you also be willing to include a simple "off the shelf" Bitmap tracking implementation so every user does not need to construct their own?

bonzini:
Sorry for not noticing this PR until now.

There is one important issue with attaching the bitmap to the GuestMemory, which is that the memory addresses can change but the dirty bitmap should not. Think for example of a large RAM region for an emulated graphics card; because it is a PCI BAR, the base address can move freely.

So I'm thinking of a different design where the bitmap is attached to the GuestRegion instead, for example as a decorator where you'd create a `GuestMemory<DirtyTrackedRegion<MmapRegion>>`. Do you think this would be hard to implement?

Regarding removing the `impl VolatileMemory for MmapRegion`, I'm not sure what you gain from removing it; you are basically inlining two methods, but there is almost no simplification. Likewise, `impl VolatileMemory for &'a mut [u8]` has at least a didactic value, as it shows that `VolatileMemory` is something that applies to sequences of bytes. In any case there was an agreement to move those to a separate PR, so there's no need to worry about it now.

alexandruag:
@rbradford Yes, the plan is to include a backend implementation as well (guarded by a feature, like `backend-mmap` for example).

@bonzini Thanks for having a look! The bitmap(s) are already attached to individual regions, and the tracking is done relative to the start of each particular region. Is this different from what you were describing?

bonzini:
> The bitmap(s) are already attached to individual regions

You're right, I misread the code. I have now looked at Firecracker too, and the main difference with my proposal in #140 is that Firecracker handles external bitmaps (such as KVM's) by hand in `SnapshotMemory::dump_dirty`.

alexandruag:
Hi! I’ve updated the PR with backend implementations (`RefSlice` based on the previous `BitmapSlice` struct, and `AtomicBitmap` which is lifted from the current Firecracker `Bitmap` object implementation) and testing. I’ve added generic unit test methods which can be exported at some point so they can be used by other crates/implementations as well.

I’ve reverted the code removals with the exception of the original `VolatileMemory` implementation for `MmapRegion`. The main reason is that `VolatileMemory` now has the `B: Bitmap` associated type which doesn’t seem to fit the `MmapRegion` semantics very well IMO. At the same time, if anyone would like to see this change reverted as well, I’ll add the `VolatileMemory` implementation back, and change `MmapRegion` to `MmapRegion<B>` (similar to what happened to `GuestRegionMmap`).

alexandruag:
@bonzini @jiangliu @aghecenco @rbradford @luminitavoicu can you please take another look? I'll go over the orderings in the `AtomicBitmap` implementation again as well, and some doc tests still need to be fixed.

src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        // There's no unsafe potential in calling this function.
+        let page_size = unsafe { libc::sysconf(libc::_SC_PAGE_SIZE) };
+        // The `unwrap` is safe to use because the above call should always succeed on the
+        // supported platforms, and the size of a page will always fit within a `usize`.
+        AtomicBitmap::new(len, usize::try_from(page_size).unwrap())
+    }
+}
+
+mod tests {
```
aghecenco, line 144: Missing `#[cfg(test)]`. This is causing the Linux unit tests to fail in CI.

aghecenco (commented):
Did a first pass and overall LGTM! I find the naming scheme a bit confusing with all the new traits and implementors, as it's hard to figure out which is which, but I have no better suggestions (others than those in comments). When this gets "promoted" to non-RFC, I think dedicated benchmarks could be useful in demonstrating the *zero* part of the zero cost abstractions.

bonzini:
I'm not sure why `VolatileMemory` would necessarily have a bitmap associated type; the tests show that even an `[u8]` makes sense as a `VolatileMemory` instance.

I'll take a look at making a decorator-like implementation where you have `DirtyTrackedRegion<MmapRegion>`, but being able to use either `MmapRegion<()>` or `MmapRegion<AtomicBitmap>` would be fine as well.

alexandruag:
The `B` associated type for `VolatileMemory` is used to determine what kind of dirty tracking slice is embedded into the `VolatileSlice`, `VolatileRef`, and `VolatileArrayRef` objects that are returned by methods such as `get_slice`, `get_ref`, etc.

alexandruag (commented):
Hi again! I've addressed the recently added comments, and I also pushed a new version where the previous removals around `MmapRegion` are reverted, and the struct now has an inner object used for tracking (i.e. it's now `MmapRegion<B>`). I have to fix some of the doc tests and Windows-specific logic as well, and then I'll bump the status to an actual PR.

I think that adding benchmarks to verify that the newly added abstractions really are zero-cost is a bit heavy-weight; so far I've validated the expectations in this regard mainly by looking at snippets of compiler output (with the regular release optimizations enabled). We can def revisit the idea if folks think it's valuable going forward. 

Looking forward to any additional feedback!

rbradford:
Hi @alexandruag, from our side @likebreath has volunteered to try and integrate this functionality into the CH live migration code. Please be gentle as this is first time interacting the Rust-VMM community :-)

alexandruag:
Hi @likebreath! I've made the remaining changes such that all tests pass and bumped the PR status from RFC to ready for review.

likebreath:
@alexandruag Thanks a lot for the effort. I started to experiment this PR with cloud hypervisor few days ago. The dirty `bitmap` tracking support looks great, and is very easy to adopt.  :) As you mentioned earlier, the major integration effort is from the signature changes on `GuestMemoryMmap`/`GuestRegionMmap`. 

I have a quick question: Does it make sense to add a new method `AtomicBitmap::get_and_rest() -> Vec<u64>` ? 

It will be useful when the consumer of the BitMap (e.g. VMM) needs to retrieve and reset the dirty page tracking while the guest is still running. A quick reference from what I am using with my experiment on cloud hypervisor: https://github.com/likebreath/vm-memory/commit/3659d872c403c7385ceccc5560d7e38b9c470695. Please let me know your thoughts.  

I am in the process of validating whether my integration actually works ( for VM live migration with ongoing VMM writes to the guest memory). I will let you know if I have more questions or comments. 

/cc @rbradford

src/volatile_memory.rs
```diff
@@ -263,10 +286,11 @@ impl<'a> VolatileSlice<'a> {
...
     /// accesses.
-    pub unsafe fn new(addr: *mut u8, size: usize) -> VolatileSlice<'a> {
+    pub unsafe fn new(addr: *mut u8, size: usize, bitmap: B) -> VolatileSlice<'a, B> {
         VolatileSlice {
             addr,
             size,
+            bitmap,
             phantom: PhantomData,
         }
     }
```
likebreath, line 296: @alexandruag I found it was needed to mark the whole slice dirty when using `VolatileSlice::as_ptr()`, as we may have write access to the  returned mutable pointer. One example is the asynchronous operations on virtio-blk queues in [Cloud Hypervisor](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/master/block_util/src/lib.rs#L320-L323).  I made the above change in my [own fork of your branch](https://github.com/likebreath/vm-memory/commits/bitmap_get_and_reset), and tested it with [Cloud Hypervisor](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/2497). Now I have Cloud Hypervisor migrating successfully  a guest workload (with intensive virtio-blk transactions) that was interrupted during live-migration previously (without vm-memory dirty page tracking). 

I didn't send a PR directly to your branch from my fork because the above change in `VolatileSlice::as_ptr()` breaks your unit test `test_volatile_slice_dirty_tracking()`. The reason is `VolatileSlice::subslice()` uses the `as_ptr()` function and  will mark the whole slice dirty. 

For the similar reason, I think  `pub fn as_ptr()` for `VolatileRef` and `VolatileArrayRef` should also mark the referred memory dirty.

Please let me know your thoughts or questions. Thank you.

src/guest_memory.rs
```diff
@@ -836,16 +841,20 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
...
+                let result = loop {
                     match src.read(&mut dst[start..end]) {
                         Ok(n) => break Ok(n),
                         Err(ref e) if e.kind() == std::io::ErrorKind::Interrupted => continue,
                         Err(e) => break Err(Error::IOError(e)),
                     }
-                }
+                };
+
+                region.bitmap().mark_dirty(start, len);
```
lauralt, line 856: There's a `region.write()` below that could benefit from a comment about why we don't manually update the bitmap there as well.

lauralt: Should we mention somewhere that the `GuestMemory` trait assumes the Bytes implementation of the regions will update the bitmap?

alexandruag:
Hi everyone! Apologies for the absence from the PR, I'll answer/address the pending comments over the following days.

src/volatile_memory.rs
alexandruag: Thanks for bringing this up, it's a very interesting discussion. Adding the semantics you propose to the `as_ptr` method(s) is definitely an option, but I think it makes the abstraction quite inflexible, because a single `as_ptr` call will mark an entire slice as dirty, no matter how large it is vs the actual access size. My preferred approach is to explicitly call something like `slice.bitmap().mark_dirty(offset, len)` in conjunction with the use of `slice.as_ptr()` when getting the pointer for write access. This adds an extra step, but allows finer grained control instead of forcing the tracking functionality to assume the entire slice is going to become dirty. Do you think this pattern is compatible with the use cases present in Cloud Hypervisor?

alexandruag (commented):
@likebreath thanks for the integration work; it's definitely great to hear that you can already successfully migrate workloads. In terms of the signature changes, I found that using type aliases can be really helpful (for example, if you used `GuestMemoryMmap` in many places before, you can now define somewhere `type GuestMemoryMmap = vm_memory::GuestMemoryMmap<BitmapImplementation>`, and only alter the use statements where necessary). The `get_and_reset` method looks helpful and we can def pull it into the implementation; I'll take a better look at the implications just to be sure and do that.

Let me know if, by any chance, I missed any of your questions. I'll address Laura's comments soon as well.

src/volatile_memory.rs
likebreath: @alexandruag Thanks for the detailed comments. Yes, I think so. I don't see it is problem to ask the user of `as_ptr` to explicitly set the  `bitmap` in Cloud Hypervisor. I will confirm that in my integration PR and keep you posted.

This pattern makes sense, and I think we need to make it louder/clearer about which functions need explicit bitmap tracking from their users/callers. I remember you added relevant comments to some other functions, and we can do the same to `as_ptr` functions at least. 

Also, for the same reason (to maximize the precision of dirty page tracking), do you think we need to enforce the same pattern in functions `VolatileMemory::aligned_as_mut/get_atomic_ref` (where we are marking dirty the whole slice being returned)?

likebreath: @alexandruag In case you missed it, I want to let you know that the pattern you mentioned above (handling special cases of dirty-page-tracking explicitly from the VMM) works fine with Cloud Hypervisor ([details](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/2497/commits/a535a6470afd90994fec642c10ef0fab840d51ff)).

Also, to complete the live-migration support in Cloud Hypervisor, we need to have the dirty-page-tracking of the `vm-memory` crate sooner than later. So we are planning to pull your PR changes into our own fork of  the `vm-memory` crate (with the addition of `get_and_reset` function), and reply on that fork before your PR is landed in the upstream.

Meanwhile, we are more than happy to continue the discussions here and hopefully we can help with landing the PR faster. Please let me know if you have any questions about my previous comments. I will also keep you posted if we have any new findings with more experiments of live-migration in Cloud Hypervisor using your patches. Thank you.

/cc @rbradford

alexandruag (commented):
Hi! I've addressed the comments and added `get_and_reset` to `AtomicBitmap`. Also, as per the previous suggestion, `VolatileMemory::aligned_as_mut/get_atomic_ref` don't automatically mark the underlying page(s) dirty by default. Instead, `mark_dirty` has to be invoked explicitly for the relevant range (same as for slices). Thanks for the reviews so far, and please have another look.

src/volatile_memory.rs
```diff
@@ -181,7 +189,9 @@ pub trait VolatileMemory {
         Ok(&*(slice.addr as *const T))
     }
 
-    /// Returns a mutable reference to an instance of `T` at `offset`.
+    /// Returns a mutable reference to an instance of `T` at `offset`. Mutable accesses performed
```
lauralt, line 192: Can we add the same comment about the mutable accesses not being marked as dirty to `as_ptr()` documentation as well?

lauralt (approved):

alexandruag (commented):
@likebreath @rbradford is the PR now aligned with the requirements that were previously mentioned? In particular, are things still working well after the adjustment mentioned in [this](https://github.com/rust-vmm/vm-memory/pull/125#discussion_r634682120) comment?

@jiangliu @bonzini are you ok with moving forward with the changes in this PR?

rbradford (dismissed):
@alexandruag Yes. I think that is a great approach.

lauralt (approved):

bonzini:
I still don't understand why it's using a new argument to the `GuestMemory` and `GuestRegion` traits instead of adding a decorator. In particular I don't think it's necessary at all to modify `GuestRegion`... It seems unnecessary to me to force changes in all users. Am I missing something, perhaps it's not necessary to change all the users?

bonzini:
That said, I haven't had the time to do anything myself, so...

alexandruag:
Hi @bonzini! So right now the `GuestMemoryRegion` trait has a `B: Bitmap` associated type primarily to identify the type of `Bitmap` that's embedded within the `VolatileSlice`s returned by methods such as `get_slice` and `as_volatile_slice` (since `VolatileSlice` and the other `Volatile*` abstractions are part of the interface). Do you have any starting point in mind for how to solve this with the decorator approach?

bonzini:
I would move the bitmap part of `VolatileSlice` into another decorator, so the slice type returned by `DirtyLogMemory<M, B>` would be `B::DirtyLogSlice<M::Slice>` or something like that.

src/bitmap/mod.rs
```diff
@@ -0,0 +1,418 @@
...
+// These methods represent the core functionality that's required by `vm-memory` abstractions
+// to implement generic tracking logic, as well as tests that can be reused by different backends.
+pub trait Bitmap: for<'a> WithBitmapSlice<'a> {
+    /// Mark the memory range specified by the given `offset` and `len` as dirtied.
+    fn mark_dirty(&self, offset: usize, len: usize);
+
+    /// Check whether the specified `offset` is marked as dirty.
+    fn dirty_at(&self, offset: usize) -> bool;
+
+    /// Return a `Self::S::Slice` of the current bitmap, starting at the specified `offset`.
```
jiangliu, line 48: Should it be "Self::S" instead of "Self::S::Slice"?

src/guest_memory.rs
```diff
@@ -165,6 +174,9 @@ impl FileOffset {
 /// Represents a continuous region of guest physical memory.
 #[allow(clippy::len_without_is_empty)]
 pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
+    /// Type used for dirty memory tracking.
```
jiangliu, line 177: In case we may add more associated types in future, how about:
```
trait GuestMemoryTypes {
       type B: Bitmap;
}

pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
     type T: GuestMemoryTypes;
}
```
It's much easier to add more associated types in future.

src/mmap.rs
```diff
@@ -125,33 +138,34 @@ pub fn check_file_offset(
 /// Represents a continuous region of the guest's physical memory that is backed by a mapping
 /// in the virtual address space of the calling process.
 #[derive(Debug)]
-pub struct GuestRegionMmap {
-    mapping: MmapRegion,
+pub struct GuestRegionMmap<B> {
```
jiangliu, line 141: Use default generic type parameter may reduce the changes needed.
```
pub struct GuestRegionMmap<B = ()> {
}
```
Also suggest to change to
```
pub struct GuestRegionMmap<T = GuestMemoryAssociatedTypes> {
}
```
to in case we need to add more associated types in future.

src/volatile_memory.rs
```diff
@@ -119,6 +120,9 @@ pub fn compute_offset(base: usize, offset: usize) -> Result<usize> {
 
 /// Types that support raw volatile access to their data.
 pub trait VolatileMemory {
+    /// Type used for dirty memory tracking.
+    type B: Bitmap;
```
jiangliu, line 124: This change will break too much existing code:(
And associated type defaults are unstable yet, so how about changing to
```
pub trait VolatileMemory<B: Bitmap = ()>
```

src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -0,0 +1,235 @@
...
+use crate::mmap::NewBitmap;
+
+/// `AtomicBitmap` implements a simple bit map on the page level with test and set operations.
+/// It is page-size aware, so it converts addresses to page numbers before setting or clearing
+/// the bits.
+#[derive(Debug)]
+pub struct AtomicBitmap {
+    map: Vec<AtomicU64>,
+    size: usize,
+    page_size: usize,
```
jiangliu, line 20: How about using page_size_shift instead of page_size? That helps to reduce the calculation needed.

src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -0,0 +1,235 @@
...
+        AtomicBitmap {
+            map,
+            size: bit_size,
+            page_size,
+        }
+    }
+
+    /// Is bit `n` set? Bits outside the range of the bitmap are always unset.
+    pub fn is_bit_set(&self, n: usize) -> bool {
+        if n <= self.size {
```
jiangliu, line 50: Should this be "n < self.size"?

src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -0,0 +1,235 @@
...
+
+        AtomicBitmap {
+            map,
+            size: bit_size,
+            page_size,
+        }
+    }
+
+    /// Is bit `n` set? Bits outside the range of the bitmap are always unset.
+    pub fn is_bit_set(&self, n: usize) -> bool {
```
jiangliu, line 49: how about rename n as bit_index?

src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -0,0 +1,235 @@
...
+        // below does not underflow.
+        if len == 0 {
+            return;
+        }
+
+        let first_bit = start_addr / self.page_size;
+        // Handle input ranges where `start_addr + len - 1` would otherwise overflow an `usize`
+        // by ignoring pages at invalid addresses.
+        let last_bit = start_addr.saturating_add(len - 1) / self.page_size;
+        for n in first_bit..=last_bit {
```
jiangliu, line 77: How about handling bits in batch of AtomicU64?

src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -0,0 +1,235 @@
...
+        }
+        let bit_size = std::cmp::max(1, num_pages);
+        // Create the map of `AtomicU64`, allowing the bit set operations to be done on a non-mut
+        // `Bitmap`, avoiding the need for a Mutex or other serialization.
+        let map_size = ((bit_size - 1) >> 6) + 1;
+        let map: Vec<AtomicU64> = (0..map_size).map(|_| AtomicU64::new(0)).collect();
+
+        AtomicBitmap {
+            map,
+            size: bit_size,
```
jiangliu, line 43: In case of byte_size is zero, size should be zero instead of one.

src/mmap.rs
```diff
@@ -37,6 +38,16 @@ pub use crate::mmap_windows::MmapRegion;
 #[cfg(windows)]
 pub use std::io::Error as MmapRegionError;
 
+/// A `Bitmap` that can be created starting from an initial size.
+pub trait NewBitmap: Bitmap + Default {
```
jiangliu, line 42: How about moving this code block into bitmap.rs?

alexandruag (commented):
@bonzini so from what I understand, you're proposing to use the decorator approach with objects such as `VolatileSlice` as well. The issue here is that unlike decorators for `GuestMemoryMmap` and region (which are still abstracted by the `GuestMemory` etc. interfaces), `VolatileSlice` is itself part of the `vm-memory` interface, so it has to  support tracking for the underlying memory range. Right now, the status of this PR w.r.t. to high level changes that affect all users is as follows:

- The `VolatileSlice`, `VolatileRef`, and `VolatileArrayRef` objects are essentially the same (after using Gerry's suggestion to leverage the default generic parameters feature). The tracking logic is there, but it's implicitly using the zero-cost trivial implementation and the default generic parameters ensure older type signatures that don't care about the new functionality are still valid.

- The `GuestMemoryRegion` and `VolatileMemory` traits have a new associated type, but it can simply be set to `()` when not interesting for a particular implementation.

- `GuestMemoryMmap`, `GuestRegionMmap`, and `MmapRegion` now have a generic `B: Bitmap` type parameter. The preferred way, from my point of view, to work around this is to use type aliases (for example, something like `type GuestMemoryMmap = vm_memory::GuestMemoryMmap<()>`), which can also be placed project-wide to re-export a definition for an entire code base. The extra type parameter is also transparent in generic usage contexts (i.e. for objects behind a `M: GuestMemory` bound). If we want to reduce the number of visible changes to the absolute minimum, then we can provide upstream. For example, we can rename the generic `GuestMemoryMmap<B>` to something like `GuestMemoryMmapBase<B>` and export a `type GuestMemoryMmap = GuestMemoryMmapBase<()>` alias together with the base type itself directly from `vm_memory` (same for `GuestRegionMmap` and `MmapRegion`). This looks more cumbersome overall to me, but it's def an option to be considered.

@rbradford @jiangliu @likebreath and other folks, how do you see things moving forward here?

jiangliu:
> @bonzini so from what I understand, you're proposing to use the decorator approach with objects such as `VolatileSlice` as well. The issue here is that unlike decorators for `GuestMemoryMmap` and region (which are still abstracted by the `GuestMemory` etc. interfaces), `VolatileSlice` is itself part of the `vm-memory` interface, so it has to support tracking for the underlying memory range. Right now, the status of this PR w.r.t. to high level changes that affect all users is as follows:
> 
> * The `VolatileSlice`, `VolatileRef`, and `VolatileArrayRef` objects are essentially the same (after using Gerry's suggestion to leverage the default generic parameters feature). The tracking logic is there, but it's implicitly using the zero-cost trivial implementation and the default generic parameters ensure older type signatures that don't care about the new functionality are still valid.
> * The `GuestMemoryRegion` and `VolatileMemory` traits have a new associated type, but it can simply be set to `()` when not interesting for a particular implementation.
> * `GuestMemoryMmap`, `GuestRegionMmap`, and `MmapRegion` now have a generic `B: Bitmap` type parameter. The preferred way, from my point of view, to work around this is to use type aliases (for example, something like `type GuestMemoryMmap = vm_memory::GuestMemoryMmap<()>`), which can also be placed project-wide to re-export a definition for an entire code base. The extra type parameter is also transparent in generic usage contexts (i.e. for objects behind a `M: GuestMemory` bound). If we want to reduce the number of visible changes to the absolute minimum, then we can provide upstream. For example, we can rename the generic `GuestMemoryMmap<B>` to something like `GuestMemoryMmapBase<B>` and export a `type GuestMemoryMmap = GuestMemoryMmapBase<()>` alias together with the base type itself directly from `vm_memory` (same for `GuestRegionMmap` and `MmapRegion`). This looks more cumbersome overall to me, but it's def an option to be considered.

Things like "GuestMemoryMmapBase" may be a burden for maintenance. 
I think current solution is OK. With the default generic parameter applied, it should need too much effort to migrate existing code.

alexandruag:
You're right; I experimented a bit more and looks like the default generic parameter works automatically in most cases. The only places where something needs to be explicitly specified are free standing statements (i.e. like the examples in the doc comments). Whenever there's context that already specifies (for example) the `GuestMemoryMmap` type as part of a struct field definition, return value or parameter type, etc (which happens most of the time) the default generic parameter works without additional changes. Thanks again for the idea Gerry!

Right now, the two suggestions about improving the `AtomicBitmap` implementation are still pending, as they need a bit more work on my side. Since they don't significantly impact the rest of the code, are you folks ok with merging this PR and then adding one commit for those changes as part of the PR for publishing a new version that will follow? @jiangliu @likebreath @rbradford

jiangliu:
> You're right; I experimented a bit more and looks like the default generic parameter works automatically in most cases. The only places where something needs to be explicitly specified are free standing statements (i.e. like the examples in the doc comments). Whenever there's context that already specifies (for example) the `GuestMemoryMmap` type as part of a struct field definition, return value or parameter type, etc (which happens most of the time) the default generic parameter works without additional changes. Thanks again for the idea Gerry!
> 
> Right now, the two suggestions about improving the `AtomicBitmap` implementation are still pending, as they need a bit more work on my side. Since they don't significantly impact the rest of the code, are you folks ok with merging this PR and then adding one commit for those changes as part of the PR for publishing a new version that will follow? @jiangliu @likebreath @rbradford

I'm OK with that because it's just an optimization:)

jiangliu (approved):

likebreath (approved):

src/guest_memory.rs
```diff
@@ -836,22 +849,29 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
...
+                let result = loop {
                     match src.read(&mut dst[start..end]) {
                         Ok(n) => break Ok(n),
                         Err(ref e) if e.kind() == std::io::ErrorKind::Interrupted => continue,
                         Err(e) => break Err(Error::IOError(e)),
                     }
-                }
+                };
+
+                region.bitmap().mark_dirty(start, len);
```
lauralt, line 864: Can we use `result` instead of `len`?

lauralt (approved):

PR #126, opened 2020-12-11
volatile-memory: add split_at()

elmarco:
This will come handy to implement https://github.com/rust-vmm/vm-virtio/pull/33

### DIFF
--- a/src/volatile_memory.rs
+++ b/src/volatile_memory.rs
@@ -290,6 +290,26 @@ impl<'a> VolatileSlice<'a> {
         self.size == 0
     }
 
+    /// Divides one slice into two at an index.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use vm_memory::VolatileMemory;
+    /// let mut mem = [0u8; 32];
+    /// let mem_ref = &mut mem[..];
+    /// let vslice = mem_ref.get_slice(0, 32).unwrap();
+    /// let (start, end) = vslice.split_at(8).unwrap();
+    /// ```
+    pub fn split_at(self, mid: usize) -> Result<(VolatileSlice<'a>, VolatileSlice<'a>)> {
+        let end = self.offset(mid)?;
+        let start = unsafe {
+            // safe because self.offset() already checked the bounds
+            VolatileSlice::new(self.addr, mid)
+        };
+        Ok((start, end))
+    }
+
     /// Returns a subslice of this [`VolatileSlice`](struct.VolatileSlice.html) starting at
     /// `offset`.
     ///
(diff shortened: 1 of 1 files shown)


elmarco:
buildkite failure isn't  fair:

pytest rust-vmm-ci/integration_tests/test_coverage.py
FAILED rust-vmm-ci/integration_tests/test_coverage.py::test_coverage - AssertionError: Coverage is increased from 85.5 to 85.6. Please update the coverage in tests/coverage.

andreeaflorescu:
@elmarco yeah, that's not nice indeed. It isn't buidlkite's fault though, it is ours because the coverage test is implemented such that if the coverage is not the same as the one configured in [coverage_config_x86_64.json](https://github.com/rust-vmm/vm-memory/blob/master/coverage_config_x86_64.json), the test fails. You'll need to manually update the coverage score so that the test passes. It is a bit bizarre though that the coverage is increased when the code you added is not tested.

elmarco:
@andreeaflorescu the added code is covered by the example test

src/volatile_memory.rs
```diff
@@ -290,6 +290,28 @@ impl<'a> VolatileSlice<'a> {
...
+    /// ```
+    /// # use vm_memory::VolatileMemory;
+    /// let mut mem = [0u8; 32];
+    /// let mem_ref = &mut mem[..];
+    /// let vslice = mem_ref.get_slice(0, 32).unwrap();
+    /// let (start, end) = vslice.split_at(8).unwrap();
+    /// assert_eq!(start.len(), 8);
+    /// assert_eq!(end.len(), 24);
+    /// ```
+    pub fn split_at(self, mid: usize) -> Result<(VolatileSlice<'a>, VolatileSlice<'a>)> {
```
jiangliu, line 306: could you please also help to add test cases for:
1) mid = 0,
2) mid = self.len() -1
3) mid = self.len()

elmarco: added

jiangliu: sorry，missed a case for "mid = self.len() + 1" then we could test the error handling code path:)

jiangliu:
> This will come handy to implement [rust-vmm/vm-virtio#33](https://github.com/rust-vmm/vm-virtio/pull/33)

when does it need to split a slice buffer into reader and writer parter?
According to virtio spec, a descriptor will be readonly or write-only, what's the case where a buffer will be rw?

src/volatile_memory.rs
```diff
@@ -290,6 +290,37 @@ impl<'a> VolatileSlice<'a> {
...
+    /// assert_eq!(end.len(), 24);
+    /// let (start, end) = vslice.split_at(0).unwrap();
+    /// # assert_eq!(start.len(), 0);
+    /// # assert_eq!(end.len(), 32);
+    /// # let (start, end) = vslice.split_at(31).unwrap();
+    /// # assert_eq!(start.len(), 31);
+    /// # assert_eq!(end.len(), 1);
+    /// # let (start, end) = vslice.split_at(32).unwrap();
+    /// # assert_eq!(start.len(), 32);
+    /// # assert_eq!(end.len(), 0);
```
andreeaflorescu, line 313: Do you mind moving the checks to the unit tests and keeping just a meaningful example here?

It looks a bit cluttered when opening the documentation with `cargo doc --open`, and it's not really IDE compatible (you cannot run them via something like Intelij easily).

elmarco:
> when does it need to split a slice buffer into reader and writer parter?
> According to virtio spec, a descriptor will be readonly or write-only, what's the case where a buffer will be rw?

This is not the concern here, but rather of DescriptorChainConsumer in rust-vmm/vm-virtio#33. It collects the readable() buffers for the Reader, and writable for the Writer. There is thus no buffer that can be dual read & write.

src/volatile_memory.rs
elmarco: sure, although the # lines shouldn't be visible in the generated doc by default.

bonzini:
Looks good once the coverage % is fixed.

jiangliu (approved):

bonzini (approved):

PR #130, opened 2020-12-29
Add GuestMemory method to return an Iterator

### LINKED ISSUE #8: Add GuestMemory method to return an Iterator

Daniel-Aaron-Bloom:
Fixes #8.

This changes is similar to #9, which was rejected as it introduced a lifetime parameter to the trait, in favor of waiting for GAT/GAL. Since GAT/GAL still looks a ways off, this change does the same thing but using [HRTBs](https://lukaskalbertodt.github.io/2018/08/03/solving-the-generalized-streaming-iterator-problem-without-gats.html#workaround-b-hrtbs--the-family-trait-pattern).

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -394,6 +394,25 @@ impl<M: GuestMemory> GuestAddressSpace for Arc<M> {
     }
 }
 
+/// Lifetime generic associated iterators. The actual iterator type is defined through associated
+/// item `Iter`, for example:
+///
+/// ```ignore
+/// impl<'a> GuestMemoryIterator<'a, MyGuestRegion> for MyGuestMemory {
+///     type Iter = MyGuestMemoryIter<'a>;
+/// }
+///
+/// pub struct MyGuestMemoryIter<`a>( /* ... */ );
+/// impl<'a> Iterator for MyGuestMemoryIter<'a> {
+///     type Item = &'a MyGuestRegion;
+///     /* ... */
+/// }
+/// ```
+pub trait GuestMemoryIterator<'a, R: 'a> {
+    /// Type of the `iter` method's return value.
+    type Iter: Iterator<Item = &'a R>;
+}
+
 /// `GuestMemory` represents a container for an *immutable* collection of
 /// `GuestMemoryRegion` objects.  `GuestMemory` provides the `Bytes<GuestAddress>`
 /// trait to hide the details of accessing guest memory by physical address.
@@ -407,6 +426,9 @@ pub trait GuestMemory {
     /// Type of objects hosted by the address space.
     type R: GuestMemoryRegion;
 
+    /// Lifetime generic associated iterators. Usually this is just `Self`.
+    type I: for<'a> GuestMemoryIterator<'a, Self::R>;
+
     /// Returns the number of regions in the collection.
     fn num_regions(&self) -> usize;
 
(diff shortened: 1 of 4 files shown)


Daniel-Aaron-Bloom:
Should I make a separate PR for all the clippy fixes and miscellanea?

bonzini:
It's okay to keep them in this pull request, but please move those that apply to the master branch in a separate commit; the others should be squashed in the commit that introduces the problem in the first place.

Also, to complete my review:

* I would leave the map_and_fold and with_regions* methods in order to avoid breaking API compatibility

* Does the `pub struct Iter<'a>` for Guest memory map need to be public at all? If not, I would rather hide it.

* the GuestMemoryIter trait needs a more complete example in the documentation.

Daniel-Aaron-Bloom:
>I would leave the map_and_fold and with_regions* methods in order to avoid breaking API compatibility

Sounds good to me. I'll make this change.

>the GuestMemoryIter trait needs a more complete example in the documentation.

There is no runtime usage for `GuestMemoryIterator`. It is only a dummy trait which enables lifetime HKTs in order to setup the returned iterator type with the proper lifetime. I'll try to improve the documentation, but more guidance on this point would be appreciated.

>Does the `pub struct Iter<'a>` for Guest memory map need to be public at all? If not, I would rather hide it.

It does need to be public as users may need to store and wrap iterators.

> It's okay to keep them in this pull request, but please move those that apply to the master branch in a separate commit; the others should be squashed in the commit that introduces the problem in the first place.

Since most (almost all) of these issues existed before, I've separated out the fixes for pre-existing clippy issues into #131. That should probably go through first and clear up the CI, then I'll rebase this and cleanup the commit history (along with the other requested changes)

jiangliu:
LGTM:)
Could you please also help to handle GuestMemoryAtomic in atomic.rs? Otherwise it breaks.

Daniel-Aaron-Bloom:
> Could you please also help to handle GuestMemoryAtomic in atomic.rs? Otherwise it breaks.

I'm confused. `GuestMemoryAtomic` doesn't implement the `GuestMemory` interface. The associated guards implement `Deref`, but that should work cleanly to access the new methods.

jiangliu:
> > Could you please also help to handle GuestMemoryAtomic in atomic.rs? Otherwise it breaks.
> 
> I'm confused. `GuestMemoryAtomic` doesn't implement the `GuestMemory` interface. The associated guards implement `Deref`, but that should work cleanly to access the new methods.

You are right:)

bonzini:
Regarding documentation, I would like something like

```rust

    /// Lifetime generic associated iterator. Usually this is just `Self`.  The actual
    /// iterator type is specified through the `GuestMemoryIterator` trait, for example:
    ///
    /// ```ignore
    /// impl<'a> GuestMemoryIterator<'a, MyGuestRegion> for MyGuestMemory {
    ///     type Iter = MyGuestMemoryIter<'a>;
    /// }
    ///
    /// pub struct MyGuestMemoryIter<`a>( ... );
    /// impl<'a> Iterator for MyGuestMemoryIter<'a> { ... }
    /// ```
    type Iterator: for<'a> GuestMemoryIterator<'a, Self::R>;
```

that explains how to use the idiom.

Finally, since it is public please rename the `Iter` type to `GuestMemoryMmapIter`. Thanks!

src/guest_memory.rs
```diff
@@ -407,6 +413,9 @@ pub trait GuestMemory {
     /// Type of objects hosted by the address space.
     type R: GuestMemoryRegion;
 
+    /// Lifetime generic associated iterator.
+    type Iterator: for<'a> GuestMemoryIterator<'a, Self::R>;
```
alexandruag, line 417: Can we rename the associated type to `I` so it's more consistent with the already existing `R`? Calling it `Iterator` is also a bit confusing, since it's not the actual type that's used as the iterator.

alexandruag (commented):
This is neat! I think we can avoid explicitly renaming the `mmap::Iter` type to `GuestMemoryMmapIter`, since the module path already provides sufficient context (this also happens for some of the iterators in the standard library, i.e. `slice::Iter`).

It's definitely a good idea to keep the old methods around some more to avoid immediately breaking compatibility, but I think we should remove them at some point in the future. What do you folks think about marking them as [deprecated](https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/an-attribute-for-deprecation.html) for now, and removing them in a later release?

Finally, we should also add some changelog entries about the publicly visible changes (in an `[Unreleased]` section right before `[0.4.0]`, that's going to be renamed using the proper version number when we prepare the next release).

Daniel-Aaron-Bloom:
I'm also considering adding a `fn find_regions(&self, start: GuestAddress, end: GuestAddress) -> <Self::I as GuestMemoryIterator<Self::R>>::Regions` to support rust-vmm/vm-virtio#33. Specifically in regards to @alexandruag comment on

>getting the minimal sequence of VolatileSlices that makes up an otherwise contiguous (and valid) GPA range

Am I correct that a `find_regions` would help resolve that issue?

jiangliu (approved):

src/guest_memory.rs
```diff
@@ -394,6 +394,25 @@ impl<M: GuestMemory> GuestAddressSpace for Arc<M> {
     }
 }
 
+/// Lifetime generic associated iterator. Usually this is just `Self`.  The actual
```
bonzini, line 397: This is the description for `type I`, not for `GuestMemoryIterator`.

bonzini (changes requested):
Just one remaining comment for the documentation.

src/guest_memory.rs
Daniel-Aaron-Bloom: Whoops.

bonzini (approved):

jiangliu (approved):

alexandruag:
> I'm also considering adding a `fn find_regions(&self, start: GuestAddress, end: GuestAddress) -> <Self::I as GuestMemoryIterator<Self::R>>::Regions` to support [rust-vmm/vm-virtio#33](https://github.com/rust-vmm/vm-virtio/pull/33). Specifically in regards to @alexandruag comment on
> 
> > getting the minimal sequence of VolatileSlices that makes up an otherwise contiguous (and valid) GPA range
> 
> Am I correct that a `find_regions` would help resolve that issue?

Apologies for missing this comment earlier. In my mind, the question has increased in scope recently, for example also touching on the subject of how we can leverage something along these lines to simplify how cross-region accesses are performed. I'll open an issue as soon as some sort of idea does eventually crystallize, and hopefully we can reduce overall complexity some more. I'm also more than happy to discuss other ideas as well.

PR #134, opened 2021-01-10
Fixed-up examples

Daniel-Aaron-Bloom:
Fixed up all the examples. Changes include:

- Un-hiding `use`s necessary for visible code
- Removing indentation to baseline
- Removing unused `use`s
- Removing unnecessary functions
- Fixing broken examples

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -95,9 +95,10 @@ pub trait Address:
     ///
     /// ```
     /// # use vm_memory::{Address, GuestAddress};
-    ///   let base = GuestAddress(0x100);
-    ///   let addr = GuestAddress(0x150);
-    ///   assert_eq!(addr.unchecked_offset_from(base), 0x50);
+    /// #
+    /// let base = GuestAddress(0x100);
+    /// let addr = GuestAddress(0x150);
+    /// assert_eq!(addr.unchecked_offset_from(base), 0x50);
     /// ```
     fn unchecked_offset_from(&self, base: Self) -> Self::V {
         self.raw_value() - base.raw_value()
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -263,27 +263,28 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
 
     /// Gets a slice of memory for the entire region that supports volatile access.
     ///
-    /// # Examples
+    /// # Examples (uses the `backend-mmap` feature)
     ///
     /// ```
     /// # #[cfg(feature = "backend-mmap")]
+    /// # {
     /// # use vm_memory::{GuestAddress, MmapRegion, GuestRegionMmap, GuestMemoryRegion};
-    /// # #[cfg(feature = "backend-mmap")]
     /// # use vm_memory::volatile_memory::{VolatileMemory, VolatileSlice, VolatileRef};
+    /// #
+    /// let region = MmapRegion::new(0x400).expect("Could not create mmap region");
+    /// let region =
+    ///     GuestRegionMmap::new(region, GuestAddress(0x0)).expect("Could not create guest memory");
+    /// let slice = region
+    ///     .as_volatile_slice()
+    ///     .expect("Could not get volatile slice");
     ///
-    /// # #[cfg(feature = "backend-mmap")]
-    /// # fn test_as_volatile_slice() {
-    ///     let region =
-    ///         GuestRegionMmap::new(MmapRegion::new(0x400).unwrap(), GuestAddress(0x0))
-    ///         .unwrap();
-    ///     let slice = region.as_volatile_slice().unwrap();
-    ///     let v = 42u32;
-    ///     let r = slice.get_ref::<u32>(0x200).unwrap();
-    ///     r.store(v);
-    ///     assert_eq!(r.load(), v);
+    /// let v = 42u32;
+    /// let r = slice
+    ///     .get_ref::<u32>(0x200)
+    ///     .expect("Could not get reference");
+    /// r.store(v);
+    /// assert_eq!(r.load(), v);
     /// # }
-    /// # #[cfg(feature = "backend-mmap")]
-    /// # test_as_volatile_slice();
     /// ```
     fn as_volatile_slice(&self) -> Result<volatile_memory::VolatileSlice> {
         self.get_slice(MemoryRegionAddress(0), self.len() as usize)
@@ -302,14 +303,14 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
 /// implementation, providing a zero-cost abstraction whenever immutable memory
 /// maps are sufficient.
 ///
-/// ```
-/// # use std::sync::Arc;
-/// # #[cfg(feature = "backend-mmap")]
-/// # use vm_memory::GuestMemoryMmap;
-/// # #[cfg(feature = "backend-atomic")]
-/// # use vm_memory::GuestMemoryAtomic;
-/// # use vm_memory::{GuestAddress, GuestMemory, GuestAddressSpace};
+/// # Examples (uses the `backend-mmap` and `backend-atomic` features)
 ///
+/// ```
+/// # #[cfg(feature = "backend-mmap")]
+/// # {
+/// # use std::sync::Arc;
+/// # use vm_memory::{GuestAddress, GuestAddressSpace, GuestMemory, GuestMemoryMmap};
+/// #
 /// pub struct VirtioDevice<AS: GuestAddressSpace> {
 ///     mem: Option<AS>,
 /// }
@@ -323,37 +324,33 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
 ///     }
 /// }
 ///
-/// # #[cfg(feature = "backend-mmap")]
-/// # fn get_mmap() -> GuestMemoryMmap {
-/// #     GuestMemoryMmap::from_ranges(&[(GuestAddress(0),0)]).unwrap()
-/// # }
+/// fn get_mmap() -> GuestMemoryMmap {
+///     let start_addr = GuestAddress(0x1000);
+///     GuestMemoryMmap::from_ranges(&vec![(start_addr, 0x400)])
+///         .expect("Could not create guest memory")
+/// }
 ///
-/// # #[cfg(feature = "backend-mmap")]
-/// # fn test_1() {
 /// // Using `VirtioDevice` with an immutable GuestMemoryMmap:
-/// let mut for_immutable_mmap: VirtioDevice<&GuestMemoryMmap> =
-///     VirtioDevice::new();
+/// let mut for_immutable_mmap = VirtioDevice::<&GuestMemoryMmap>::new();
 /// let mmap = get_mmap();
 /// for_immutable_mmap.activate(&mmap);
-/// let mut another: VirtioDevice<&GuestMemoryMmap> =
-///     VirtioDevice::new();
+/// let mut another = VirtioDevice::<&GuestMemoryMmap>::new();
 /// another.activate(&mmap);
-/// # }
 ///
-/// # #[cfg(all(feature = "backend-mmap", feature = "backend-atomic"))]
-/// # fn test_2() {
+/// # #[cfg(feature = "backend-atomic")]
+/// # {
+/// # use vm_memory::GuestMemoryAtomic;
 /// // Using `VirtioDevice` with a mutable GuestMemoryMmap:
-/// let mut for_mutable_mmap: VirtioDevice<GuestMemoryAtomic<GuestMemoryMmap>> =
-///     VirtioDevice::new();
+/// let mut for_mutable_mmap = VirtioDevice::<GuestMemoryAtomic<GuestMemoryMmap>>::new();
 /// let atomic = GuestMemoryAtomic::new(get_mmap());
 /// for_mutable_mmap.activate(atomic.clone());
-/// let mut another: VirtioDevice<GuestMemoryAtomic<GuestMemoryMmap>> =
-///     VirtioDevice::new();
+/// let mut another = VirtioDevice::<GuestMemoryAtomic<GuestMemoryMmap>>::new();
 /// another.activate(atomic.clone());
+///
 /// // atomic can be modified here...
 /// # }
+/// # }
 /// ```
-
 pub trait GuestAddressSpace {
     /// The type that will be used to access guest memory.
     type M: GuestMemory;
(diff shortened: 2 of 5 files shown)


src/mmap.rs
```diff
@@ -288,13 +312,18 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
...
-    ///   gm.write_to(start_addr, &mut file, 128).unwrap();
+    /// use vm_memory::{Address, Bytes, GuestAddress, GuestMemoryMmap};
+    ///
+    /// let start_addr = GuestAddress(0x1000);
+    /// let gm = GuestMemoryMmap::from_ranges(&vec![(start_addr, 0x400)])
+    ///     .expect("Could not create guest memory");
+    /// # let mut file = TempFile::new().unwrap().into_file();
+    /// # /*
+    /// let mut file = /* temp file */;
+    /// # */
```
lauralt, line 323: This is interesting, but somehow makes reading the code example unnecessary hard. Do you have a strong preference for a generated example that doesn't use `TempFile`?

src/address.rs
```diff
@@ -94,10 +94,12 @@ pub trait Address:
...
-    /// # use vm_memory::{Address, GuestAddress};
-    ///   let base = GuestAddress(0x100);
-    ///   let addr = GuestAddress(0x150);
-    ///   assert_eq!(addr.unchecked_offset_from(base), 0x50);
+    /// use vm_memory::{Address, GuestAddress};
+    ///
+    /// let base = GuestAddress(0x100);
+    /// let addr = GuestAddress(0x150);
+    ///
+    /// assert_eq!(addr.unchecked_offset_from(base), 0x50);
```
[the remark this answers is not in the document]
Daniel-Aaron-Bloom: I was using the standard library as my source of truth for example style, but it seems like you're right, most crates do it the way you have suggested. I will undo the relevant changes.

andreeaflorescu (commented):
I ran a `cargo fmt` on the code from examples, and there are a few errors. It is hard to go over them and see which ones are introduced by this PR though.

If you'd like to also run `cargo fmt`, you can do it by installing rust nightly, and then add a `rustfmt.toml` in the root directory of the project. In the rustfmt toml file, you need to add:
```toml
format_code_in_doc_comments = true
```

This is a nice to have.

src/guest_memory.rs
```diff
@@ -267,23 +267,24 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
...
-    /// # fn test_as_volatile_slice() {
-    ///     let region =
-    ///         GuestRegionMmap::new(MmapRegion::new(0x400).unwrap(), GuestAddress(0x0))
-    ///         .unwrap();
-    ///     let slice = region.as_volatile_slice().unwrap();
-    ///     let v = 42u32;
-    ///     let r = slice.get_ref::<u32>(0x200).unwrap();
-    ///     r.store(v);
-    ///     assert_eq!(r.load(), v);
+    /// let region = MmapRegion::new(0x400)
```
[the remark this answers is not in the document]
andreeaflorescu: I am blind! @lauralt explained that all the code now runs only for `feature = "backend-mmap"`. Nice!

I am still thinking that maybe we should suggest somehow that this example is based on `GuestMemoryMmap` which is only available with that feature. Could we maybe add this information to the heading? Something like:
``` # Example using the `backend-mmap` feature```

This problem is not directly related to the PR, so feel free to consider this comment a nit.

File: src/mmap.rs (hunk 1 above)
Comment (Daniel-Aaron-Bloom) [reply src/mmap.rs:323]: `TempFile` seemed like just a random test helper class, referencing it seemed more confusing than helpful.

File: src/guest_memory.rs (hunk 3 above)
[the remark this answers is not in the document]
Comment (Daniel-Aaron-Bloom) [reply src/guest_memory.rs]: I mean, this whole PR is basically a nit so... 😝 

Let me know if what I wrote works for you or if I missed any locations.

File: src/mmap.rs (hunk 1 above)
Comment (Daniel-Aaron-Bloom) [reply src/mmap.rs]: I've aligned this and with the other `write_to` examples in `guest_memory.rs`. Now both use `/dev/null` in the visible example and use `TempFile` as a fallback for non `cfg!(unix)` environments.

Thoughts?

File: src/address.rs (hunk 2 above)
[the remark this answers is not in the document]
Comment (Daniel-Aaron-Bloom) [reply src/address.rs]: All uses should be hidden and gaps should be minimized. As with everything else, let me know if I missed any.

File: src/guest_memory.rs [hunk 4]
```diff
@@ -920,6 +917,36 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
         })
     }
 
+    /// # Examples
```
andreeaflorescu, line 920: nit: Missing comment about mmap backend.

src/guest_memory.rs
```diff
@@ -469,24 +491,21 @@ pub trait GuestMemory {
     ///   and dividing their sizes to 1024, then summing up the values in an accumulator.
     ///
     /// ```
-    /// # #[cfg(feature = "backend-mmap")]
+    /// ##[cfg(feature = "backend-mmap")]
```
andreeaflorescu, line 494: nit: missing the "(uses the `backend-mmap` feature)"

src/guest_memory.rs
```diff
@@ -507,26 +526,18 @@ pub trait GuestMemory {
...
```
andreeaflorescu, line 529: nit: missing the "(uses the `backend-mmap` feature)"

src/guest_memory.rs
```diff
@@ -542,19 +553,16 @@ pub trait GuestMemory {
     /// # Examples
     ///
     /// ```
-    /// # #[cfg(feature = "backend-mmap")]
+    /// ##[cfg(feature = "backend-mmap")]
```
andreeaflorescu, line 556: nit: missing the "(uses the `backend-mmap` feature)"

src/guest_memory.rs
```diff
@@ -782,30 +781,26 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
     /// * Read bytes from /dev/urandom
     ///
     /// ```
-    /// # #[cfg(feature = "backend-mmap")]
-    /// # use vm_memory::{Address, Bytes, GuestAddress, mmap::GuestMemoryMmap};
+    /// ##[cfg(feature = "backend-mmap")]
```
andreeaflorescu, line 784: nit: missing the "(uses the `backend-mmap` feature)"

src/guest_memory.rs
```diff
@@ -749,22 +752,18 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
     /// * Read a slice of length 16 at guestaddress 0x1000.
     ///
     /// ```
-    /// # #[cfg(feature = "backend-mmap")]
+    /// ##[cfg(feature = "backend-mmap")]
```
andreeaflorescu, line 755: nit: missing the "(uses the `backend-mmap` feature)"

src/guest_memory.rs
```diff
@@ -718,21 +725,17 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
     /// * Write a slice at guestaddress 0x1000.
     ///
     /// ```
-    /// # #[cfg(feature = "backend-mmap")]
+    /// ##[cfg(feature = "backend-mmap")]
```
andreeaflorescu, line 728: nit: missing the "(uses the `backend-mmap` feature)"

andreeaflorescu (approved):

src/guest_memory.rs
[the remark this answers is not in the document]
andreeaflorescu: There were a few in guest_memory.rs, but maybe it was intentional. I added comments with them, but they're optional. Thanks for also fixing the formatting of the doc examples :+1:

src/volatile_memory.rs
```diff
@@ -296,10 +292,18 @@ impl<'a> VolatileSlice<'a> {
...
-    /// let vslice = mem_ref.get_slice(0, 32).unwrap();
-    /// let (start, end) = vslice.split_at(8).unwrap();
+    /// #
+    /// # // Create a buffer
+    /// # let mut mem = [0u8; 32];
+    /// # let mem_ref = &mut mem[..];
+    /// #
+    /// # // Get a `VolatileSlice` from the buffer
+    /// # let vslice = mem_ref.get_slice(0, 32)
+    /// #    .expect("Could not get VolatileSlice");
```
lauralt, line 302: We should keep these lines un-hidden as well so the asserts make sense in the generated documentation.

src/volatile_memory.rs
```diff
@@ -362,20 +366,23 @@ impl<'a> VolatileSlice<'a> {
...
-    /// let mem_ref = &mut mem[..];
-    /// let vslice = mem_ref.get_slice(0, 32).map_err(|_| ())?;
+    /// #
+    /// # // Create a buffer
+    /// # let mut mem = [0u8; 32];
+    /// # let mem_ref = &mut mem[..];
+    /// #
+    /// # // Get a `VolatileSlice` from the buffer
+    /// # let vslice = mem_ref.get_slice(0, 32)
+    /// #    .expect("Could not get VolatileSlice");
```
lauralt, line 377: Same suggestion as above.

src/volatile_memory.rs
```diff
@@ -505,13 +525,19 @@ impl Bytes<usize> for VolatileSlice<'_> {
...
+    /// # let mut mem = [0u8; 1024];
+    /// # let mut mem_ref = &mut mem[..];
+    /// #
+    /// # // Get a `VolatileSlice` from the buffer
+    /// # let vslice = mem_ref.as_volatile_slice();
+    /// #
+    /// let res = vslice.write(&[1, 2, 3, 4, 5], 1020);
+    ///
+    /// assert!(res.is_ok());
+    /// assert_eq!(res.unwrap(), 4);
```
lauralt, line 540: Looks like the same thing happens here as well :( Feel free to ignore this nit since it's not a change from this PR.

src/guest_memory.rs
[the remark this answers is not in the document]
Daniel-Aaron-Bloom: Nope. Completely unintentional. Fixing these up.

src/guest_memory.rs
```diff
@@ -263,27 +263,28 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
...
-    ///     let v = 42u32;
-    ///     let r = slice.get_ref::<u32>(0x200).unwrap();
-    ///     r.store(v);
-    ///     assert_eq!(r.load(), v);
+    /// #
+    /// # let region = MmapRegion::new(0x400)
+    /// #    .expect("Could not create mmap region");
+    /// # let region = GuestRegionMmap::new(region, GuestAddress(0x0))
+    /// #    .expect("Could not create guest memory");
+    /// # let slice = region.as_volatile_slice()
```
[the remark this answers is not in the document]
Daniel-Aaron-Bloom: Whoops, overzealous copy-pasting. Sorry about that.

src/guest_memory.rs
```diff
@@ -302,14 +303,14 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
 /// implementation, providing a zero-cost abstraction whenever immutable memory
 /// maps are sufficient.
 ///
+/// # Examples (uses the `backend-mmap` and `backend-atomic` features)
+///
 /// ```
+/// ##[cfg(feature = "backend-mmap")]
```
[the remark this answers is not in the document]
Daniel-Aaron-Bloom: Ahh, this is probably because I have to remove all the `#[cfg(...)]`s to get fmt to actually work on the doc comments. Must have made some error when I unremoved them.

src/guest_memory.rs
```diff
@@ -859,31 +854,33 @@ impl<T: GuestMemory> Bytes<GuestAddress> for T {
...
-    ///     let mut file = OpenOptions::new()
-    ///         .write(true)
-    ///         .open("/dev/null")
-    ///         .expect("Could not open /dev/null");
+    /// let mut file = OpenOptions::new()
+    ///     .write(true)
+    ///     .open("/dev/null")
+    ///     .expect("Could not open /dev/null");
+    /// #   file
+    /// # } else {
```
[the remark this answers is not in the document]
Daniel-Aaron-Bloom: The example doesn't look any different on windows, it just runs different. The way I look at this example now is just that since Windows doesn't support `/dev/null`, `TempFile` is a rough equivalent that enables some runtime test coverage of this example on Windows.

The thing I have against `TempFile` is that it's not a structure I would expect users to be familiar with or care about, since it's only really used internally in vmm tests. Using it in as a visible part of an example distracts from the actual purpose of the example. I personally got confused enough by this example at first glance to waste time looking up first where `TempFile` was defined, and second what it actually did.

I don't think either is necessary, since almost no examples I've seen show both the unix and windows cases, but I'd be more amenable to something like a call to a hidden function (e.g. `get_temp_file`).

src/volatile_memory.rs
Daniel-Aaron-Bloom: Wait, what's happening here? Too many hidden lines?

src/mmap.rs
```diff
@@ -212,20 +222,26 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
...
     /// # use vm_memory::{Address, Bytes, GuestAddress, GuestMemoryMmap};
     /// # use std::fs::File;
     /// # use std::path::Path;
+    /// #
     /// # let start_addr = GuestAddress(0x1000);
-    /// # let gm = GuestMemoryMmap::from_ranges(&vec![(start_addr, 0x400)]).unwrap();
-    ///   let mut file = if cfg!(unix) {
-    ///       File::open(Path::new("/dev/urandom")).unwrap()
-    ///   } else {
-    ///       File::open(Path::new("c:\\Windows\\system32\\ntoskrnl.exe")).unwrap()
```
lauralt, line 223: Just curious, should we leave the `windows` branch as well? (sorry, I only noticed this change now)

src/volatile_memory.rs
lauralt: Yep, but it seems fixed now, thanks!

lauralt (commented):
Looks good!

src/mmap.rs
Daniel-Aaron-Bloom: Indeed we should, hiding it like the `TempFile` setup. Also copying this over to the other `/dev/urandom` examples.

andreeaflorescu (approved):

alexandruag (approved):

PR #144, opened 2021-03-05
Deprecate `GuestMemory` stopgap "iterator" methods

### LINKED ISSUE #133: Deprecate `GuestMemory` stopgap "iterator" methods
`with_regions`, `with_regions_mut` and `map_and_fold` were just papering over the lack of iterator-based access to the regions in `GuestMemory`. After #130 closes out #8, they will no longer be needed and should be [deprecated](https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/an-attribute-for-deprecation.html) and eventually removed.

defunctio:
* Deprecates `with_regions` `with_regions_mut` and `map_and_fold`.
* Update unit tests

Resolves #133

### DIFF
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -1196,16 +1196,14 @@ mod tests {
         ];
         let mut iterated_regions = Vec::new();
         let gm = GuestMemoryMmap::from_ranges(&regions).unwrap();
-        let res: guest_memory::Result<()> = gm.with_regions(|_, region| {
+
+        for region in gm.iter() {
             assert_eq!(region.len(), region_size as GuestUsize);
-            Ok(())
-        });
-        assert!(res.is_ok());
-        let res: guest_memory::Result<()> = gm.with_regions_mut(|_, region| {
+        }
+
+        for region in gm.iter() {
             iterated_regions.push((region.start_addr(), region.len() as usize));
-            Ok(())
-        });
-        assert!(res.is_ok());
+        }
         assert_eq!(regions, iterated_regions);
 
         assert!(regions
@@ -1228,16 +1226,13 @@ mod tests {
         let gm = Arc::new(GuestMemoryMmap::from_ranges(&regions).unwrap());
         let mem = gm.memory();
 
-        let res: guest_memory::Result<()> = mem.with_regions(|_, region| {
+        for region in mem.iter() {
             assert_eq!(region.len(), region_size as GuestUsize);
-            Ok(())
-        });
-        assert!(res.is_ok());
-        let res: guest_memory::Result<()> = mem.with_regions_mut(|_, region| {
+        }
+
+        for region in mem.iter() {
             iterated_regions.push((region.start_addr(), region.len() as usize));
-            Ok(())
-        });
-        assert!(res.is_ok());
+        }
         assert_eq!(regions, iterated_regions);
 
         assert!(regions
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
 # Changelog
+## [Unreleased]
+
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "vm-memory"
-version = "0.5.0"
+version = "0.5.1"
```
jiangliu, line 3: it's a break change, better to bump to "0.6"

jiangliu: My bad, it's deprecating instead of removing. "0.5.1" should be ok:)

src/mmap.rs
```diff
@@ -1196,16 +1196,14 @@ mod tests {
         ];
         let mut iterated_regions = Vec::new();
         let gm = GuestMemoryMmap::from_ranges(&regions).unwrap();
-        let res: guest_memory::Result<()> = gm.with_regions(|_, region| {
```
andreeaflorescu, line 1199: Should we update the unit tests when we actually remove the functions? Even if they're deprecated, they can still be used.

Cargo.toml
andreeaflorescu: We typically update the version when doing the release. Do we plan to do a release once this is merged? We should update the changelog as well, and specify which functions were deprecated.

defunctio: I would think we would want a minor version update just to indicate when they were deprecated. I'll update the changelog unless requested otherwise.

src/mmap.rs
defunctio: Ah, true. I suppose the deprecated functions currently cover the iterator implementation for now too. I'll revert these.

Cargo.toml
andreeaflorescu: I would prefer to update the version on the same PR when we are doing the release to follow the same (undocumented :laughing: ) process as before.

andreeaflorescu: In this case would the `deprecated` macro still work?

defunctio: Yep, still works. It would seem that when the deprecated version is a future version, the current version/branch also displays the deprecation warning. Do you have a version number preference for deprecation?

andreeaflorescu: For the next release we want to include dirty page tracking, so I think we will have it at `0.6.0`. I would just add a no-op issue to validate that we are releasing the version specified in the `deprecate` macros.

andreeaflorescu:
The warnings look legit to me. Should we update the atomic.rs tests to use iter instead of `with_regions`?

defunctio:
> The warnings look legit to me. Should we update the atomic.rs tests to use iter instead of `with_regions`?

That would be the tests for atomic.rs and mmap.rs I updated in the original PR :sweat_smile: . Let me add them back.

andreeaflorescu:
> > The warnings look legit to me. Should we update the atomic.rs tests to use iter instead of `with_regions`?
> 
> That would be the tests for atomic.rs and mmap.rs I updated in the original PR  . Let me add them back.

Sorry about that. I didn't realize that the reason you were changing the tests was because of the warnings.

defunctio:
> Sorry about that. I didn't realize that the reason you were changing the tests was because of the warnings.

No worries, it wasn't clear :)

andreeaflorescu:
> Just code coverage complaints from CI now.

Can you update the coverage score? You can do that by changing the [`coverage_config_x86_64.json`](https://github.com/rust-vmm/vm-memory/blob/master/coverage_config_x86_64.json) file with the new coverage value.

andreeaflorescu (approved):

jiangliu (approved):

PR #145, opened 2021-03-09
benches: fix safe conversion

iulianbarbu:
# Reason for this PR

`usize` is architecture dependent and can represent 4/8 bytes, while `u64` is an 8 byte data type. The function `create_guest_memory_mmap` implies a hidden conversion from `u64` to `usize`, which might imply data loss.

# Description of this PR

Implied conversions that can not lose data for the above case.

### DIFF
--- a/benches/main.rs
+++ b/benches/main.rs
@@ -15,10 +15,12 @@ mod volatile;
 use volatile::benchmark_for_volatile;
 
 #[cfg(feature = "backend-mmap")]
-pub fn create_guest_memory_mmap(size: u64, count: u64) -> GuestMemoryMmap {
+// Use this function with caution. It does not check against overflows
+// and `GuestMemoryMmap::from_ranges` errors.
+fn create_guest_memory_mmap(size: usize, count: u64) -> GuestMemoryMmap {
     let mut regions: Vec<(GuestAddress, usize)> = Vec::new();
     for i in 0..count {
-        regions.push((GuestAddress(i * size), size as usize));
+        regions.push((GuestAddress(i * size as u64), size));
     }
 
     GuestMemoryMmap::from_ranges(regions.as_slice()).unwrap()
--- a/benches/mmap/mod.rs
+++ b/benches/mmap/mod.rs
@@ -15,7 +15,7 @@ use criterion::{black_box, Criterion};
 
 use vm_memory::{ByteValued, Bytes, GuestAddress, GuestMemory};
 
-const REGION_SIZE: u64 = 0x8000_0000;
+const REGION_SIZE: usize = 0x8000_0000;
 const REGIONS_COUNT: u64 = 8;
 const ACCESS_SIZE: usize = 0x200;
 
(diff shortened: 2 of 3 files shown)


benches/main.rs
```diff
@@ -15,10 +15,11 @@ mod volatile;
 use volatile::benchmark_for_volatile;
 
 #[cfg(feature = "backend-mmap")]
-pub fn create_guest_memory_mmap(size: u64, count: u64) -> GuestMemoryMmap {
+fn create_guest_memory_mmap(size: usize, count: usize) -> GuestMemoryMmap {
     let mut regions: Vec<(GuestAddress, usize)> = Vec::new();
     for i in 0..count {
-        regions.push((GuestAddress(i * size), size as usize));
+        // Safe conversion to `u64`, since it does not truncate.
```
lauralt, line 21: I don't think that an `(usize * usize) as u64` can ever be safe, or even `(usize * whatever_type) as u64` (on 64-bit platforms especially) since it can overflow; using `as` will result in panicking if we try to multiply with overflow. We can use `checked_mul()` or something similar here to stay on the safe side. Also, I don't think the `usize -> u64` change of the `count` type in the function signature is necessary useful. Since you need an `u64` here you can just `size as u64 * count` (which is, of course, unsafe).

iulianbarbu: Since this function is used only for benches, I would avoid investing too much effort in handling overflows and maybe the unwrap when `GuestMemoryMmap` is created from ranges. We must not have undefined behavior when using this code, which can happen while converting, but panicking due to improper usage in the benches code, can be directly detected if it happens, so I would not stress about it too much. Do you agree?

Related to the second point, I think you meant `u64 -> usize` change from the signature of the `count` parameter. If this is the case, I agree we can keep `count` as `u64` and only convert `size` from `usize` to `u64` when we form the `GuestAddress`. WDYT?

lauralt: > Since this function is used only for benches, I would avoid investing too much effort in handling overflows and maybe the unwrap when `GuestMemoryMmap` is created from ranges. We must not have undefined behavior when using this code, which can happen while converting, but panicking due to improper usage in the benches code, can be directly detected if it happens, so I would not stress about it too much. Do you agree?

:+1: the comment should be updated though 
> Related to the second point, I think you meant `u64 -> usize` change from the signature of the `count` parameter. If this is the case, I agree we can keep `count` as `u64` and only convert `size` from `usize` to `u64` when we form the `GuestAddress`. WDYT?

Yes, that's what I meant.

benches/mmap/mod.rs
```diff
@@ -90,12 +90,12 @@ pub fn benchmark_for_mmap(c: &mut Criterion) {
     let accesses = &[
         AccessKind::InRegion(0),
         AccessKind::CrossRegion(0),
-        AccessKind::CrossRegion(REGIONS_COUNT - 2),
-        AccessKind::InRegion(REGIONS_COUNT - 1),
+        AccessKind::CrossRegion((REGIONS_COUNT - 2) as u64),
```
lauralt, line 93: `REGIONS_COUNT` is already an u64, no need for the explicit conversion

benches/mmap/mod.rs
```diff
@@ -53,10 +53,10 @@ enum AccessKind {
 }
 
 impl AccessKind {
-    fn make_offset(&self, access_size: usize) -> u64 {
+    fn make_offset(&self, access_size: usize) -> usize {
```
lauralt, line 56: Can't we leave this line as it was before and avoid all the `let obj_off = access.make_offset(size_of::<SmallDummy>()) as u64;` changes? You are also doing now an unsafe conversion below: the `idx as usize` (idx is an u64).

iulianbarbu:
@lauralt , PTAL! I've addressed you comments.

benches/main.rs
lauralt: The comment is still not rephrased/removed.

lauralt (approved):

jiangliu (approved):

Issue #141, opened 2021-03-01
Simplify the vm-memory interfaces and implementation


andreeaflorescu:
This is a high level issue (and should be updated as we find other simplification directions).

- [ ] https://github.com/rust-vmm/vm-memory/issues/123
- [ ] https://github.com/rust-vmm/vm-memory/issues/133

After closing #140 we can identify further simplification directions.

PR #148, opened 2021-06-06
Release v0.6

jiangliu:
Release vm-memory v0.6, with latest dirty track bitmap implementation.

### DIFF
--- a/src/volatile_memory.rs
+++ b/src/volatile_memory.rs
@@ -1368,9 +1368,8 @@ mod tests {
     use std::mem::size_of_val;
     use std::path::Path;
     use std::sync::atomic::{AtomicUsize, Ordering};
-    use std::sync::Arc;
-    use std::thread::{sleep, spawn};
-    use std::time::Duration;
+    use std::sync::{Arc, Barrier};
+    use std::thread::spawn;
 
     use matches::assert_matches;
     use vmm_sys_util::tempfile::TempFile;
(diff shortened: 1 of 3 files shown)


sboeuf (dismissed):

andreeaflorescu:
@jiangliu can we push this a bit until we fix this comment: https://github.com/rust-vmm/vm-memory/pull/125#discussion_r643697285

@lauralt wants to submit a PR with it.

jiangliu:
> @jiangliu can we push this a bit until we fix this comment: [#125 (comment)](https://github.com/rust-vmm/vm-memory/pull/125#discussion_r643697285)
> 
> @lauralt wants to submit a PR with it.

sure, then I will try to create a patch for #149

bonzini:
I wonder if it's a good idea to break the exported types due to the dirty bitmap implementation. Perhaps export the new version from some sub-package, and leave those exported directly from `vm_memory` as aliases with `()` as the bitmap type?

jiangliu:
> I wonder if it's a good idea to break the exported types due to the dirty bitmap implementation. Perhaps export the new version from some sub-package, and leave those exported directly from `vm_memory` as aliases with `()` as the bitmap type?

I have give a try, but it always breaks existing users of GuestMemoryRegion. For example, I have tried to rename "GuestMemoryRegion" to "GuestMemoryRegionImpl", and defines
```
pub trait GuestMemoryRegion: GuestMemoryRegionImpl<B = ()> {}
```
Then existing users need to "use GuestMemoryRegionImpl" too, which still breaks:(

jiangliu:
I have tried several times to avoid breaking compatibility but failed:(
So should we release v1.0.0 instead of v0.6.0?

andreeaflorescu:
We don't really give any guarantees related to the interface until we reach 1.0.0. I think we were also considering some other interface simplifications (#141), and I would be more inclined to first agree on those, implement them, and only then release 1.0.0.

For this one, I would be fine with 0.6.0 even if we break the interface.

src/volatile_memory.rs
```diff
@@ -1574,6 +1576,7 @@ mod tests {
         // somehow and this helps to ensure the sleep actually happens before the store rather then
```
lauralt, line 1576: I would update the comments, so they do not mention about `sleep` anymore.

jiangliu: fixed

alexandruag (approved):

PR #150, opened 2021-06-07
Implemnt builder for MmapRegion

jiangliu:
Currently the MmapRegion provides several methods to create MmapRegion
objects, but we often need to change the existing methods to support
new functionalities. So it would be better to adopt Builder pattern
and implement MmapRegionBuilder, so we could add new functionality
without affecting existing users.

### DIFF
--- a/src/mmap_unix.rs
+++ b/src/mmap_unix.rs
@@ -74,6 +74,152 @@ impl error::Error for Error {}
 
 pub type Result<T> = result::Result<T, Error>;
 
+/// A factory struct to build `MmapRegion` objects.
+pub struct MmapRegionBuilder<B = ()> {
+    size: usize,
+    prot: i32,
+    flags: i32,
+    file_offset: Option<FileOffset>,
+    raw_ptr: Option<*mut u8>,
+    hugetlbfs: Option<bool>,
+    bitmap: Option<B>,
+}
+
+impl<B: NewBitmap> MmapRegionBuilder<B> {
+    /// Create a new `MmapRegionBuilder`.
+    pub fn new(size: usize) -> Self {
+        MmapRegionBuilder {
+            size,
+            prot: 0,
+            flags: 0,
+            file_offset: None,
+            raw_ptr: None,
+            hugetlbfs: None,
+            bitmap: None,
+        }
+    }
+
+    /// Create the `MmapRegion` object with the specified mmap memory protection flag `prot`.
+    pub fn with_mmap_prot(mut self, prot: i32) -> Self {
+        self.prot = prot;
+        self
+    }
+
+    /// Create the `MmapRegion` object with the specified mmap `flags`.
+    pub fn with_mmap_flags(mut self, flags: i32) -> Self {
+        self.flags = flags;
+        self
+    }
+
+    /// Create the `MmapRegion` object with the specified `file_offset`.
+    pub fn with_file_offset(mut self, file_offset: FileOffset) -> Self {
+        self.file_offset = Some(file_offset);
+        self
+    }
+
+    /// Create the `MmapRegion` object with the specified `hugetlbfs` flag.
+    pub fn with_hugetlbfs(mut self, hugetlbfs: bool) -> Self {
+        self.hugetlbfs = Some(hugetlbfs);
+        self
+    }
+
+    /// Create the `MmapRegion` object with the specified `bitmap`.
+    pub fn with_bitmap(mut self, bitmap: B) -> Self {
+        self.bitmap = Some(bitmap);
+        self
+    }
+
+    /// Create the `MmapRegion` object with pre-mmapped raw pointer.
+    ///
+    /// # Safety
+    ///
+    /// To use this safely, the caller must guarantee that `raw_addr` and `self.size` define a
+    /// region within a valid mapping that is already present in the process.
+    pub unsafe fn with_raw_mmap_pointer(mut self, raw_ptr: *mut u8) -> Self {
+        self.raw_ptr = Some(raw_ptr);
+        self
+    }
+
+    /// Build the `MmapRegion` object.
+    pub fn build(mut self) -> Result<MmapRegion<B>> {
+        if self.raw_ptr.is_some() {
+            return self.build_raw();
+        }
+
+        // Forbid MAP_FIXED, as it doesn't make sense in this context, and is pretty dangerous
+        // in general.
+        if self.flags & libc::MAP_FIXED != 0 {
+            return Err(Error::MapFixed);
+        }
+
+        let (fd, offset) = if let Some(ref f_off) = self.file_offset {
+            check_file_offset(f_off, self.size)?;
+            (f_off.file().as_raw_fd(), f_off.start())
+        } else {
+            (-1, 0)
+        };
+
+        // This is safe because we're not allowing MAP_FIXED, and invalid parameters cannot break
+        // Rust safety guarantees (things may change if we're mapping /dev/mem or some wacky file).
+        let addr = unsafe {
+            libc::mmap(
+                null_mut(),
+                self.size,
+                self.prot,
+                self.flags,
+                fd,
+                offset as libc::off_t,
+            )
+        };
+
+        if addr == libc::MAP_FAILED {
+            return Err(Error::Mmap(io::Error::last_os_error()));
+        }
+
+        let bitmap = self.bitmap.take().unwrap_or_else(|| B::with_len(self.size));
+
+        Ok(MmapRegion {
+            addr: addr as *mut u8,
+            size: self.size,
+            bitmap,
+            file_offset: self.file_offset,
+            prot: self.prot,
+            flags: self.flags,
+            owned: true,
+            hugetlbfs: self.hugetlbfs,
+        })
+    }
+
+    fn build_raw(mut self) -> Result<MmapRegion<B>> {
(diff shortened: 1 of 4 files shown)


src/mmap_unix.rs
```diff
@@ -74,6 +74,156 @@ impl error::Error for Error {}
...
+    }
+
+    /// Create the `MmapRegion` object with the specified mmap `flags`.
+    pub fn with_mmap_flags(mut self, flags: i32) -> Self {
+        self.flags = flags;
+        self
+    }
+
+    /// Modify mmap flags of the builder.
+    pub fn set_mmap_flags(mut self, set: i32, clear: i32) -> Self {
```
andreeaflorescu, line 115: Can we split this in two functions? `set_mmap_flags` and `clear_mmap_flags`. When reading the code that calls `set_mmap_flags` it will not be obvious what the flags represent. We will also need to make clear what is the difference between `set_mmap_flags` and `with_mmap_flags`. Maybe we can name `set_mmap_flags` to `add_mmap_flags` instead because then the name suggests that the previous set flags are not overwritten?

src/mmap_unix.rs
```diff
@@ -74,6 +74,156 @@ impl error::Error for Error {}
...
+    bitmap: B,
+}
+
+impl<B: Bitmap + NewBitmap> MmapRegionBuilder<B> {
+    /// Create a new MmapRegionBuilder.
+    pub fn new(size: usize) -> Self {
+        MmapRegionBuilder {
+            size,
+            prot: libc::PROT_READ | libc::PROT_WRITE,
+            flags: libc::MAP_ANONYMOUS | libc::MAP_NORESERVE | libc::MAP_PRIVATE,
```
andreeaflorescu, line 94: Is it expected for these fields to have this value initially?

src/mmap_unix.rs
```diff
@@ -74,6 +74,156 @@ impl error::Error for Error {}
...
+            file_offset: None,
+            raw_ptr: None,
+            hugetlbfs: None,
+            bitmap: B::with_len(size),
```
andreeaflorescu, line 98: I didn't review the dirty bitmap tracking PR, so sorry if this is a stupid question. Is it expected for all MmapRegions to have an associated bitmap? I was wondering if there is a way to opt out which we might want to factor in here in the builder.

I find it a bit strange that we initialize it here with a default value, and then we also have a `with_bitmap` function because that suggests that the Bitmap is not initialized beforehand.

src/mmap_unix.rs
```diff
@@ -74,6 +74,156 @@ impl error::Error for Error {}
...
+        // in general.
+        if self.flags & libc::MAP_FIXED != 0 {
+            return Err(Error::MapFixed);
+        }
+
+        if self.raw_ptr.is_some() {
+            return self.build_raw();
+        }
+
+        let (fd, offset) = if let Some(ref f_off) = self.file_offset {
```
andreeaflorescu, line 163: Should we move this check where the user can set the file_offset instead?

src/mmap_unix.rs
jiangliu: Changed bitmap to `Option<B>` and will create the required bitmap on demand.

src/mmap_unix.rs
jiangliu: By that way, we need to enhance the builder to return error code and that may cause inconsistence to Builder methods. So do the check when really building the MmapRegion object.

src/mmap_unix.rs
jiangliu: good suggestion, setting `prot` and `flags` to zero simplifies the implementations:)

src/mmap_unix.rs
jiangliu: When `flags` is initialized to zero, we don't need this interface anymore:)

src/mmap_unix.rs
```diff
@@ -74,6 +74,152 @@ impl error::Error for Error {}
...
+    pub fn with_bitmap(mut self, bitmap: B) -> Self {
+        self.bitmap = Some(bitmap);
+        self
+    }
+
+    /// Create the `MmapRegion` object with pre-mmapped raw pointer.
+    ///
+    /// # Safety
+    ///
+    /// To use this safely, the caller must guarantee that `addr` and `size` define a region within
```
lauralt, line 136:
```suggestion
    /// To use this safely, the caller must guarantee that `raw_ptr` and `self.size` define a region within
```

src/mmap_unix.rs
```diff
@@ -74,6 +74,152 @@ impl error::Error for Error {}
...
+    size: usize,
+    prot: i32,
+    flags: i32,
+    file_offset: Option<FileOffset>,
+    raw_ptr: Option<*mut u8>,
+    hugetlbfs: Option<bool>,
+    bitmap: Option<B>,
+}
+
+impl<B: Bitmap + NewBitmap> MmapRegionBuilder<B> {
```
lauralt, line 88: Isn't `NewBitmap` enough here?

src/mmap_unix.rs
```diff
@@ -74,6 +74,152 @@ impl error::Error for Error {}
...
+    pub unsafe fn with_raw_mmap_pointer(mut self, raw_ptr: *mut u8) -> Self {
+        self.raw_ptr = Some(raw_ptr);
+        self
+    }
+
+    /// Build the `MmapRegion` object.
+    pub fn build(mut self) -> Result<MmapRegion<B>> {
+        // Forbid MAP_FIXED, as it doesn't make sense in this context, and is pretty dangerous
+        // in general.
+        if self.flags & libc::MAP_FIXED != 0 {
```
lauralt, line 147: Does this still make sense if `self.raw_ptr` is `Some`?

src/mmap_unix.rs
```diff
@@ -74,6 +74,152 @@ impl error::Error for Error {}
...
+
+        // Check that the size is a multiple of the page size.
+        if self.size & (page_size - 1) != 0 {
+            return Err(Error::InvalidSize);
+        }
+
+        Ok(MmapRegion {
+            addr,
+            size: self.size,
+            bitmap: self.bitmap.unwrap(),
```
lauralt, line 213: We should add a comment why this unwrap is safe.

src/mmap_unix.rs
andreeaflorescu: Sounds good.

andreeaflorescu (dismissed):

src/mmap_unix.rs
jiangliu: The with_bitmap() method needs "B: Bitmap" bound.

src/mmap_unix.rs
jiangliu: fixed:)

src/mmap_unix.rs
jiangliu: solved in another way.

src/mmap_unix.rs
jiangliu: Good point, we should skip the check for build_raw().

src/mmap_unix.rs
lauralt: Do we have to add it explicitly if NewBitmap is a Bitmap itself?

jiangliu: yeah, not realized `NewBitmap: Bitmap`.

src/mmap_unix.rs
```diff
@@ -74,6 +74,152 @@ impl error::Error for Error {}
...
+            return Err(Error::InvalidSize);
+        }
+
+        let bitmap = self.bitmap.take().unwrap_or_else(|| B::with_len(self.size));
+
+        Ok(MmapRegion {
+            addr,
+            size: self.size,
+            bitmap,
+            file_offset: self.file_offset,
```
lauralt, line 214: Can `file_offset` be `Some` here?

jiangliu: It's possible with new builder api.

lauralt: So if the user provides a file_offset and a raw_ptr when building `MmapRegionBuilder`, file_offset is basically ignored. Should `file_offset` be None here then? :-?

jiangliu: Even with  build_raw(), user still could attach an file object to the MmapRegion object and retrieve the attached file by MmapRegion::file_offset(). This way it will be helpful for the user to make use of build_raw().

alexandruag:
Hi Gerry, thanks for the PR, it looks really helpful. My only comment is a proposal about basing the builder on a `Bitmap` bound instead of `NewBitmap`. The latter was only introduced to support bitmap objects without altering the existing constructor interfaces of `MmapRegion`, `GuestRegionMmap`, etc. (btw, as many of us have already noticed, the current interface is quite inflexible so adding more builders going forward would be super helpful). However, `NewBitmap` also generates some awkwardness; for example, if we use an `Option<Bitmap>` there's no way to convey whether we want a `None` or `Some(something)`.

To that end, and since this is a new interface, how about we use a `impl<B: Bitmap> MmapRegionBuilder<B>` trait bound, remove the `with_bitmap` method, and add a `bitmap: B` argument to the `build` and `build_raw` methods instead? This way we are not required to implement `NewBitmap` for types where it might not make a lot of sense (and maybe we can get rid of it entirely at some point). I had some other options in mind as well, but was really curious what are your thoughts about this in particular.

PR #155, opened 2021-06-30
Implement builder for MmapRegion v2

alexandruag:
This PR is entirely based on Gerry's earlier PR from [here](https://github.com/rust-vmm/vm-memory/pull/150), together with several changes with the aim of having a `B: Bitmap` trait bound (instead of `B: NewBitmap`) for the `B` generic type parameter of `MmapRegionBuilder` (a bit more context in [this](https://github.com/rust-vmm/vm-memory/pull/150#issuecomment-867705404) comment).

### DIFF
--- a/src/mmap_unix.rs
+++ b/src/mmap_unix.rs
@@ -74,6 +74,154 @@ impl error::Error for Error {}
 
 pub type Result<T> = result::Result<T, Error>;
 
+/// A factory struct to build `MmapRegion` objects.
+pub struct MmapRegionBuilder<B = ()> {
+    size: usize,
+    prot: i32,
+    flags: i32,
+    file_offset: Option<FileOffset>,
+    raw_ptr: Option<*mut u8>,
+    hugetlbfs: Option<bool>,
+    bitmap: B,
+}
+
+impl<B: Bitmap + Default> MmapRegionBuilder<B> {
+    /// Create a new `MmapRegionBuilder` using the default value for
+    /// the inner `Bitmap` object.
+    pub fn new(size: usize) -> Self {
+        Self::new_with_bitmap(size, B::default())
+    }
+}
+
+impl<B: Bitmap> MmapRegionBuilder<B> {
+    /// Create a new `MmapRegionBuilder` using the provided `Bitmap` object.
+    ///
+    /// When instantiating the builder for a region that does not require dirty bitmap
+    /// bitmap tracking functionality, we can specify a trivial `Bitmap` implementation
+    /// such as `()`.
+    pub fn new_with_bitmap(size: usize, bitmap: B) -> Self {
+        MmapRegionBuilder {
+            size,
+            prot: 0,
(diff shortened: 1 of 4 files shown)


src/mmap_unix.rs
```diff
@@ -510,6 +601,13 @@ mod tests {
         assert_eq!(r.prot(), libc::PROT_READ | libc::PROT_WRITE);
         assert_eq!(r.flags(), libc::MAP_NORESERVE | libc::MAP_PRIVATE);
         assert!(r.owned());
+
+        let builder = MmapRegionBuilder::new(0x1_0000, ())
```
lauralt, line 605: Can we test the builder with an actual bitmap as well?

src/mmap_unix.rs
```diff
@@ -74,6 +74,142 @@ impl error::Error for Error {}
...
+    prot: i32,
+    flags: i32,
+    file_offset: Option<FileOffset>,
+    raw_ptr: Option<*mut u8>,
+    hugetlbfs: Option<bool>,
+    bitmap: B,
+}
+
+impl<B: Bitmap> MmapRegionBuilder<B> {
+    /// Create a new `MmapRegionBuilder`.
```
lauralt, line 89: Should we add a comment mentioning that the `bitmap` parameter needs to be `()` if the user doesn't want the dirty page tracking?

src/mmap_unix.rs
```diff
@@ -74,6 +74,142 @@ impl error::Error for Error {}
...
+    pub fn new(size: usize, bitmap: B) -> Self {
```
jiangliu, line 90: how about renaming "new()" to "new_with_bitmap()" and add:
```
impl<B: Bitmap + Default> MmapRegionBuilder<B> {
         pub fn new(size: usize) -> Self {
               Self::new_with_bitmap(size, B::default)
         }
}

src/mmap_unix.rs
alexandruag: Hmm it looks like building with an actual `Bitmap` wouldn't make that much of a difference here because we're not testing the actual tracking functionality as well; the bitmap is just moved around. Did you have something more than that in mind?

src/mmap_unix.rs
alexandruag: Added a comment about that.

src/mmap_unix.rs
alexandruag: Cool, done.

alexandruag (commented):
Thanks for the comments. I've pushed a new version.

src/mmap_unix.rs
alexandruag: Ok we chatted a bit, and now I'm building and configuring the region with a non-trivial bitmap, calling `build`, and applying the generic volatile memory dirty tracking test on the result. I've also changed the default flags value for builder to `ANONYMOUS | PRIVATE` since that seems more like an  expected default value.

lauralt (approved):

jiangliu (approved):

PR #157, opened 2021-07-09
bitmap: add Default bound to trait BitmapSlice

jiangliu:
When migrating vm-memory consumers to the new vm-memory interfaces, they
have dependence on "()::default". So add a "BitmapSlice: Default" to
ease the migration.

### DIFF
--- a/src/bitmap/backend/ref_slice.rs
+++ b/src/bitmap/backend/ref_slice.rs
@@ -29,6 +29,12 @@ impl<'a, B: Bitmap> WithBitmapSlice<'a> for RefSlice<'_, B> {
 
 impl<B: Bitmap> BitmapSlice for RefSlice<'_, B> {}
 
+impl<B: Bitmap> Default for RefSlice<'_, B> {
+    fn default() -> Self {
+        panic!("default() method is not supported for RefSlice");
+    }
+}
+
 impl<'a, B: Bitmap> Bitmap for RefSlice<'a, B> {
     /// Mark the memory range specified by the given `offset` (relative to the base offset of
     /// the slice) and `len` as dirtied.
--- a/src/bitmap/mod.rs
+++ b/src/bitmap/mod.rs
@@ -25,7 +25,7 @@ pub trait WithBitmapSlice<'a> {
 /// Trait used to represent that a `BitmapSlice` is a `Bitmap` itself, but also satisfies the
 /// restriction that slices created from it have the same type as `Self`.
 pub trait BitmapSlice:
-    Bitmap + Clone + Copy + Debug + for<'a> WithBitmapSlice<'a, S = Self>
+    Bitmap + Clone + Copy + Debug + Default + for<'a> WithBitmapSlice<'a, S = Self>
 {
 }
 


sboeuf (approved):

alexandruag:
Hi! I think that adding `Default` as a trait bound here is sort of unnatural for the slices (i.e. it would not be supported by most implementations IMO). Also, implementing the trait and generating a panic at runtime when `default()` is called (because it's not actually supported) looks quite dangerous and surprising :( Gerry, can you please share some more details around the problem we're trying to avoid here? Maybe we can identify some other way around it.

jiangliu:
The fuse-backend-rs crate is a fuse library to support fuse daemon or virtiofsd.  The fuse-backend-rs core needs to deal with two transport protocols: virtiofs-fs and /dev/fuse. After enabling dirty-tracking for virtio-fs transport layer, then we need to add dummy dirty tracking for /dev/fuse transport layer which doesn't use vm-memory. So we need "()::default()" to make it work.
```
impl<'a, S: BitmapSlice> Reader<'a, S> {
    /// Construct a new Reader wrapper over `desc_chain`.
    ///
    /// 'request`: Fuse request from clients read from /dev/fuse
    pub fn new(buf: FuseBuf<'a>) -> Result<Reader<'a, S>> {
        let mut buffers: VecDeque<VolatileSlice<'a, S>> = VecDeque::new();
        // Safe because Reader has the same lifetime with buf.
        buffers.push_back(unsafe {
            VolatileSlice::with_bitmap(buf.mem.as_mut_ptr(), buf.mem.len(), S::default())
        });

        Ok(Reader {
            buffers: IoBuffers {
                buffers,
                bytes_consumed: 0,
            },
        })
    }
}
``` 
And it's not ease to change "S: BitmapSlice" to "S: BitmapSlice + Default" here because it causes too much changes to users of Reader.

jiangliu:
> Hi! I think that adding `Default` as a trait bound here is sort of unnatural for the slices (i.e. it would not be supported by most implementations IMO). Also, implementing the trait and generating a panic at runtime when `default()` is called (because it's not actually supported) looks quite dangerous and surprising :( Gerry, can you please share some more details around the problem we're trying to avoid here? Maybe we can identify some other way around it.

So let's try to solve this problem in another way:
https://github.com/rust-vmm/vm-memory/pull/158

alexandruag:
Thanks for the examples Gerry! Looking at this (and at the current state from the main branch of `fuse-backend-rs`), it seems the `Reader` above only works with `FustBuf` (which warps a simple byte slice) which does not appear to require tracking; any particular reason why the generic `S` type parameter is added above instead of always using `()`? Also, just wondering, why does adding `: BitmapSlice + Default` (instead of only `: BitmapSlice`) make a difference when there was no bound before at all? We can chat on Slack if there are more nuances here, and I've also started looking at the PR you posted.

jiangliu:
> Thanks for the examples Gerry! Looking at this (and at the current state from the main branch of `fuse-backend-rs`), it seems the `Reader` above only works with `FustBuf` (which warps a simple byte slice) which does not appear to require tracking; any particular reason why the generic `S` type parameter is added above instead of always using `()`? Also, just wondering, why does adding `: BitmapSlice + Default` (instead of only `: BitmapSlice`) make a difference when there was no bound before at all? We can chat on Slack if there are more nuances here, and I've also started looking at the PR you posted.

There's another version when virtio-fs is enabled and the  core file system code depends on `Reader<'_, S>` so we could seamless switch between virtio-fs and fusedev transport layers.
```
impl<'a, M: GuestMemory> Reader<'a, M> {
    /// Construct a new Reader wrapper over `desc_chain`.
    #[allow(clippy::new_ret_no_self)]
    pub fn new(mem: &'a M, desc_chain: DescriptorChain<'a, M>) -> Result<Reader<'a, MS<'a, M>>> {
    }
}

impl<F: FileSystem<S> + Sync, D: AsyncDrive, S: BitmapSlice> Server<F, D, S> {
    pub fn handle_message(
        &self,
        mut r: Reader<'_, S>,
        w: Writer<'_, S>,
        vu_req: Option<&mut dyn FsCacheReqHandler>,
        hook: Option<&dyn MetricsHook>,
    ) -> Result<usize> {
    }
}
```

alexandruag:
Oh I see, thanks for the extra details. I have some ideas about how we could fix that, but I think it's quicker if we pause to have a chat to flesh out the context a bit more, and then circle back here with a summary.

PR #158, opened 2021-07-14
bitmap: simplify dirty tracking implementation

jiangliu:
We have implemented guest memory dirty tracking by using atomic
integers. When migrating existing user to the new vm-memory interface
with dirty tracking, the dirty tracking subsystem is too complex
with three traits BitmapSlice, WithBitmapSlice and Bitmap. And the
most challege issue is to deal with the lifetime of RefSlice.

So let us simplify the dirty tracking interface a bit, with one
Bitmap trait and AtomicBitmap struct, without any lifetime parameter.


alexandruag:
Just finished having a better look; overall, I think there's value in keeping the three traits, while also addressing the reference-counted approach to slicing into bitmaps that's effectively implemented in the PR. To quickly re-cap the purpose of the traits, `WithBitmapSlice` is essentially a workaround based on HRTBs (higher ranked trait bounds) to have an associated type which may have a lifetime parameter as well (but doesn't have to), and we can get rid of it when more straightforward support lands in Rust. `BitmapSlice` is present because, unlike just `Bitmap`, it carries the additional semantics that a slice taken from a `BitmapSlice` must have the same type, and cloning/copying a slice keeps referring to the same `Bitmap` object. Getting rid of everything except `Bitmap` basically forces implementors to always use some sort of reference counting (even if plain references or an equivalent would be preferable), as well as `Clone` for `Bitmap` implementations always having reference-counted semantics.

One way to harmonize these things is to add something like an `ArcSlice` (for lack of a better term), which does pretty much what happens in this PR, and then keep `AtomicBitmap` with the `RefSlice` and also have an `AtomicBitmapArc` that uses the reference-counted slicing approach. There's a lot of overlap between these abstractions, so they can be implemented with just a few additions on top of what we already have. I'm building an example, and will return and post it here.

jiangliu:
> Just finished having a better look; overall, I think there's value in keeping the three traits, while also addressing the reference-counted approach to slicing into bitmaps that's effectively implemented in the PR. To quickly re-cap the purpose of the traits, `WithBitmapSlice` is essentially a workaround based on HRTBs (higher ranked trait bounds) to have an associated type which may have a lifetime parameter as well (but doesn't have to), and we can get rid of it when more straightforward support lands in Rust. `BitmapSlice` is present because, unlike just `Bitmap`, it carries the additional semantics that a slice taken from a `BitmapSlice` must have the same type, and cloning/copying a slice keeps referring to the same `Bitmap` object. Getting rid of everything except `Bitmap` basically forces implementors to always use some sort of reference counting (even if plain references or an equivalent would be preferable), as well as `Clone` for `Bitmap` implementations always having reference-counted semantics.
> 
> One way to harmonize these things is to add something like an `ArcSlice` (for lack of a better term), which does pretty much what happens in this PR, and then keep `AtomicBitmap` with the `RefSlice` and also have an `AtomicBitmapArc` that uses the reference-counted slicing approach. There's a lot of overlap between these abstractions, so they can be implemented with just a few additions on top of what we already have. I'm building an example, and will return and post it here.

I have also worked out a high performance version by using unsafe code to replace Arc. Basically I'm trying to avoid lifetime parameter, which does cause many pitfalls.
```
    /// A fast version of `AtomicBitmap` by using unsafe `*const AtomicBitmapState` instead of
    /// `Arc<AtomicBitmapState>`.
    ///
    /// The object generated by `AtomicBitmap::new()` owns the underlying `AtomicBitmapState`
    /// object, and those objects generated by `AtomicBitmap::slice_at()` hold immutable references
    /// to the underlying `AtomicBitmapState` object. So the caller must take the responsibility to
    /// correctly manage lifetime of `AtomicBitmapState` objects. The GuestMemoryRegion owns the
    /// `AtomicBitmap` object and satisfies the requirement.
    #[derive(Debug)]
    pub struct AtomicBitmap {
        // The `state` field store a pointer of `* const AtomicBitmapState` and some flags, assuming
        // the lowest three bits are not used by pointer.
        // Bit 0: 1 if current object owns the AtomicBitmapState object
        // Bit 3-32/64: *const AtomicBitmapState
        state: usize,
        base: usize,
    }

    #[allow(clippy::len_without_is_empty)]
    impl AtomicBitmap {
        /// Create a new bitmap of `byte_size`, with one bit per page.
        ///
        /// This is effectively rounded up, and we get a new vector of the next multiple of 64 bigger
        /// than `bit_size`.
        pub fn new(byte_size: usize, page_size: usize) -> Self {
            let state = Box::into_raw(Box::new(AtomicBitmapState::new(byte_size, page_size)));
            let bitmap = AtomicBitmap {
                state: (state as usize) | 0x1,
                base: 0,
            };

            bitmap
        }

        /// Get the length of the bitmap in bits (i.e. in how many pages it can represent).
        pub fn len(&self) -> usize {
            self.state().len().saturating_sub(self.base)
        }

        /// Get the underlying dirty tracking state object.
        pub fn state(&self) -> &AtomicBitmapState {
            unsafe { &*((self.state & !0x7) as *const AtomicBitmapState) }
        }
    }

    impl Default for AtomicBitmap {
        fn default() -> Self {
            AtomicBitmap::new(0, 0x1000)
        }
    }
    impl Clone for AtomicBitmap {
        fn clone(&self) -> Self {
            AtomicBitmap {
                state: self.state & !0x1,
                base: self.base,
            }
        }
    }

    impl Drop for AtomicBitmap {
        fn drop(&mut self) {
            if self.state & 0x1 != 0 {
                let state = self.state & !0x7;

                unsafe { Box::from_raw(state as *mut AtomicBitmapState) };
            }
        }
    }

   impl Bitmap for AtomicBitmap {
        fn mark_dirty(&self, offset: usize, len: usize) {
            self.state()
                .set_addr_range(offset.saturating_add(self.base), len)
        }

        fn dirty_at(&self, offset: usize) -> bool {
            self.state().is_addr_set(offset.saturating_add(self.base))
        }

        fn slice_at(&self, base: usize) -> Self {
            Self {
                state: self.state & !0x1,
                base: self.base.saturating_add(base),
            }
        }
    }

```

alexandruag:
That's an interesting approach! The `unsafe` blocks are kinda scary; I'd need to take a longer look to (dis)confirm, but from a high level it seems like there's some race condition danger. I experimented a bit and created a prototype in [this](https://github.com/alexandruag/vm-memory/commits/arcref) branch, which adds support for bitmap slices which don't necessarily have to have associated lifetimes. It does not solve the  problem described in https://github.com/rust-vmm/vm-memory/pull/157 on its own, but it's something we can include if it becomes part of a solution. Let's switch the discussion back to #157, and return here after we have a conclusion.

PR #160, opened 2021-07-27
Add `ArcRef` and `AtomicBitmapArc` bitmap backend implementations

alexandruag:
This PR adds support for using reference counted tracking abstractions with the `bitmap` backend. `ArcSlice` stands for a bitmap into a slice that's based on `Arc` objects instead of references, and thus does not require lifetimes. `AtomicBitmapArc` is a small wrapper over `Arc<AtomicBitmap>` that makes use of `ArcSlice`s. This simplifies situations where it was difficult to handle the additional lifetimes.

### DIFF
--- a/src/bitmap/backend/atomic_bitmap_arc.rs
+++ b/src/bitmap/backend/atomic_bitmap_arc.rs
@@ -1,0 +1,86 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+use std::ops::Deref;
+use std::sync::Arc;
+
+use crate::bitmap::{ArcSlice, AtomicBitmap, Bitmap, WithBitmapSlice};
+
+#[cfg(feature = "backend-mmap")]
+use crate::mmap::NewBitmap;
+
+/// A `Bitmap` implementation that's based on an atomically reference counted handle to an
+/// `AtomicBitmap` object.
+pub struct AtomicBitmapArc {
+    inner: Arc<AtomicBitmap>,
+}
+
+impl AtomicBitmapArc {
+    pub fn new(inner: AtomicBitmap) -> Self {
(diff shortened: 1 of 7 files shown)


src/bitmap/backend/atomic_bitmap_arc.rs
```diff
@@ -0,0 +1,81 @@
+use std::ops::Deref;
```
jiangliu, line 1: Please add license and copyright header.

src/bitmap/backend/slice.rs
```diff
@@ -48,31 +64,26 @@ impl<'a, B: Bitmap> Bitmap for RefSlice<'a, B> {
...
         // Dummy impl for now.
         write!(f, "(bitmap slice)")
     }
 }
 
+/// A `BitmapSlice` implementation that wraps a reference to a `Bitmap` object.
+pub type RefSlice<'a, B> = BaseSlice<&'a B>;
+
+/// A `BitmapSlice` implementation that uses an `Arc` handle to a `Bitmap` object.
+pub type ArcSlice<B> = BaseSlice<Arc<B>>;
```
jiangliu, line 85: Could you please also help to add some test cases for ArcSlice?

src/bitmap/backend/atomic_bitmap_arc.rs
alexandruag: done

src/bitmap/backend/slice.rs
alexandruag: The existing `test_ref_slice` unit test effectively covers `ArcSlice` as well (and all instantiations of `BaseSlice` basically), since the only difference is the smart pointer being used, which is transparent w.r.t. the test as long as `Deref` is properly implemented. Renamed `test_ref_slice` to `test_slice` to better express the generality.

jiangliu (approved):

jiangliu (approved):

lauralt (approved):

PR #186, opened 2022-01-26
do not return out bound error for read and write when volatile slice is empty

aftersnow:
To keep Consistent with the standard library, return ok(0) instead of out bound error when volatile slice is empty.


andreeaflorescu:
The change LGTM, but can you please check the CI failures? You should sign your commit, and add a test to check the newly introduced behavior.

lauralt:
@aftersnow can you update the PR or should we push to your branch? Thanks.

PR #190, opened 2022-02-21
fix read/write when volatile slice is empty

andreeaflorescu:
### Summary of the PR

This keeps the vm-memory interface consistent with the standard I/O, returning Ok(0) when the reader/writer is empty.

This PR adds a test on top of #186 and fixes the other issues reported by the CI so we can merge it.

### Requirements

requirements:

- [ ] All commits in this PR are signed (with `git commit -s`), and follow
  [the 50/72 git commit rule](https://www.midori-global.com/blog/2018/04/02/git-50-72-rule).
  test.


jiangliu (dismissed):

src/volatile_memory.rs
```diff
@@ -588,6 +588,10 @@ impl<B: BitmapSlice> Bytes<usize> for VolatileSlice<'_, B> {
     /// assert_eq!(res.unwrap(), 4);
     /// ```
     fn write(&self, buf: &[u8], addr: usize) -> Result<usize> {
+        if self.size == 0 {
```
andreeaflorescu, line 591: Hmm, I think we should check instead the size of buffer, not of self here.

lauralt: Yes so if the size of the initial slice is 0, I think we should actually fail for whatever operation we are trying to do to that slice.

andreeaflorescu: Updated it.

src/volatile_memory.rs
```diff
@@ -588,6 +588,10 @@ impl<B: BitmapSlice> Bytes<usize> for VolatileSlice<'_, B> {
     /// assert_eq!(res.unwrap(), 4);
     /// ```
     fn write(&self, buf: &[u8], addr: usize) -> Result<usize> {
+        if buf.len() == 0 {
```
lauralt, line 591: I think we are already returning Ok(0) in this case, but I'll double-check.

lauralt: This seems to be the case, so I think we can keep only the tests here.

andreeaflorescu: This is not the case when the volatile slice is empty. I've added a test case for both use cases:
- volatile slice is empty, read/write size is 0
- volatile slice is not empty, read/write size is 0

jiangliu (approved):

andreeaflorescu:
@bonzini can you take a look? it's a revamp of #186

stsquad (approved):
LGTM! :thumbsup:

PR #192, opened 2022-03-08
Add new method as_any() to GuestAddressSpace

jiangliu:
Add new method as_any() to GuestAddressSpace, which returns reference
to a std::any::Any trait object. Then downcast_ref() could be used to
get the concrete type of GuestAddressSpace trait object.

### Summary of the PR

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -12,6 +12,7 @@
 extern crate arc_swap;
 
 use arc_swap::{ArcSwap, Guard};
+use std::any::Any;
 use std::ops::Deref;
 use std::sync::{Arc, LockResult, Mutex, MutexGuard, PoisonError};
 
@@ -75,12 +76,16 @@ impl<M: GuestMemory> GuestMemoryAtomic<M> {
     }
 }
 
-impl<M: GuestMemory> GuestAddressSpace for GuestMemoryAtomic<M> {
-    type T = GuestMemoryLoadGuard<M>;
+impl<M: GuestMemory + 'static> GuestAddressSpace for GuestMemoryAtomic<M> {
     type M = M;
+    type T = GuestMemoryLoadGuard<M>;
 
     fn memory(&self) -> Self::T {
         GuestMemoryLoadGuard { guard: self.load() }
+    }
+
+    fn as_any(&self) -> &dyn Any {
+        self
     }
 }
 
@@ -147,6 +152,23 @@ mod tests {
     type GuestMemoryMmap = crate::GuestMemoryMmap<()>;
     type GuestRegionMmap = crate::GuestRegionMmap<()>;
     type GuestMemoryMmapAtomic = GuestMemoryAtomic<GuestMemoryMmap>;
+
+    #[test]
+    fn test_as_any() {
+        let region_size = 0x400;
+        let regions = vec![
+            (GuestAddress(0x0), region_size),
+            (GuestAddress(0x1000), region_size),
+        ];
+        let gmm = GuestMemoryMmap::from_ranges(&regions).unwrap();
+        let gm = GuestMemoryMmapAtomic::new(gmm);
+
+        assert!(gm
+            .as_any()
+            .downcast_ref::<GuestMemoryMmapAtomic>()
+            .is_some());
+        assert!(gm.as_any().downcast_ref::<String>().is_none());
+    }
 
     #[test]
     fn test_atomic_memory() {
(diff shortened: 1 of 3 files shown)


sboeuf (dismissed):
LGTM

src/atomic.rs
```diff
@@ -75,13 +76,17 @@ impl<M: GuestMemory> GuestMemoryAtomic<M> {
...
-    type T = GuestMemoryLoadGuard<M>;
+impl<M: GuestMemory + 'static> GuestAddressSpace for GuestMemoryAtomic<M> {
     type M = M;
+    type T = GuestMemoryLoadGuard<M>;
 
     fn memory(&self) -> Self::T {
         GuestMemoryLoadGuard { guard: self.load() }
     }
+
+    fn as_any(&self) -> &dyn Any {
```
alexandruag, line 87: Hi, super sorry for taking so long to look at this. The main question in my mind why does this method need to be part of the trait, and not composed or super-imposed somehow externally? Asking because the functionality is quite orthogonal to this trait (and does come up with some extra constraints; not sure how important, but they are there). Is this going to become a pattern that will repeat for other traits we expose in the future?

There are a bunch of ways to get similar functionality without embedding it in the trait proper when downcasting is needed locally in a project (one option/example is illustrated in [this](https://github.com/rust-vmm/event-manager/blob/main/tests/endpoint.rs) test from the `event-manager` crate). Another option is to define a separate trait just for the `Any` related part, and include it in a trait bound where required. How do these fit over the use cases you have in mind so far?

jiangliu: All existing code takes an <AS: GuestMemoryAddress>, so we have two ways to extend existing code:
1) Change all instances of `<AS: GuestMemoryAddress>` to `<AS: GuestMemoryAddress + AsAny>`
2) Enhance `GuestMemoryAddress` to provide `as_any()`.
So it's a tradeoff to reduce code changes.

alexandruag: Hi, sorry spent a bit more time experimenting with a few alternative solutions, but ultimately it turns out the simplest non-invasive (w.r.t. to the upstream interfaces) are still the aforementioned ones. So to describe the alternative in a bit more detail, I presume the mention of existing code having `<AS: GuestMemoryAddress>` refers to various local implementations. A more straightforward approach to altering each and every occurence of `<AS: GuestAddressSpace>` to `<AS: GuestAddressSpace + AsAny>` is to define a local trait also named `GuestAddressSpace` (to keep even simple renamings to a minimum) as `GuestAddressSpace = upstream::GuestAddressSpace + AsAny`. Then, by `use`-ing the local variant instead of the upstream one (which still requires some changes, but at a hopefully significantly reduced scale), all the existing method/impl definitions should work in a transparent manner (there are some caveats if trait objects are required, but from what we discussed so far that doesn't seem to be a requirement). Is that something that's feasible to do as opposed to the kind of alteration you mentioned?

Adding the method to the upstream trait is also something that can be done, but it really feels like it doesn't actually belong there and that it solves a completely orthogonal problem. Also, if we were to go that path, I'm worried that it might become tempting to do this more easily to other traits as well (particularly since this generates additional constraints and alters the semantics at least to some extent).

Ultimately, if the cost is really high for the alternative we can implement the changes for `GuestAddressSpace`, but I'm happy to help with a quick PoC to get the alternative rolling and/or showcase it's actually a working solution (if you can point me at a branch somewhere to add the changes on top of). I really thing it's better overall to avoid conflating the functionality upstream.

alexandruag (approved):
Had some discussions with Gerry and we concluded to move forward with this approach while considering how the upstream interface should evolve such that various pieces of functionality don't need to rely solely on downcasting.

lauralt:
@jiangliu can you fix the conflicts so we can merge the PR? Thanks.

jiangliu:
> @jiangliu can you fix the conflicts so we can merge the PR? Thanks.

@lauralt hi, I have updated the PR, thanks:)

lauralt (approved):

bonzini:
Sorry, I think the extra `'static` bound is a problem, especially when the GuestAddressSpace is just `&M`. I don't think this pull request can be salvaged, you should just use external crates like [`as_any`](https://docs.rs/as-any/latest/as_any/trait.AsAny.html) or [`downcast`](https://docs.rs/downcast/latest/downcast/).

bonzini:
(Did not mean to close, sorry).

bonzini:
Closing - the blockers about `'static` being too restrictive cannot really be solved in the context of this PR so might as well open a new one.

Issue #195, opened 2022-05-13
MmapRegion::build fails when the backing object is a device


slp:
Currently `mmap::check_file_offset` attempts to validate the backing object by checking, among other things, its size to confirm the mapping will fit on it. This check is implemented by using `std::fs::metadata` to obtain the file size, but this approach doesn't work when the backing object is a block or char device, as the returned value is always zero.

This was discovered when [virtiofsd](https://gitlab.com/virtio-fs/virtiofsd) users noticed that adding a memory object in QEMU backed by a device (i.e. `/dev/kvmfr0`) would cause the daemon to fail with `Waiting for daemon failed: HandleRequest(ReqHandlerError(Custom { kind: Other, error: MappingPastEof }))` ([issue 46](https://gitlab.com/virtio-fs/virtiofsd/-/issues/46)).

GZibrat:
Thanks for putting in a fix! However, I don't think this fixes the issue entirely. For example, running fstat on a vfio device returns 0 for the size which makes virtiofsd return MmapRegion(MappingPastEof).

Is there a reason in particular this check is needed at all? mmap will verify all of this and return failure if any condition is not met.

roypat:
> Thanks for putting in a fix! However, I don't think this fixes the issue entirely. For example, running fstat on a vfio device returns 0 for the size which makes virtiofsd return MmapRegion(MappingPastEof).

Just to clarify - the PR that closed this issue was replacing fstat with seek exactly to avoid issues like this where fstat returns a size of zero. Are you saying that there's devices where neither seek nor fstat return anything meaningful?

> Is there a reason in particular this check is needed at all? mmap will verify all of this and return failure if any condition is not met.

I was thinking the same the other day admittedly. I don't think this check has too much value. I tried patching over some other rough edges it has in #320, but I'd also be happy to accept a PR that eliminates `check_file_offset` altogether.

GZibrat:
> Just to clarify - the PR that closed this issue was replacing fstat with seek exactly to avoid issues like this where fstat returns a size of zero. Are you saying that there's devices where neither seek nor fstat return anything meaningful?

Sorry! I got confused by the order of things here. But yes, vfio devices do not support seek or stat (returns 0 size). I tried the seek implementation that is in now, this [commit](https://github.com/rust-vmm/vm-memory/pull/320/commits/cbced4a2c833d9da029986c494f3e335169d3dec), and a simple c program that calls fstat. None seemed to provide a valid size.

[...]

PR #196, opened 2022-05-13
mmap: get file size with seek instead of metadata

slp:
Currently mmap::check_file_offset attempts to validate the backing
object by checking, among other things, its size to confirm the
mapping will fit on it. This check is implemented by using
std::fs::metadata to obtain the file size, but this approach doesn't
work when the backing object is a block or char device, as the
returned value is always zero.

In this commit, we replace the use of std::fs:metadata with
std::io::Seek::seek(SeekEnd(0)), which works fine even with special
files such as block and char devices. As a reference, this is the
method that QEMU uses for the same goal.

Fixes: #195
Signed-off-by: Sergio Lopez <slp@redhat.com>

### DIFF
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -16,6 +16,8 @@ use std::borrow::Borrow;
 use std::error;
 use std::fmt;
 use std::io::{Read, Write};
+#[cfg(unix)]
+use std::io::{Seek, SeekFrom};
 use std::ops::Deref;
 use std::result;
 use std::sync::atomic::Ordering;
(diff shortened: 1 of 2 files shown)


rbradford (approved):
Yes, thanks @slp. This is the correct approach and we use this technique elsewhere.

src/mmap.rs
```diff
@@ -116,14 +118,17 @@ pub fn check_file_offset(
...
 
     if let Some(end) = start.checked_add(size as u64) {
-        if let Ok(metadata) = file.metadata() {
-            if metadata.len() < end {
-                return Err(MmapRegionError::MappingPastEof);
-            }
+        let filesize = file
+            .seek(SeekFrom::End(0))
+            .map_err(MmapRegionError::SeekEnd)?;
+        file.seek(SeekFrom::Start(0))
```
alexandruag, line 128: Hmm, do we actually need to reset the position by going back to the start?

slp: It's not strictly needed, since the next thing we'll do on the file is create the mapping. But I thought it was more `future-proof` if we returned the fd it to the expected state. I'm fine dropping it either.

alexandruag (approved):

PR #200, opened 2022-08-05
docs: enable all features in docs.rs

songzhi:


andreeaflorescu (approved):

Cargo.toml
```diff
@@ -37,3 +37,6 @@ harness = false
 [profile.bench]
 lto = true
 codegen-units = 1
+
+[package.metadata.docs.rs]
+all-features = true
```
lauralt, line 42: What is this change supposed to do? Update the generated documentation to show the modules under a feature or? I tried a couple of things and didn't see any difference.

songzhi: This change is for docs.rs, for local builds, `cargo doc --all-features` would generate the same results.

lauralt: `cargo doc --all-features --open` is generating the modules under a feature as well (e.g. mmap), but I assume with your change, plain `cargo doc --open` should do the same, right? It doesn't look like that (at least on my machine), so that's why I'm asking.

songzhi: No, this change is just for [vm-memory docs](https://docs.rs/vm-memory/latest/vm_memory/) in docs.rs, see `[package.metadata.docs.rs]` in the Cargo.toml. Otherwise this docs aren't friendly enough for new lib users.

lauralt (approved):

PR #202, opened 2022-08-12
use caret requirements for dependencies & prepare 0.9.0 release

### LINKED ISSUE #199: Proposal: Improve Cargo dependency format
The crates in rust-vmm use `>=` [comparison-requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#comparison-requirements) to specify specific dependencies. This places a significant maintenance burden on the user who uses it.

## Problem
We use `vm-memory` as an example because the structure of `vm-memory` runs through the vmm ecosystem and cannot use two different versions. See [version-incompatibility-hazards](https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/resolver.md#version-incompatibility-hazards) for reasons.

1. When I develop a library `my-lib 0.1.0`, it depends on `linux-loader` and `vm-memory`, and `linux-loader` depends on `vm-memory >= 0.6`. The version of `my-lib` that depends on `vm-memory` requires:
  * If use default [Caret requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements), then you can only use the latest `0.8.0`, not older versions.  
  * Also use >=, Finally, it will also use the latest 0.8.0.
2. When an [incompatible](https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/semver.md) version of `vm-memory 0.9.0` is released, it will be a disaster. Since cargo's strategy is to use the greatest version. Then, all libraries that depend on `vm-memory >=` must be fixed and released a new version. And the previous `my-lib 0.1.0` will not work. This break is silent.
3. If we don't want to do an update. Can lock the old version for binary crates. But for libraries, we need to use some hacks. e.g. `cargo update -p vm-memory:0.9.0 --precise 0.8.0`. And we have to lock the `linux-loader` as well.  But when my-lib is released, it still can't work.

## Solution

According to cargo [recommendations](https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/resolver.md#recommendations):

Use caret requirements for dependencies, such as `1.2.3`, for most situations. This ensures that the resolver can be maximally flexible in choosing a version while maintaining build compatibility.

And most open sources also use the SemVer version requirement(`x.y.z`), which is equivalent to Caret requirements(`^x.y.z`).
Mentioned in document [specifying-dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies):

> It actually specifies a range of versions and allows SemVer compatible updates.
> An update is allowed if the new version number does not modify the left-most non-zero digit in the major, minor, patch grouping.

The SemVer version requirement can solve the problems mentioned earlier.
1. We can specify any version, as they all have a corresponding compatibility range. e.g. `linux-loader 0.5.x` corresponds to `vm-memory 0.9.x`, x upgrade does not break compatibility either.
2. Incompatible versions are released, and will not affect other crates.
3. If you do not upgrade the dependencies, it will also compile normally.

> NOTE:
> An incompatible version means **the left-most non-zero digit** in x.y.z grouping is different.

## Drawbacks
1. When an incompatible version is released, the rust-vmm crate that depends on this crate needs to provide a new version. So that users can choose when they need to upgrade.
2. When an incompatible version is released, it may be necessary to create a corresponding branch, in order to prepare for a patch release later.
  You can also not create it, then the incompatible version needs to be released immediately after the API incompatible change is merged.

## Required Work
1. Version releases need to be strictly follow [SemVer Compatibility](https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/semver.md) to determine API compatibility, and release the corresponding version number.
2. Modify all rust-vmm crate dependencies to SemVer version requirements. and release a new incompatible version.
3. [Option] Create branches for incompatible versions.

## Progress
image
- [ ] update vmm-sys-util.
- [ ] update vm-memory, kvm-bindings.
- [ ] update linux-loader, virtio-queue, kvm-ioctls.
- [ ] update others.

Thanks.

lauralt:
### Summary of the PR

Updated the deps to use caret requirements.
After merging this, we can have a new release of vm-memory.
Fixes #199 

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,17 +18,17 @@ backend-mmap = []
 backend-atomic = ["arc-swap"]
 
 [dependencies]
-libc = ">=0.2.39"
-arc-swap = { version = ">=1.0.0", optional = true }
+libc = "0.2.39"
+arc-swap = { version = "1.0.0", optional = true }
 
 [target.'cfg(windows)'.dependencies.winapi]
-version = ">=0.3"
+version = "0.3"
 features = ["errhandlingapi", "sysinfoapi"]
 
 [dev-dependencies]
 criterion = "0.3.0"
-matches = ">=0"
-vmm-sys-util = ">= 0.4.0"
+matches = "0.1.0"
+vmm-sys-util = "0.10.0"
 
 [[bench]]
 name = "main"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "vm-memory"
-version = "0.8.0"
+version = "0.9.0"
 description = "Safe abstractions for accessing the VM physical memory"
 keywords = ["memory"]
 categories = ["memory-management"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,24 @@
 # Changelog
 ## [Unreleased]
 
+## [v0.9.0]
+
+### Fixed
+
+- [[#195]](https://github.com/rust-vmm/vm-memory/issues/195):
+  `mmap::check_file_offset` is doing the correct size validation for block and
+  char devices as well.
+
+### Changed
+
+- [[#198]](https://github.com/rust-vmm/vm-memory/pull/198): atomic: enable 64
+  bit atomics on ppc64le and s390x.
+- [[#200]](https://github.com/rust-vmm/vm-memory/pull/200): docs: enable all
+  features in `docs.rs`.
+- [[#199]](https://github.com/rust-vmm/vm-memory/issues/199): Update the way
+  the dependencies are pulled such that we don't end up with incompatible
+  versions.
+
 ## [v0.8.0]
 
 ### Fixed


wllenyj:
Thanks for your work.
I was thinking of waiting for a new version of `vmm-sys-util` to be released before updating vm-memory. But it doesn't seem to be a problem now. 
`vmm-sys-util` and `vm-memory` have the same dependency `libc`, and their version can be covered.

rbradford (approved):

Cargo.toml
```diff
@@ -18,17 +18,17 @@ backend-mmap = []
...
 [target.'cfg(windows)'.dependencies.winapi]
-version = ">=0.3"
+version = "0.3"
 features = ["errhandlingapi", "sysinfoapi"]
 
 [dev-dependencies]
 criterion = "0.3.0"
-matches = ">=0"
-vmm-sys-util = ">= 0.4.0"
+matches = "0"
```
andreeaflorescu, line 30: This dependency will be updated up until < 1.0. If we want to only update as specified in the commit message, we should add this as `0.1` instead.

andreeaflorescu (dismissed):

Cargo.toml
lauralt: Good catch! Fixed it.

andreeaflorescu (approved):

jiangliu (approved):

Issue #194, opened 2022-04-05
Implement PartialEq for errors in vm-memory


andreeaflorescu:
The lack of PartialEq in vm-memory is forcing weird patterns across all rust-vmm crates (where errors need to be compared using `match` or even worse where errors are not compared at all). The problem with implementing `PartialEq` resides in the underlying io::Error which does not implement `PartialEq`. Since the use case for errors is pretty well defined in vm-memory, we should be able to manually implement `PartialEq` because we can reason about what errors should actually be equal.

PR #206, opened 2022-10-10
Partialeq

AlexandruCihodaru:
### Summary of the PR

Implement PartialEq for Error enums.

Fixes: https://github.com/rust-vmm/vm-memory/issues/194

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -73,6 +73,33 @@ pub enum Error {
     HostAddressNotAvailable,
 }
 
+impl PartialEq for Error {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Error::InvalidGuestAddress(left), Error::InvalidGuestAddress(right)) => left == right,
+            (Error::InvalidBackendAddress, Error::InvalidBackendAddress) => true,
+            (Error::HostAddressNotAvailable, Error::HostAddressNotAvailable) => true,
+            (
+                Error::PartialBuffer {
+                    expected: left_expected,
+                    completed: left_completed,
+                },
+                Error::PartialBuffer {
+                    expected: right_expected,
+                    completed: right_completed,
+                },
+            ) => left_expected == right_expected && left_completed == right_completed,
+            (Error::IOError(left), Error::IOError(right)) => {
+                // error.kind should be enough to assert equallity because each error
+                // has the kind field set and the OS error numbers can be converted
+                // to ErrorKind through `sys::decode_error_kind`.
+                left.kind() == right.kind()
+            }
+            _ => false,
+        }
+    }
+}
+
 impl From<volatile_memory::Error> for Error {
     fn from(e: volatile_memory::Error) -> Self {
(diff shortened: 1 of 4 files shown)


src/guest_memory.rs
```diff
@@ -73,6 +73,33 @@ pub enum Error {
...
+                    expected: left_expected,
+                    completed: left_completed,
+                },
+                Error::PartialBuffer {
+                    expected: right_expected,
+                    completed: right_completed,
+                },
+            ) => left_expected == right_expected && left_completed == right_completed,
+            (Error::IOError(left), Error::IOError(right)) => {
+                // error.kind should be enough to assert equallity because each error
```
andreeaflorescu, line 93: While this is true, it does not mean that `kind` can be used to compare errors. The reason why io::Error does not have PartialEq implemented is because it cannot be implemented truthfully. The first example that comes to mind is `NotFound`. A not found error thrown for `file1` is not equal to an error thrown for `file2`. This is one of the reasons why we cannot actually compare the `io::Error` like that because it has other metadata that is specific to the error type. We can say instead that we are willing to accept the loss of information in comparing errors because vm-memory errors are widely used in other crates as well and it is very limiting to not have `PartialEq` implemented.

src/guest_memory.rs
```diff
@@ -73,6 +73,33 @@ pub enum Error {
     HostAddressNotAvailable,
 }
 
+impl PartialEq for Error {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
```
andreeaflorescu, line 78: One reason I don't particularly like these kinds of implementations for PartialEq is that it will silently fail when you add a new error type if you don't explicitly add it in the below match as well. I would propose we add a test that just fails when a new variant is added and has a comment something like: "If this failed it means you added a new error type. Make sure you also add that to the `PartialEq` implementation. Once you've done that feel free to fix this test". This will serve as a warning when adding new types. I am also open to other ideas as this is rather dummy 😆

bonzini:
Given the issues that @andreeaflorescu mentioned, I'd rather add an `assert_matches!` macro and use it to improve the tests.

roypat:
> Given the issues that @andreeaflorescu mentioned, I'd rather add an `assert_matches!` macro and use it to improve the tests.

Agreed.

We've actually replaced all the `assert_eq!(error.to_string(), "string representation of error")` things with `assert!(matches!())` over the last few months. And we actually already have a dev-dependency on the `matches` crate, which provides an `assert_matches!` macro (and we even use it in a few places), soo.... https://github.com/rust-vmm/vm-memory/pull/336.

As for all the assertions on .is_err()/is_ok(), I suspect a lot of them we dont really care about matching the actual error, and just replacing it with .unwrap()/.unwrap_err() is fine (and produces significantly better error messages in case of failure). There's a clippy lint against these as well (clippy::assertion_on_result_states) - I'll open an issue for converting these, in case someone is really bored one day

PR #212, opened 2023-01-02
Improve unsafe code

andreeaflorescu:
### Summary of the PR

- Added some more comments about the unsafe code in volatile_slice.
- Added a negative test to check the behavior of `read_from` when the size of the read is bigger than the size of the reader.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/volatile_memory.rs
+++ b/src/volatile_memory.rs
@@ -1375,6 +1375,8 @@ mod tests {
     use super::*;
 
     use std::fs::File;
+    use std::io::Cursor;
+    use std::mem;
     use std::mem::size_of_val;
     use std::path::Path;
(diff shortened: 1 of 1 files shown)


bonzini (approved):
Nice! The test should have been there from the beginning in particular.

rbradford (approved):
Why can't we use `#[must_use]` ?

andreeaflorescu:
> Why can't we use `#[must_use]` ?

The `must_use` will apply to the `Result`, and `Result` is already `must_use` in Rust. There is no way that I could find by which to specify that the inner type in a Result is a must_use. The only alternative that I could think of is changing the return type to a custom result type, but that comes with a breaking change and I didn't find it to be worth it. Do you know about an alternative to this? I might be missing something.

PR #223, opened 2023-04-17
fix: `thiserror::Error`

JonathanWoollett-Light:
### Summary of the PR

Implements the`thiserror::Error` derive macro.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -42,7 +42,6 @@
 //! `VolatileSlice`, `VolatileRef`, or `VolatileArrayRef`.
 
 use std::convert::From;
-use std::fmt::{self, Display};
 use std::fs::File;
 use std::io::{self, Read, Write};
 use std::ops::{BitAnd, BitOr, Deref};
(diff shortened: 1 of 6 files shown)


src/guest_memory.rs
```diff
@@ -95,30 +99,6 @@ impl From<volatile_memory::Error> for Error {
 /// Result of guest memory operations.
 pub type Result<T> = std::result::Result<T, Error>;
 
-impl std::error::Error for Error {}
-
-impl Display for Error {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Guest memory error: ")?;
```
roypat, line 102: This prefix seems to be missing with the `thiserror` impls

roypat (commented):
Looks good. We already use this dependency in a lot of other rust-vmm crates, so no concerns there [[1]].

[1]: https://github.com/search?q=org%3Arust-vmm%20thiserror&type=code

src/guest_memory.rs
JonathanWoollett-Light: This is unfortunate design, but you are right, added the prefix to be identical.

roypat (approved):

andreeaflorescu (approved):

PR #224, opened 2023-04-18
Use NonNull<u8> in MmapRegionBuilder for saving space

FuuuOverclocking:
### Summary of the PR

Updates the `raw_ptr` field in `MmapRegionBuilder` (mmap_unix.rs) from `Option<*mut u8>` to `Option<NonNull<u8>>`. This modification saves 8 bytes on x86-64 systems, for example.

The public API and functionality remain unchanged.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/mmap_unix.rs
+++ b/src/mmap_unix.rs
@@ -12,7 +12,7 @@
 
 use std::io;
 use std::os::unix::io::AsRawFd;
-use std::ptr::null_mut;
+use std::ptr::{null_mut, NonNull};
 use std::result;
 
 use crate::bitmap::{Bitmap, BS};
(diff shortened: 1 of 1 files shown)


FuuuOverclocking:
This commit caused the test to fail due to `fn test_mmap_region_build_raw()` (vm_memory::mmap_unix::tests) as `addr` is 0 in this test. I'm wondering, is `addr` being 0 a valid parameter?

andreeaflorescu:
@FuuuOverclocking what is the reason for this PR? 8 bytes in the great scheme of things does not feel like a lot.

FuuuOverclocking:
In Rust, using `Option<NonNull<u8>>` is a common pattern when both `Option` and pointer are needed.

epilys:
@FuuuOverclocking Hello, I think this PR should also wrap `MmapRegion::addr` in `NonNull<_>` as well. Would you be interested in rebasing and making this change?

jiangliu (approved):

roypat:
This seems reasonable to me, I don't think there's any use case in mapping in the zero page, and `Option<NonNull<u8>>` is indeed the more rust-y approach. 

I also agree with @epilys that if we do this, we should probably be consistent with it, and use `NonNull<u8>` everywhere (but in this case, this probably will need a changelog entry). 

Also, it currently seems like unittests are failing, could you look into that? :o

PR #231, opened 2023-05-19
Xen/mmapv2

vireshk:
Xen allows two memory mapping mechanisms:
- Foreign mapping: entire guest space is mapped at once.
- Grant mapping: guest controls what backend is allowed to map.

Both these mechanisms required additional mapping support, for example
an ioctl() needs to be issued (with arguments passed from backend),
along with mmap().

Also for grant memory mapping, the mapping can't be done in advance and
is done on the fly and so the changes to VolatileSlice.

Add memory mapping support for both these cases, under a new feature
"xen".

Signed-off-by: Viresh Kumar <viresh.kumar@linaro.org>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -54,7 +54,7 @@ mod mmap_windows;
 #[cfg(feature = "backend-mmap")]
 pub mod mmap;
 #[cfg(feature = "backend-mmap")]
-pub use mmap::{Error, GuestMemoryMmap, GuestRegionMmap, MmapRegion};
+pub use mmap::{Error, GuestMemoryMmap, GuestMmapRange, GuestRegionMmap, MmapRange, MmapRegion};
 
 pub mod volatile_memory;
 pub use volatile_memory::{
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,6 +48,21 @@ pub use guest_memory::{
 #[cfg(all(feature = "backend-mmap", unix))]
 mod mmap_unix;
 
+#[cfg(all(feature = "xen", unix))]
+mod mmap_xen;
+
+#[cfg(all(feature = "backend-mmap", feature = "xen", unix))]
+pub use crate::mmap_xen::MmapXenFlags;
+
+#[cfg(all(feature = "backend-mmap", feature = "xen", unix))]
+use crate::mmap_xen::{MmapXen as MmapInfo, MmapXenSlice as MmapSlice};
+
+#[cfg(all(feature = "backend-mmap", not(feature = "xen"), unix))]
+use crate::mmap_unix::Mmap as MmapInfo;
+
+#[cfg(any(not(feature = "backend-mmap"), not(unix)))]
+type MmapInfo = std::marker::PhantomData<()>;
+
 #[cfg(all(feature = "backend-mmap", windows))]
 mod mmap_windows;
 
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -950,7 +950,7 @@ mod tests {
         let m =
             GuestRegionMmap::from_range(&GuestMmapRange::new(GuestAddress(0), 5, None)).unwrap();
         let s = m.get_slice(MemoryRegionAddress(2), 3).unwrap();
-        assert_eq!(s.as_ptr(), unsafe { m.as_ptr().offset(2) });
+        assert_eq!(s.ptr_guard().as_ptr(), unsafe { m.as_ptr().offset(2) });
     }
 
     #[test]
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -105,12 +105,47 @@ pub struct GuestMmapRange {
     pub(crate) size: usize,
     // File offset
     pub(crate) file: Option<FileOffset>,
+
+    // Xen mmap flags
+    #[cfg(all(feature = "xen", unix))]
+    pub(crate) mmap_flags: u32,
+    // Xen mmap data
+    #[cfg(all(feature = "xen", unix))]
+    pub(crate) mmap_data: u32,
 }
 
 impl GuestMmapRange {
     /// Creates a new instance of `GuestMmapRange`
     pub fn new(addr: GuestAddress, size: usize, file: Option<FileOffset>) -> Self {
-        Self { addr, size, file }
+        Self {
+            addr,
+            size,
+            file,
+
+            // Defaults to standard UNIX mapping. Use `with_xen()` for Xen specific mappings.
+            #[cfg(all(feature = "xen", unix))]
+            mmap_flags: super::MmapXenFlags::UNIX.bits(),
+            #[cfg(all(feature = "xen", unix))]
+            mmap_data: 0,
+        }
+    }
+
+    #[cfg(all(feature = "xen", unix))]
+    /// Creates a new instance of `GuestMmapRange` for Xen mappings.
+    pub fn with_xen(
+        addr: GuestAddress,
+        size: usize,
+        file: Option<FileOffset>,
+        mmap_flags: u32,
+        mmap_data: u32,
+    ) -> Self {
+        Self {
+            addr,
+            size,
+            file,
+            mmap_flags,
+            mmap_data,
+        }
     }
 }
 
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -12,7 +12,6 @@
 //!
 //! This implementation is mmap-ing the memory of the guest into the current process.
 
-use std::borrow::Borrow;
 use std::io::{Read, Write};
 #[cfg(unix)]
 use std::io::{Seek, SeekFrom};
@@ -31,10 +30,10 @@ use crate::volatile_memory::{VolatileMemory, VolatileSlice};
 use crate::{AtomicAccess, Bytes};
 
 #[cfg(unix)]
-pub use crate::mmap_unix::{Error as MmapRegionError, MmapRegion, MmapRegionBuilder};
+pub use crate::mmap_unix::{Error as MmapRegionError, MmapRange, MmapRegion, MmapRegionBuilder};
 
 #[cfg(windows)]
-pub use crate::mmap_windows::MmapRegion;
+pub use crate::mmap_windows::{MmapRange, MmapRegion};
 #[cfg(windows)]
 pub use std::io::Error as MmapRegionError;
 
@@ -97,6 +96,34 @@ pub fn check_file_offset(
     Ok(())
 }
 
+/// `GuestMmapRange` represents a range of arguments required to create Mmap regions.
+#[derive(Clone, Debug)]
+pub struct GuestMmapRange {
+    // Guest address
+    pub(crate) addr: GuestAddress,
+    // Region's size
+    pub(crate) size: usize,
+    // File offset
+    pub(crate) file: Option<FileOffset>,
+}
+
+impl GuestMmapRange {
+    /// Creates a new instance of `GuestMmapRange`
+    pub fn new(addr: GuestAddress, size: usize, file: Option<FileOffset>) -> Self {
+        Self { addr, size, file }
+    }
+}
+
+impl PartialEq for GuestMmapRange {
+    fn eq(&self, other: &Self) -> bool {
+        // Matches only None value of the file offsets for now.
+        self.addr == other.addr
+            && self.size == other.size
+            && self.file.is_none()
+            && other.file.is_none()
+    }
+}
+
 /// [`GuestMemoryRegion`](trait.GuestMemoryRegion.html) implementation that mmaps the guest's
 /// memory region in the current process.
 ///
@@ -131,20 +158,10 @@ impl<B: Bitmap> GuestRegionMmap<B> {
 }
 
 impl<B: NewBitmap> GuestRegionMmap<B> {
-    /// Create a new memory-mapped memory region from guest's physical memory, size and file.
-    pub fn from_range(
-        addr: GuestAddress,
-        size: usize,
-        file: Option<FileOffset>,
-    ) -> result::Result<Self, Error> {
-        let region = if let Some(ref f_off) = file {
-            MmapRegion::from_file(f_off.clone(), size)
-        } else {
-            MmapRegion::new(size)
-        }
-        .map_err(Error::MmapRegion)?;
-
-        Self::new(region, addr)
+    /// Create a new memory-mapped memory region from guest's `GuestMmapRange`.
+    pub fn from_range(r: &GuestMmapRange) -> result::Result<Self, Error> {
+        let region = MmapRegion::new(r.clone().into()).map_err(Error::MmapRegion)?;
+        Self::new(region, r.addr)
     }
 }
 
@@ -155,10 +172,11 @@ impl<B: Bitmap> Bytes<MemoryRegionAddress> for GuestRegionMmap<B> {
     /// * Write a slice at guest address 0x1200.
     ///
     /// ```
-    /// # use vm_memory::{Bytes, GuestAddress, GuestMemoryMmap};
+    /// # use vm_memory::{Bytes, GuestAddress, GuestMemoryMmap, GuestMmapRange};
     /// #
     /// # let start_addr = GuestAddress(0x1000);
-    /// # let mut gm = GuestMemoryMmap::<()>::from_ranges(&vec![(start_addr, 0x400)])
+    /// # let mut gm =
+    /// #    GuestMemoryMmap::<()>::from_ranges(&vec![GuestMmapRange::new(start_addr, 0x400, None)])
     /// #    .expect("Could not create guest memory");
     /// #
     /// let res = gm
@@ -178,10 +196,11 @@ impl<B: Bitmap> Bytes<MemoryRegionAddress> for GuestRegionMmap<B> {
     /// * Read a slice of length 16 at guestaddress 0x1200.
     ///
     /// ```
-    /// # use vm_memory::{Bytes, GuestAddress, GuestMemoryMmap};
+    /// # use vm_memory::{Bytes, GuestAddress, GuestMemoryMmap, GuestMmapRange};
     /// #
     /// # let start_addr = GuestAddress(0x1000);
-    /// # let mut gm = GuestMemoryMmap::<()>::from_ranges(&vec![(start_addr, 0x400)])
+    /// # let mut gm =
+    /// #    GuestMemoryMmap::<()>::from_ranges(&vec![GuestMmapRange::new(start_addr, 0x400, None)])
     /// #    .expect("Could not create guest memory");
     /// #
     /// let buf = &mut [0u8; 16];
@@ -219,12 +238,13 @@ impl<B: Bitmap> Bytes<MemoryRegionAddress> for GuestRegionMmap<B> {
     /// * Read bytes from /dev/urandom
     ///
     /// ```
-    /// # use vm_memory::{Address, Bytes, GuestAddress, GuestMemoryMmap};
+    /// # use vm_memory::{Address, Bytes, GuestAddress, GuestMemoryMmap, GuestMmapRange};
     /// # use std::fs::File;
     /// # use std::path::Path;
     /// #
     /// # let start_addr = GuestAddress(0x1000);
-    /// # let gm = GuestMemoryMmap::<()>::from_ranges(&vec![(start_addr, 0x400)])
+    /// # let gm =
+    /// #    GuestMemoryMmap::<()>::from_ranges(&vec![GuestMmapRange::new(start_addr, 0x400, None)])
     /// #    .expect("Could not create guest memory");
     /// # let addr = GuestAddress(0x1010);
     /// # let mut file = if cfg!(unix) {
@@ -264,12 +284,13 @@ impl<B: Bitmap> Bytes<MemoryRegionAddress> for GuestRegionMmap<B> {
     /// * Read bytes from /dev/urandom
     ///
     /// ```
-    /// # use vm_memory::{Address, Bytes, GuestAddress, GuestMemoryMmap};
+    /// # use vm_memory::{Address, Bytes, GuestAddress, GuestMemoryMmap, GuestMmapRange};
     /// # use std::fs::File;
     /// # use std::path::Path;
     /// #
     /// # let start_addr = GuestAddress(0x1000);
-    /// # let gm = GuestMemoryMmap::<()>::from_ranges(&vec![(start_addr, 0x400)])
+    /// # let gm =
+    /// #    GuestMemoryMmap::<()>::from_ranges(&vec![GuestMmapRange::new(start_addr, 0x400, None)])
     /// #    .expect("Could not create guest memory");
     /// # let addr = GuestAddress(0x1010);
     /// # let mut file = if cfg!(unix) {
@@ -313,10 +334,11 @@ impl<B: Bitmap> Bytes<MemoryRegionAddress> for GuestRegionMmap<B> {
     /// ```
     /// # #[cfg(not(unix))]
     /// # extern crate vmm_sys_util;
-    /// # use vm_memory::{Address, Bytes, GuestAddress, GuestMemoryMmap};
+    /// # use vm_memory::{Address, Bytes, GuestAddress, GuestMemoryMmap, GuestMmapRange};
     /// #
     /// # let start_addr = GuestAddress(0x1000);
-    /// # let gm = GuestMemoryMmap::<()>::from_ranges(&vec![(start_addr, 0x400)])
+    /// # let gm =
+    /// #    GuestMemoryMmap::<()>::from_ranges(&vec![GuestMmapRange::new(start_addr, 0x400, None)])
     /// #    .expect("Could not create guest memory");
     /// # let mut file = if cfg!(unix) {
     /// # use std::fs::OpenOptions;
@@ -358,10 +380,11 @@ impl<B: Bitmap> Bytes<MemoryRegionAddress> for GuestRegionMmap<B> {
     /// ```
     /// # #[cfg(not(unix))]
     /// # extern crate vmm_sys_util;
-    /// # use vm_memory::{Address, Bytes, GuestAddress, GuestMemoryMmap};
+    /// # use vm_memory::{Address, Bytes, GuestAddress, GuestMemoryMmap, GuestMmapRange};
     /// #
     /// # let start_addr = GuestAddress(0x1000);
-    /// # let gm = GuestMemoryMmap::<()>::from_ranges(&vec![(start_addr, 0x400)])
+    /// # let gm =
+    /// #    GuestMemoryMmap::<()>::from_ranges(&vec![GuestMmapRange::new(start_addr, 0x400, None)])
     /// #    .expect("Could not create guest memory");
     /// # let mut file = if cfg!(unix) {
     /// # use std::fs::OpenOptions;
@@ -481,26 +504,12 @@ impl<B: NewBitmap> GuestMemoryMmap<B> {
 
     /// Creates a container and allocates anonymous memory for guest memory regions.
     ///
-    /// Valid memory regions are specified as a slice of (Address, Size) tuples sorted by Address.
-    pub fn from_ranges(ranges: &[(GuestAddress, usize)]) -> result::Result<Self, Error> {
-        Self::from_ranges_with_files(ranges.iter().map(|r| (r.0, r.1, None)))
-    }
-
-    /// Creates a container and allocates anonymous memory for guest memory regions.
-    ///
-    /// Valid memory regions are specified as a sequence of (Address, Size, Option<FileOffset>)
-    /// tuples sorted by Address.
-    pub fn from_ranges_with_files<A, T>(ranges: T) -> result::Result<Self, Error>
-    where
-        A: Borrow<(GuestAddress, usize, Option<FileOffset>)>,
-        T: IntoIterator<Item = A>,
-    {
+    /// Valid memory regions are specified as a slice of `GuestMmapRange` items sorted by Address.
+    pub fn from_ranges(ranges: &[GuestMmapRange]) -> result::Result<Self, Error> {
         Self::from_regions(
             ranges
-                .into_iter()
-                .map(|x| {
-                    GuestRegionMmap::from_range(x.borrow().0, x.borrow().1, x.borrow().2.clone())
-                })
+                .iter()
+                .map(GuestRegionMmap::from_range)
                 .collect::<result::Result<Vec<_>, Error>>()?,
         )
     }
@@ -651,7 +660,7 @@ mod tests {
 
     #[test]
     fn basic_map() {
-        let m = MmapRegion::new(1024).unwrap();
+        let m = MmapRegion::new(MmapRange::new(1024)).unwrap();
         assert_eq!(1024, m.size());
     }
 
@@ -685,7 +694,12 @@ mod tests {
     fn new_guest_memory_mmap(
         regions_summary: &[(GuestAddress, usize)],
     ) -> Result<GuestMemoryMmap, Error> {
-        GuestMemoryMmap::from_ranges(regions_summary)
+        let ranges: Vec<GuestMmapRange> = regions_summary
+            .iter()
+            .map(|(region_addr, region_size)| GuestMmapRange::new(*region_addr, *region_size, None))
+            .collect();
+
+        GuestMemoryMmap::from_ranges(&ranges)
     }
 
     fn new_guest_memory_mmap_from_regions(
@@ -695,7 +709,12 @@ mod tests {
             regions_summary
                 .iter()
                 .map(|(region_addr, region_size)| {
-                    GuestRegionMmap::from_range(*region_addr, *region_size, None).unwrap()
+                    GuestRegionMmap::from_range(&GuestMmapRange::new(
+                        *region_addr,
+                        *region_size,
+                        None,
+                    ))
+                    .unwrap()
                 })
                 .collect(),
         )
@@ -708,7 +727,14 @@ mod tests {
             regions_summary
                 .iter()
                 .map(|(region_addr, region_size)| {
-                    Arc::new(GuestRegionMmap::from_range(*region_addr, *region_size, None).unwrap())
+                    Arc::new(
+                        GuestRegionMmap::from_range(&GuestMmapRange::new(
+                            *region_addr,
+                            *region_size,
+                            None,
+                        ))
+                        .unwrap(),
+                    )
                 })
                 .collect(),
         )
@@ -717,17 +743,17 @@ mod tests {
     fn new_guest_memory_mmap_with_files(
         regions_summary: &[(GuestAddress, usize)],
     ) -> Result<GuestMemoryMmap, Error> {
-        let regions: Vec<(GuestAddress, usize, Option<FileOffset>)> = regions_summary
+        let regions: Vec<GuestMmapRange> = regions_summary
             .iter()
             .map(|(region_addr, region_size)| {
                 let f = TempFile::new().unwrap().into_file();
                 f.set_len(*region_size as u64).unwrap();
 
-                (*region_addr, *region_size, Some(FileOffset::new(f, 0)))
+                GuestMmapRange::new(*region_addr, *region_size, Some(FileOffset::new(f, 0)))
             })
             .collect();
 
-        GuestMemoryMmap::from_ranges_with_files(&regions)
+        GuestMemoryMmap::from_ranges(&regions)
     }
 
     #[test]
@@ -886,7 +912,8 @@ mod tests {
 
     #[test]
     fn slice_addr() {
-        let m = GuestRegionMmap::from_range(GuestAddress(0), 5, None).unwrap();
+        let m =
+            GuestRegionMmap::from_range(&GuestMmapRange::new(GuestAddress(0), 5, None)).unwrap();
         let s = m.get_slice(MemoryRegionAddress(2), 3).unwrap();
         assert_eq!(s.as_ptr(), unsafe { m.as_ptr().offset(2) });
     }
@@ -899,7 +926,12 @@ mod tests {
         assert!(f.write_all(sample_buf).is_ok());
 
         let file = Some(FileOffset::new(f, 0));
-        let mem_map = GuestRegionMmap::from_range(GuestAddress(0), sample_buf.len(), file).unwrap();
+        let mem_map = GuestRegionMmap::from_range(&GuestMmapRange::new(
+            GuestAddress(0),
+            sample_buf.len(),
+            file,
+        ))
+        .unwrap();
         let buf = &mut [0u8; 16];
         assert_eq!(
             mem_map.as_volatile_slice().unwrap().read(buf, 0).unwrap(),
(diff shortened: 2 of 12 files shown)


vireshk:
Doc tests fail currently as following doesn't work and some definitions aren't found.

#[cfg(all(any(test, doctest), feature = "xen"))]

Will it be possible to get some review on this now ? @andreeaflorescu

src/mmap.rs
```diff
@@ -525,6 +560,34 @@ impl<B: NewBitmap> GuestMemoryMmap<B> {
...
 
+#[cfg(feature = "xen")]
+impl<B: NewBitmap> GuestMemoryMmap<B> {
+    /// Creates a container and allocates anonymous memory for guest memory regions.
+    ///
+    /// Valid memory regions are specified as a sequence of (Address, Size, Option<FileOffset>)
+    /// tuples sorted by Address.
+    pub fn from_ranges_with_files<A, T>(ranges: T) -> result::Result<Self, Error>
+    where
+        A: Borrow<(GuestAddress, usize, Option<FileOffset>, u32, u32)>,
```
Ablu, line 571: I find the A a bit confusing here. Is it just to be able to spell out the item separately?

If so, it should also be possible to just leave out the `Item` when constraining `T` and add a line: `T::Item: Borrow<...>`.

src/mmap.rs
```diff
@@ -525,6 +560,34 @@ impl<B: NewBitmap> GuestMemoryMmap<B> {
...
+        Self::from_regions(
+            ranges
+                .into_iter()
+                .map(|x| {
+                    GuestRegionMmap::from_range(
+                        x.borrow().0,
+                        x.borrow().1,
+                        x.borrow().2.clone(),
+                        x.borrow().3,
+                        x.borrow().4,
```
Ablu, line 583: Would it maybe make sense to introduce a struct instead of a tuple? I guess rust-analyzer's code lenses help when reading it, but I feel this is still becoming a fairly long list of ints with various meanings.

src/mmap_unix.rs
```diff
@@ -74,6 +74,25 @@ impl error::Error for Error {}
 
 pub type Result<T> = result::Result<T, Error>;
 
+pub(crate) fn mmap_unix(
+    size: usize,
+    prot: i32,
+    flags: i32,
+    fd: i32,
+    f_off: u64,
+) -> Result<*mut u8> {
```
Ablu, line 93: I wonder: If we now add more dynamic mmap behaviour, would it maybe make sense to wrap the result pointer in some kind of guard object that can do automatic unmapping if the object is dropped? My feeling is that this may also provide some options to simplify the `unsafe`-use that will be required to use this return pointer.

Ablu (commented):
Some early comments... Stopped midway when my brain stopped working. Will comment on the rest later :)

src/mmap_xen.rs
```diff
@@ -0,0 +1,607 @@
...
+            (map, refs)
+        };
+
+        map.count = count as u32;
+        map.pad = 0;
+        map.index = 0;
+
+        for (i, r) in refs.iter_mut().enumerate().take(count) {
+            r.domid = domid;
+            r.r#ef = base + i as u32;
```
Ablu, line 302: Where does this `#` come from in `r#ef`?

src/mmap_xen.rs
```diff
@@ -0,0 +1,607 @@
...
+        }
+
+        map
+    }
+}
+
+impl Drop for GntDevMapGrantRef {
+    fn drop(&mut self) {
```
Ablu, line 310: Will anyone ever call this drop? The `new()` function is claiming to return a `&'static`. I do not understand who would trigger the drop to free the alloc that happens in `new()`...

Ablu (commented):
Had another go while waiting for a long build... Still not entirely through...

src/mmap_xen.rs
```diff
@@ -0,0 +1,802 @@
...
+            if ptr.is_null() {
+                handle_alloc_error(layout);
+            }
+
+            let map: &mut Self = &mut *(ptr as *mut Self);
+            let refs: &mut [GntDevGrantRef] =
+                slice::from_raw_parts_mut(ptr.add(size_of::<Self>()) as *mut GntDevGrantRef, count);
+
+            (map, refs)
+        };
```
roypat, line 294: This is triggering undefined behavior: the `ptr` returned by `alloc::alloc` points to uninitialized memory. Then in line 289, you're dereferencing this pointer to get a mutable reference to this uninitialized memory [[1]]. Since you also write through this pointer, it causes further UB as uninitialized memory will get dropped. 

I'm not familiar with the API in question, but this looks a bit like a FAM, so maybe the vmm-sys-utils abstractions for those can be used to make this safe https://github.com/rust-vmm/vmm-sys-util/blob/main/src/fam.rs?

[1]: https://doc.rust-lang.org/reference/behavior-considered-undefined.html

roypat (changes requested):
Hey, I've wanted to give this a look over next week (am really busy the rest of this week sadly), but one thing caught my eye so am just leaving a comment ahead of time

src/volatile_memory.rs
```diff
@@ -1418,6 +1417,28 @@ mod copy_slice_impl {
...
     }
+
+    /// Copies `total` bytes from `slice` to `dst`
+    pub(super) fn copy_from_volatile_slice<B: BitmapSlice>(
+        dst: *mut u8,
+        slice: &VolatileSlice<'_, B>,
+        total: usize,
+    ) -> usize {
+        // SAFETY: `slice` and `dst` must be point to a contiguously allocated memory region of at
+        // least length `total`. The regions must not overlap.
```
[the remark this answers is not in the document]
vireshk: Hmm, I normally add `unsafe` only to routines which directly contain `unsafe` rust code. This makes the users of the routine clean of the `unsafe` mess. I am not sure if this should be marked `unsafe` or not now.

@lauralt @andreeaflorescu any comments ?

src/volatile_memory.rs
```diff
@@ -1291,21 +1296,58 @@ where
...
+                addr = addr.add(1);
+            };
+            i += 1;
+        }
+        i
+    }
 
-            self.bitmap
-                .mark_dirty(0, addr as usize - self.addr as usize)
+    fn write_volatile_aref(&self, buf: &[T], mut addr: *mut Packed<T>, count: usize) {
```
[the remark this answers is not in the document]
vireshk: no clippy doesn't complain at all. clippy has: not_unsafe_ptr_arg_deref, but that is only when you dereference the pointer, which this function doesn't.

src/lib.rs
```diff
@@ -48,6 +48,21 @@ pub use guest_memory::{
 #[cfg(all(feature = "backend-mmap", unix))]
 mod mmap_unix;
 
+#[cfg(all(feature = "xen", unix))]
```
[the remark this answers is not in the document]
vireshk: That was a mistake, the new version will fix it.

src/mmap.rs
```diff
@@ -149,6 +149,7 @@ impl<B: Bitmap> GuestRegionMmap<B> {
     }
 }
 
+#[cfg(not(feature = "xen"))]
```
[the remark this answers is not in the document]
vireshk: Yes, some documentation will be good. Daemons need to have `xen` feature too now and build separately for both. Though I have now added a UNIX mapping model with xen, where it can also do UNIX type mappings, kvm i.e.

src/volatile_memory.rs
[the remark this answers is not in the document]
Ablu: > Hmm, I normally add `unsafe` only to routines which directly contain `unsafe` rust code.

But this allows completely unsafe-free code (the user of the function) to trigger unsafe behavior unless your function prevents any case of unsafe behavior. In this example, the caller of this function is responsible for making sure that the conditions are verified. Declaring the function unsafe makes that explicit and easier to review.

`unsafe` is more than just a way to turn off checks in the compiler. It also is part of the function interface that helps the user write safe code. For example, many of the `unchecked` functions [1] do technically do anything that _requires_ an unsafe, but they are still declared unsafe, because they would otherwise easily allow introducing undefined/unsafe behaviour.

[1] https://doc.rust-lang.org/src/alloc/string.rs.html#845

> This makes the users of the routine clean of the `unsafe` mess.

I agree that things may be messy for functions that are used very commonly in the code. But if an unsafe function is so fundamental that it is used everywhere, then one should rather take it as a hint that this might be a candidate for creating a safe wrapper around it. I am not sure how practical that would be in this case, but one could, for example wrap the pointer with something that also holds an "allowed access range". Then a _safe_ function could check that the access stays within range and then perform the access using an _unsafe_ block without needing to turn the entire function unsafe.

src/volatile_memory.rs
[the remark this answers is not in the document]
Ablu: Ok. I guess clippy does not complain here since it does not see that the ptr is dereferenced. But write_volatile will still dereference it :)

src/mmap.rs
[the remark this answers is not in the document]
Ablu: Ok. I think this may become awkward when packaging stuff, but I will leave this to the maintainers to decide. Code complexity wise I personally do not see a huge difference between conditional compilation and checks at runtime.

src/mmap_xen.rs
```diff
@@ -0,0 +1,607 @@
...
+        Self {
+            grant: None,
+            unmap_addr: null_mut(),
+            addr,
+            size: 0,
+            index: 0,
+        }
+    }
+
+    fn new_with(grant: MmapXenGrant, offset: usize, is_read: bool, size: usize) -> Result<Self> {
```
Ablu, line 506: While doing my hacks, I realized that this `is_read` may be easy to interpret incorrectly. Maybe an enum of `READ_ONLY`, `WRITE_ONLY` would make things more clear?

src/mmap_xen.rs
```diff
@@ -0,0 +1,607 @@
...
+use crate::mmap_unix::{self, Error, Result};
+
+fn page_size() -> u64 {
+    // SAFETY: Safe because this call just returns the page size and doesn't have any side effects.
+    unsafe { libc::sysconf(_SC_PAGESIZE) as u64 }
+}
+
+fn pages(size: usize) -> (usize, usize) {
+    let page_size = page_size() as usize;
+    let num = (size + page_size - 1) / page_size;
```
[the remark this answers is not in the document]
vireshk: Xen allocates memory with granularity of pages, I am just aligning size as a number of pages here.

src/mmap_xen.rs
```diff
@@ -0,0 +1,607 @@
...
+
+    (num, page_size * num)
```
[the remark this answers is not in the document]
vireshk: Are you suggesting this to be returned here ?

`
struct Size {
    size: usize,
    pages: usize, // or page_count
}
`

src/mmap_xen.rs
```diff
@@ -0,0 +1,607 @@
...
+}
+
+fn validate_file(file_offset: &Option<FileOffset>) -> Result<(i32, u64)> {
+    let (fd, f_off) = match file_offset {
```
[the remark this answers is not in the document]
vireshk: `f_off` was already used in the existing code and hence I didn't touch it :)

Updated all instances to `f_offset` now though.

src/mmap_xen.rs
```diff
@@ -0,0 +1,607 @@
...
+    pub fn is_valid(flags: u32) -> bool {
+        // Only of one of FOREIGN or GRANT should be set and mmap_in_advance() should be true with
+        // FOREIGN.
+        if Self::is_foreign(flags) {
+            if Self::is_grant(flags) || !Self::mmap_in_advance(flags) {
+                return false;
+            }
+        } else if !Self::is_grant(flags) {
+            return false;
+        }
```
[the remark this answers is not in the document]
vireshk: I have updated this already as things changed. Please see if you like that one or not.

src/mmap_xen.rs
```diff
@@ -0,0 +1,607 @@
...
+
+        true
+    }
+
+    /// Is xen foreign memory.
+    pub fn is_foreign(flags: u32) -> bool {
+        flags & Self::FOREIGN.bits() != 0
```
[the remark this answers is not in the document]
vireshk: Yeah, that looks good.

src/mmap_xen.rs
vireshk: `ref` is a keyword, had to do this to use the same name. Named it `reference` now.

src/mmap_xen.rs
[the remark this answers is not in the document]
Ablu: I think my brain was just doing weird math... Somehow thought it would sometimes allocate too much.

src/mmap_xen.rs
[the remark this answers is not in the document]
Ablu: I was thinking something like that yes. Though I would maybe call it `PageInfo` or something that has `Page` in the name. But I am not feeling too strongly about this. I guess people will quickly figure it out if they mix up the tuple order :).

src/mmap_xen.rs
vireshk: Fixed, thanks. Good catch.

src/mmap.rs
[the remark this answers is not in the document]
vireshk: I have updated README with this inforamtion now.

src/volatile_memory.rs
```diff
@@ -954,31 +954,49 @@ where
         &self.bitmap
     }
 
-    /// Does a volatile write of the value `v` to the address of this ref.
+    /// Converts this to a [`VolatileSlice`](struct.VolatileSlice.html) with the same size and
+    /// address.
+    pub fn to_slice(&self) -> VolatileSlice<'a, B> {
```
andreeaflorescu, line 981: All the functions that don't check the invariants from the `SAFETY` section need to be marked as unsafe when calling the said functions.

src/mmap_unix.rs
```diff
@@ -51,6 +51,49 @@ pub enum Error {
 
 pub type Result<T> = result::Result<T, Error>;
 
+pub(crate) fn mmap_unix(
```
andreeaflorescu: No need to name this mmap_unix, you can name it mmap, and when you use it in the higher level abstractions you can either use it as mmap_unix::mmap, or do a `use mmap_unix::mmap as mmap_unix`. This way we use the namespace in which the function is defined. Same for `mmap_with_file`.

Why do you need these helpers? Can you explain it in the commit message as well? This is not immediately clear.

src/mmap.rs
```diff
@@ -130,15 +130,16 @@ impl<B: Bitmap> GuestRegionMmap<B> {
...
     /// Create a new memory-mapped memory region from guest's physical memory, size and file.
     pub fn from_range(
         addr: GuestAddress,
         size: usize,
         file: Option<FileOffset>,
     ) -> result::Result<Self, Error> {
-        let region = if let Some(ref f_off) = file {
-            MmapRegion::from_file(f_off.clone(), size)
+        let region = if let Some(ref f_offset) = file {
+            MmapRegion::from_file(f_offset.clone(), size)
```
andreeaflorescu, line 142: This change is not needed, you can leave it now, but in the future it would be great if we limit the number of changes to make the PRs easier to review. This is especially useful when we have PR with many lines of code like this one.

src/mmap.rs
```diff
@@ -506,6 +541,50 @@ impl<B: NewBitmap> GuestMemoryMmap<B> {
...
+                crate::mmap_xen::MmapXenFlags::UNIX.bits(),
+                0,
+            )
+        }))
+    }
+
+    /// Creates a container and allocates anonymous memory for guest memory regions.
+    ///
+    /// Valid memory regions are specified as a sequence of
+    /// (Address, Size, Option<FileOffset>, u32, u32) tuples sorted by Address.
```
andreeaflorescu, line 565: To make the code easier to read and reason about, it would be great to have a structure for these parameters instead.

src/mmap_unix.rs
```diff
@@ -47,50 +51,91 @@ pub enum Error {
...
+            size: 0,
+            owned: false,
+        }
+    }
+
+    pub fn addr(&self) -> *mut u8 {
+        self.addr
+    }
+
+    #[allow(dead_code)]
```
andreeaflorescu, line 119: Why do we need to allow dead_code?

src/mmap.rs
```diff
@@ -728,6 +728,57 @@ mod tests {
...
 
+    #[cfg(not(all(feature = "xen", unix)))]
+    mod defs {
+        use super::*;
+
+        pub fn as_range(
+            addr: GuestAddress,
+            size: usize,
+            file: Option<FileOffset>,
+        ) -> (GuestAddress, usize, Option<FileOffset>) {
```
andreeaflorescu, line 739: We should just have a specialised return type for `as_range` instead. This way we also don't need to duplicate these defines. The type for xen will have the 5 fields, while the other one will have 3. Then the `as_range` function will have the same definition in both cases.

andreeaflorescu (changes requested):
This PR is bigger than I was expecting it to be. There is still quite a lot of duplication needed to accomodate Xen, that I think we can get rid of if we define some structures for configuring the mmap and returning the result, such that both Xen on KVM have the same signature. In some cases, the code is close enough that they can even have the same definition and just use #[cfg(xen)] inside the function body.

The more important problem as it's also pointed out by the other reviewers is that the code is wrapping unsafe code in safe functions that are in fact not safe.

src/volatile_memory.rs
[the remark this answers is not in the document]
vireshk: I was already updating the code with `unsafe` when your comments came. Done now.

src/volatile_memory.rs
```diff
@@ -37,7 +37,7 @@ use crate::atomic_integer::AtomicInteger;
 use crate::bitmap::{Bitmap, BitmapSlice, BS};
 use crate::{AtomicAccess, ByteValued, Bytes};
 
-use copy_slice_impl::copy_slice;
+use copy_slice_impl::*;
```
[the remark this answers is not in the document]
vireshk: Everything from `copy_clice` is required here, and it is a local module to this file. I will update this to write all names though.

src/volatile_memory.rs
vireshk: This function is already present in main branch and is not marked `unsafe` and so I didn't touch it. Though I updated all the helpers I have added now.

roypat:
Mh, before I dive more into the implementation details here, I have a sorta high-level design question: If I understand this implementation correctly, the on-demand mapping nature of the Xen regions happens in VolatileSlice. This means the flow is: get volatile slice -> call some read/write on it -> mmap is called -> memory is accessed -> munmap is called in drop impl of XenMmapSlice. 

This means the VolatileSlice has to know about the details of the Xen memory model. I don't think it should. In my mind, a VolatileSlice is a VMM-safe version of a rust-slice. This means to me it should always be pointing to existing memory, and it should be completely agnostic about whether that is standard unix mmap'd memory or some xen grant. So I propose we change the flow to be the following:

VMM knows it wants to access xen memory -> do the xen mmap/ioctl -> get a volatile slice to the now existing memory -> use volatile slice to access it -> drop volatile slice and do the munmap (to ensure we dont unmap anything while volatile slices are alive, set the lifetime of the volatile slice to the lifetime of a reference to the XenMmapSlice)

This way, none of the changes to `volatile_slice.rs` should be needed, and the xen specific code gets "contained" a bit more. 

Please let me know if I am missing something here though, as I have never worked with Xen before and am only going off of information contained in this PR

vireshk:
Thanks for your comments Patrick. And yes I still need to reply to your UB related comment, I haven't got enough time to get to it yet :)

> Mh, before I dive more into the implementation details here, I have a sorta high-level design question: If I understand this implementation correctly, the on-demand mapping nature of the Xen regions happens in VolatileSlice. This means the flow is: get volatile slice -> call some read/write on it -> mmap is called -> memory is accessed -> munmap is called in drop impl of XenMmapSlice.
>

Right. I think this is going to be required for other use cases as well, like pKVM, as full access to the guest memory may not be always available but Xen is first at least in our list.

From Xen's (and guest kernel's) standpoint, whenever a buffer is passed to backend, a grant mapping is created in the guest to allow the same to be accesssed by the backend running on host. The backend can only map the regions which are allowed by the guest. As soon as the access of the buffers is over and before they are returned to the guest, they should be unmapped by the backend. Also note that the mapping has memory protection in place, that is most of the time you can map memory as READ only  or WRITE only and not both.

> This means the VolatileSlice has to know about the details of the Xen memory model. I don't think it should. In my mind, a VolatileSlice is a VMM-safe version of a rust-slice. This means to me it should always be pointing to existing memory, and it should be completely agnostic about whether that is standard unix mmap'd memory or some xen grant. So I propose we change the flow to be the following:
> 
> VMM knows it wants to access xen memory -> do the xen mmap/ioctl -> get a volatile slice to the now existing memory -> use volatile slice to access it -> drop volatile slice and do the munmap (to ensure we dont unmap anything while volatile slices are alive, set the lifetime of the volatile slice to the lifetime of a reference to the XenMmapSlice)
> 

The problem here is that when the slice is created, we aren't sure if the user is going to read or write to the slice and so mapping can't be done there. It is only when we try to access the memory within the slice, we get to know about it.

Also about "VMM knows it wants to access xen memory", this is a bit different in our case. The backends run as standalone applications and they are mostly unaware of what hypervisor they are running with, KVM or XEN. And then they are supposed to be hypervisor agnostic, no special code there for Xen.

Right now the volatile slice can be used by vm-memory crate itself, where it just returns the data without the user knowing about the VolatileSlice at all. Or, like in case of Virtiofsd, the user (backend) can take a big Volatile slice and process it as it wishes. These are all very complicated usage patterns and the only single place solution that I could think of was to put the mapping logic in VolatileSlice, as that is central to the entire thing.

> This way, none of the changes to `volatile_slice.rs` should be needed, and the xen specific code gets "contained" a bit more.
> 
> Please let me know if I am missing something here though, as I have never worked with Xen before and am only going off of information contained in this PR

I hope I was able to clarify this a bit. Thanks.

src/mmap_xen.rs
vireshk: I have passed `prot` directly now, which can be libc::PROT_READ/WRITE.

Ablu: Ok. Though that may tempt people to try passing `READ | WRITE` (at least I tried when hacking the virtio-fs test together), but that did not seem to work since the mapping needs to be either readable OR writable apparently?

Ablu: Yeah, I was just doing it as a quick and dirty hack. But mmap blew up, so I figured it out quickly.

roypat:
> Thanks for your comments Patrick. And yes I still need to reply to your UB related comment, I haven't got enough time to get to it yet :)

No worries, I'm not exactly fast with following up on things myself right >.> everything's kinda hectic :(

> > Mh, before I dive more into the implementation details here, I have a sorta high-level design question: If I understand this implementation correctly, the on-demand mapping nature of the Xen regions happens in VolatileSlice. This means the flow is: get volatile slice -> call some read/write on it -> mmap is called -> memory is accessed -> munmap is called in drop impl of XenMmapSlice.
> 
> Right. I think this is going to be required for other use cases as well, like pKVM, as full access to the guest memory may not be always available but Xen is first at least in our list.
> 
> From Xen's (and guest kernel's) standpoint, whenever a buffer is passed to backend, a grant mapping is created in the guest to allow the same to be accesssed by the backend running on host. The backend can only map the regions which are allowed by the guest. As soon as the access of the buffers is over and before they are returned to the guest, they should be unmapped by the backend. Also note that the mapping has memory protection in place, that is most of the time you can map memory as READ only or WRITE only and not both.

Mh, so the guest creates the grant mapping, which I presume sets up the memory protection, too? And then the flow would be "guest creates grant mapping/memory protection" -> "guest passes details of mapping (addr/size?) to backend" -> "backend mmaps in the region to do its accesses"? Sorry for asking so many questions >.>

> > This means the VolatileSlice has to know about the details of the Xen memory model. I don't think it should. In my mind, a VolatileSlice is a VMM-safe version of a rust-slice. This means to me it should always be pointing to existing memory, and it should be completely agnostic about whether that is standard unix mmap'd memory or some xen grant. So I propose we change the flow to be the following:
> > VMM knows it wants to access xen memory -> do the xen mmap/ioctl -> get a volatile slice to the now existing memory -> use volatile slice to access it -> drop volatile slice and do the munmap (to ensure we dont unmap anything while volatile slices are alive, set the lifetime of the volatile slice to the lifetime of a reference to the XenMmapSlice)
> 
> The problem here is that when the slice is created, we aren't sure if the user is going to read or write to the slice and so mapping can't be done there. It is only when we try to access the memory within the slice, we get to know about it.

Mh, why doesn't it? If the memory protection is done in the guest, then the backend should know ahead of time whether it can read or write to the memory region, so it could use the information to map it ahead of time. Because if the user backend just decides whether it can access via read/write on the fly (and if the guest doesn't enforce anything), then we might as well just mmap it as read/write ahead-of-time because there'd be no enforcement of "read/write-only" in place anyway :thinking:.

Also, isn't doing it on-the-fly like this really inefficient? Won't we end up doing mmap/munmap on every single call to a volatile slice method (which I presume we'll have many of)

> Also about "VMM knows it wants to access xen memory", this is a bit different in our case. The backends run as standalone applications and they are mostly unaware of what hypervisor they are running with, KVM or XEN. And then they are supposed to be hypervisor agnostic, no special code there for Xen.

gotcha, makes sense.

> Right now the volatile slice can be used by vm-memory crate itself, where it just returns the data without the user knowing about the VolatileSlice at all. Or, like in case of Virtiofsd, the user (backend) can take a big Volatile slice and process it as it wishes. These are all very complicated usage patterns and the only single place solution that I could think of was to put the mapping logic in VolatileSlice, as that is central to the entire thing.

Mh, I'm still not convinced it has to be in VolatileSlice :sweat_smile: 

> > This way, none of the changes to `volatile_slice.rs` should be needed, and the xen specific code gets "contained" a bit more.
> > Please let me know if I am missing something here though, as I have never worked with Xen before and am only going off of information contained in this PR
> 
> I hope I was able to clarify this a bit. Thanks.

Yes, thank you so much for taking the time to explain this to me!

src/mmap_xen.rs
vireshk: > This is triggering undefined behavior: the `ptr` returned by `alloc::alloc` points to uninitialized memory. Then in line 289, you're dereferencing this pointer to get a mutable reference to this uninitialized memory [[1](https://doc.rust-lang.org/reference/behavior-considered-undefined.html)]. Since you also write through this pointer, it causes further UB as uninitialized memory will get dropped.
> 
> I'm not familiar with the API in question, but this looks a bit like a FAM, so maybe the vmm-sys-utils abstractions for those can be used to make this safe https://github.com/rust-vmm/vmm-sys-util/blob/main/src/fam.rs?

Thanks for pointing to this. I have now migrated updated the code to FAM.

roypat:
> > Mh, why doesn't it? If the memory protection is done in the guest, then the backend should know ahead of time whether it can read or write to the memory region, so it could use the information to map it ahead of time. Because if the user backend just decides whether it can access via read/write on the fly (and if the guest doesn't enforce anything), then we might as well just mmap it as read/write ahead-of-time because there'd be no enforcement of "read/write-only" in place anyway thinking.
> 
> When I said "when the slice is created, we aren't sure ...", it meant from vm-memory's perspective. Right now the API provided by vm-memory crate, lets a user (like backend) to get part of guest memory as slice (Volatile memory) and then accesses it. The vm-memory crate doesn't have any information about read/write permissions for the slice.
> 
> Also we won't wanna add special code to do the extra mappings to the backends, but make this all work magically for the user.
>
> There are few complex cases as well, like Virtiofsd. It takes a large area as a slice I guess and then does a lot of operations on it by itself and doesn't use the APIs provided by vm-memory crate. I am not sure if it does both read and write to separate areas after a volatile slice is created.

Mh, this brings up a separate point that I missed so far. This PR makes the `as_ptr()` method private, but a lot of the ecosystem depends on this function (for example to pass pointers to guest memory to ioctls or normal syscalls). Removing it breaks a lot of dependencies without providing a workaround (for example, the very virtiofsd operation you describe wouldn't work anymore, since it would now be forced to use vm-memory functions). I understand why you did this (because it allows the user to circumvent the mapping mechanisms for Xen), but really that just ties back to my original concern: Conceptually, when we have a `VolatileSlice`, the memory it points to needs to already be mapped. 

I understand not wanting to have to explicitly do the mapping in the backend, but I've had a look at the accompanying PR in the vhost crate, and it looks like even with this implementation, you still need Xen specific code there, too. So unless I misunderstand something (which I very much might do - both xen and vhost user are not familiar to me), just doing all the required mapping in the backend would simplify the code here (with vm-memory providing the required utility functions for creating the mappings in the `mmap_xen.rs` files, and doing all the heavy lifting), while not changing the status quo for the backend much. Surely we must know what sections of memory we are granted read and which sections of memory we are granted write access to, right?

vireshk:
> Mh, this brings up a separate point that I missed so far. This PR makes the `as_ptr()` method private, but a lot of the ecosystem depends on this function (for example to pass pointers to guest memory to ioctls or normal syscalls). Removing it breaks a lot of dependencies without providing a workaround (for example, the very virtiofsd operation you describe wouldn't work anymore, since it would now be forced to use vm-memory functions). I understand why you did this (because it allows the user to circumvent the mapping mechanisms for Xen), but really that just ties back to my original concern: Conceptually, when we have a `VolatileSlice`, the memory it points to needs to already be mapped.

Key takeaways from yesterday's discussion with Patrick (for others):

- It doesn't seem possible to do the mapping from vhost/vhost-user-backend crates or the backends.
- We probably need to do it in `VolatileMemory` only.
- Viresh agreed that we can't really remove `as_ptr()` and it is still required by many, but we need a separate implementation for Xen at least. Will try something for that in the next iteration.

roypat (dismissed):
~~wrong PR, sorry~~

src/mmap_unix.rs
```diff
@@ -537,6 +691,30 @@ mod tests {
...
+
+        use crate::mmap_xen::MmapXenFlags;
+
+        impl<B: Bitmap> MmapRegionBuilder<B> {
+            pub fn with_mmap_info_def(self) -> Result<Self> {
+                self.with_mmap_info(GuestAddress(0), MmapXenFlags::UNIX.bits(), 0)
+            }
+        }
+    }
+
```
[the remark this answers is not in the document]
roypat: (this is not the entire review, I accidentally published the in-progress review when I wanted to do so on the other PR. Really sorry about that)

vireshk:
> * Viresh agreed that we can't really remove `as_ptr()` and it is still required by many, but we need a separate implementation for Xen at least. Will try something for that in the next iteration.

@roypat 

I have preserved the existing functionality of as_ptr() for non-xen consumers and provided a different API for everyone [01bc496f2848](https://github.com/rust-vmm/vm-memory/pull/231/commits/01bc496f2848ef53a4ff49a355fea7d7137e662e).

src/volatile_memory.rs
```diff
@@ -235,6 +237,55 @@ impl<'a> From<&'a mut [u8]> for VolatileSlice<'a, ()> {
...
+            if is_read {
+                libc::PROT_READ
+            } else {
+                libc::PROT_WRITE
+            },
+        );
+
+        Self {
+            // SAFETY: safe because this is built from an already available slice.
+            slice: unsafe { slice::from_raw_parts_mut(mmap_slice.addr(), len) },
```
Ablu, line 275: But who guards us from mapping this region twice? That would cause unsafety, right?

Also, I think this is unsafe because we are still dealing with volatile memory that a guest may write to (even if it is forbidden to do) which would probably be undefined behaviour on the Rust side. VolatileSlice seems to dance around this carefully by not treating the memory as slice but only offering to copy it to and from slices (at leasts that was my understanding when I last reviewed the code). So probably something similar would need to be done here (while also giving access to the raw pointer for zero-copy mechanisms). In fact, I would argue that if you call this API, the convenience functions of VolatileSlice are not enough for you (since you are using Xen and want access to the pointer for zero-copy read/writes). So maybe we can just get rid of the slice completely and only offer a pointer?

src/volatile_memory.rs
```diff
@@ -285,10 +336,16 @@ impl<'a, B: BitmapSlice> VolatileSlice<'a, B> {
     /// Returns a pointer to the beginning of the slice. Mutable accesses performed
     /// using the resulting pointer are not automatically accounted for by the dirty bitmap
     /// tracking functionality.
+    #[cfg(not(all(feature = "xen", unix)))]
     pub fn as_ptr(&self) -> *mut u8 {
         self.addr
     }
 
+    /// Returns a guard for the pointer to the underlying memory.
+    pub fn as_ptr_guard(&self, is_read: bool) -> VolatileSliceGuard {
```
Ablu, line 345: I am trying to compare this to other APIs that I have seen within the standard library... Usually `as_*` are just simpler conversion methods that have close to 0 overhead, right? Would it maybe make sense to just call this `(m)map()`?

Ablu (commented):
Tested successfully with virtiofsd (after converting to the `as_ptr_guard` API) on Xen + xen-vhost-frontend (with a minor patch to fix to switch to `GuestMmapRange::with_xen`)

Ablu:
Should we maybe mark the as_ptr() as deprecated? Then we can steer people to use the Xen-compatible API without breaking all the existing code.

src/volatile_memory.rs
vireshk: Discussed with @Ablu offline and decided to remove the slice support and rather expose the pointer directly in the new implementation.

@Ablu please resolve this comment if you are satisfied with the new patches.

src/volatile_memory.rs
vireshk: Named this as ptr_guard() and ptr_guard_mut() now.

Ablu: LGTM (cannot mark as resolved since I commented on a commit that no longer exists)

src/volatile_memory.rs
Ablu: LGTM (cannot mark as resolved since I commented on a commit that no longer exists)

src/volatile_memory.rs
```diff
@@ -219,13 +235,95 @@ impl<'a> From<&'a mut [u8]> for VolatileSlice<'a, ()> {
 #[repr(C, packed)]
 struct Packed<T>(T);
 
+struct VolatileSliceGuardInner {
```
roypat, line 238: I don't think we need to encapsulate this. Something like 

```rs
impl VolatileSliceGuard {
   fn new_read(...)
   fn new_write(...)
```

should work instead of `VolatileSliceGuard` and `VolatileSliceGuardMut`. However, this does not actually work the way it is intended, because it doesn't "guard": Since there's no lifetime on raw pointers, nothing prevents the guard (and thus the underlying MmapSlice) from being dropped while we still hold on to the return value of `as_ptr`. I don't think there is a way to properly express this, so if this isn't _needed_ for Xen (which based on our convo in the last meeting it doesn't seem to be), maybe we should offer this? (or just make mmap_slice() public and have people use its `as_ptr` on xen.

Ablu: > However, this does not actually work the way it is intended, because it doesn't "guard": Since there's no lifetime on raw pointers, nothing prevents the guard (and thus the underlying MmapSlice) from being dropped while we still hold on to the return value of `as_ptr`. I don't think there is a way to properly express this, so if this isn't _needed_ for Xen (which based on our convo in the last meeting it doesn't seem to be), maybe we should offer this? (or just make mmap_slice() public and have people use its `as_ptr` on xen.

@roypat when I sketched the guard idea as a hack to test virtiofsd during an earlier iteration (originally it used `MmapXenSlice` and just hacked it public) I mostly thought about it as "guarding the pointer" to be available. Of course nothing is ensuring that the pointer will still be valid, but I guess that is the nature of the pointer? Nothing keeps it valid in case someone just calls .as_ptr() too. But I agree that the name Guard may be misleading. We discussed it this morning and came up with nothing better :(. Also merging with the existing `MmapXenSlice` may be worthwile. However, I wonder: do we really want to fracture the ecosystem by providing two APIs for the same thing?

I understand that it is sensible to keep `.as_ptr()` for backwards compatibility, but only providing the new API through `mmap_slice` (which is Xen specific) would push the conditional compilation into the consumer of this lib. My fear is that nobody will bother and people will only implement one or the other variant and the ecosystem will be split between Xen and KVM.

I think one could maybe rename `MmapXenSlice` to something generic that also exists for KVM (but does nothing really) and expose the mmap() call, but that feels like a weird API for KVM where nothing happens there. @vireshk and me actually discussed calling the guard creation function `mmap` this morning, but rejected that idea because of it feeling weird for KVM.

So maybe we should first discuss which differences between KVM and Xen we want to allow in the API first? Personally, I would suggest to advertise an identical API for both KVM and Xen in order to minimize fragmention. Of course removing .as_ptr() is be too invasive, but we could for example just mark it as deprecated? The extra calls are empty for KVM and should get optimized away, so I do not think that we have any performance issues there.

We are really not aiming at a "on xen" and "on kvm" world here. We want to abstract the differences away as much as possible :).

vireshk: > I don't think we need to encapsulate this. Something like
> 
> ```rust
> impl VolatileSliceGuard {
>    fn new_read(...)
>    fn new_write(...)
> ```

I can't create the Guard without knowing if it is a `read` or `write` operation, as the mapping needs that information. Once I have done the mapping, the only thing I need to return is a pointer, which the consumers can use to avoid an extra copy of data. I am not sure how these new_read/write routines are supposed to work, return. Sorry, a bit confused here :(

> should work instead of `VolatileSliceGuard` and `VolatileSliceGuardMut`. However, this does not actually work the way it is intended, because it doesn't "guard": Since there's no lifetime on raw pointers, nothing prevents the guard (and thus the underlying MmapSlice) from being dropped while we still hold on to the return value of `as_ptr`.

Right. Maybe we need a better name here, as what we are looking for here is a guaranted way by which the memory can be accessed safely.

> I don't think there is a way to properly express this, so if this isn't _needed_ for Xen (which based on our convo in the last meeting it doesn't seem to be), maybe we should offer this? (or just make mmap_slice() public and have people use its `as_ptr` on xen.

We would need this for Xen too, for example we intend to reused Virtiofsd crate, which uses this to avoid extra copying of data. And so a standard way by which they can get the pointer, instead of adding `#cfg(xen)`, would be preferred here.

vireshk: > I understand that it is sensible to keep `.as_ptr()` for backwards compatibility, but only providing the new API through `mmap_slice` (which is Xen specific) would push the conditional compilation into the consumer of this lib. My fear is that nobody will bother and people will only implement one or the other variant and the ecosystem will be split between Xen and KVM.

Also note that we will need something very similar going forward with p-KVM (protected-kvm) as well, as the entire guest memory being made available to the backend won't work for very long. People will definitely want to restrict it. I am not sure of the interface that is going to be available then, but mapping surely must happen on-demand somehow.

roypat: Right, what I meant with the code change is that the `as_guard` and `as_guard_mut` methods return `VSGuard/VSGuardMut::new` (which then calls VSGuardInner::new with read/write access specified), but instead can just immediately return `VSGuardInner::new_read` and `VSGuardInner::new_write` (or with the current function name, immediately return `VolatileSliceGuardInner::new(..., proto: READ/WRITE)`). No need for the *Mut encapsulation. 

So I was under the impression this wasn't needed for Xen as an earlier iteration of this PR did not include this functionality (as as_ptr was made private and thus not available at all) and I was told that virtiofsd still worked. So in that sense I think it makes more sense to err on the side of less functionality (as opposed to committing something we're unsure of and is not required) and revisit this question at a later point in time (e.g. leave the API unchanged for unix and think about a unification later). I'm open to deprecating as_ptr and finding a generalization, but as a follow up if its not required immediately, to keep this PR minimal wrt to Xen support.

vireshk: > Right, what I meant with the code change is that the `as_guard` and `as_guard_mut` methods return `VSGuard/VSGuardMut::new` (which then calls VSGuardInner::new with read/write access specified), but instead can just immediately return `VSGuardInner::new_read` and `VSGuardInner::new_write` (or with the current function name, immediately return `VolatileSliceGuardInner::new(..., proto: READ/WRITE)`). No need for the *Mut encapsulation.

Ahh I understood it now. I implemented two variants so they can have different `as_ptr()` methods, the read one returning `*const u8` and write one retuning `*mut u*`. If it is fine to return `*mut u8` for both read and write, then I can work with just one structure instead of three. Sure.
 
> So I was under the impression this wasn't needed for Xen as an earlier iteration of this PR did not include this functionality (as as_ptr was made private and thus not available at all) and I was told that virtiofsd still worked.

Maybe I wasn't able to clarify that earlier. Lemme do so now. The early version was tested with vhost-device implementations, which didn't use `as_ptr()` and worked. Then @Ablu tested virtiofsd and it didn't work. We hacked it up a bit to see if Xen grant is working or not and at the same time I privatized `as_ptr()` to make sure it is clear that we can't use it anymore.

In short, we still need something like `as_ptr()` for virtiofsd to work.

> So in that sense I think it makes more sense to err on the side of less functionality (as opposed to committing something we're unsure of and is not required) and revisit this question at a later point in time (e.g. leave the API unchanged for unix and think about a unification later). I'm open to deprecating as_ptr and finding a generalization, but as a follow up if its not required immediately, to keep this PR minimal wrt to Xen support.

As i explained above, it is required right away for virtiofsd. Sorry for the confusion.

vireshk:
Superseeded by https://github.com/rust-vmm/vm-memory/pull/241

PR #235, opened 2023-05-29
Mmap/reorg

vireshk:
This reorganizes some code in the vm-memory crate, which will later be used by:

https://github.com/rust-vmm/vm-memory/pull/231


src/mmap_unix.rs
```diff
@@ -51,6 +51,73 @@ pub enum Error {
...
+    owned: bool,
+}
+
+impl Mmap {
+    pub fn new(size: usize, prot: i32, flags: i32, fd: i32, f_offset: u64) -> Result<Self> {
+        let addr =
+            // SAFETY: This is safe because we're not allowing MAP_FIXED, and invalid parameters
+            // cannot break Rust safety guarantees (things may change if we're mapping /dev/mem or
+            // some wacky file).
+            unsafe { libc::mmap(null_mut(), size, prot, flags, fd, f_offset as libc::off_t) };
```
Ablu, line 67: Do we need some mechanism to prevent mapping the same region twice? Especially with subslicing going on it seems to be relatively easy to map the same region twice? Does not seem to be a new issue though... The code just moved.

src/volatile_memory.rs
```diff
@@ -1390,6 +1434,28 @@ mod copy_slice_impl {
...
+        slice: &VolatileSlice<'_, B>,
+        total: usize,
+    ) -> usize {
+        // SAFETY: `slice` and `dst` must be point to a contiguously allocated memory region of at
+        // least length `total`. The regions must not overlap.
+        unsafe { copy_slice(dst, slice.as_ptr(), total) }
+    }
+
+    /// Copies `total` bytes from 'src' to `slice`
+    pub(super) unsafe fn copy_to_volatile_slice<B: BitmapSlice>(
```
Ablu, line 1450: I think in general, it would be better to add the safety comment on this function and then, inside the body just do  something like: `// SAFETY: guaranteed by function invariants`. Though this is just `pub(super)`, so...

src/volatile_memory.rs
```diff
@@ -1261,24 +1270,59 @@ where
...
+            unsafe {
+                *v = read_volatile(addr).0;
+                addr = addr.add(1);
+            };
+            i += 1;
+        }
+        i
+    }
+
+    unsafe fn write_volatile_aref(&self, buf: &[T], mut addr: *mut Packed<T>, count: usize) {
```
Ablu, line 1301: why does this include `&self`, but the `read_volatile_aref` does not?

Ablu (commented):
just some minor comments. People with deeper code knowledge will have to comment on the architecture :).

src/mmap_unix.rs
vireshk: This is not an issue with the current code at least, it can happen only after Xen bits come in. And there too, I am not sure if we will end up having that optimization to avoid code complexity.

src/volatile_memory.rs
vireshk: sure

src/volatile_memory.rs
vireshk: I was trying to update the bitmap from within it, changed now to make both of them look same.

src/mmap_unix.rs
```diff
@@ -242,6 +287,30 @@ impl<B: Bitmap> MmapRegionBuilder<B> {
     }
 }
 
+impl<B: NewBitmap> MmapRegionBuilder<B> {
+    fn from_range(range: MmapRange) -> Result<Self> {
```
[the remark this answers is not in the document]
vireshk: `from_range()` is used for multiple structures already and so I thought to keep it like this. Maybe we can convert them all later..

src/mmap_unix.rs
```diff
@@ -338,11 +377,13 @@ impl<B: NewBitmap> MmapRegion<B> {
     ///
     /// To use this safely, the caller must guarantee that `addr` and `size` define a region within
     /// a valid mapping that is already present in the process.
-    pub unsafe fn build_raw(addr: *mut u8, size: usize, prot: i32, flags: i32) -> Result<Self> {
-        MmapRegionBuilder::new_with_bitmap(size, B::with_len(size))
+    pub unsafe fn build_raw(addr: *mut u8, range: MmapRange) -> Result<Self> {
+        if range.file.is_some() {
+            return Err(Error::InvalidRange(range));
+        }
```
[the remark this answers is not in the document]
vireshk: This checks for the presence of the file, which is an error condition here. `from_range()` doesn't check that.

src/mmap_unix.rs
```diff
@@ -47,10 +47,55 @@ pub enum Error {
...
+    InvalidRange(MmapRange),
 }
 
 pub type Result<T> = result::Result<T, Error>;
 
+/// `MmapRange` represents a range of arguments required to create Mmap regions.
+#[derive(Clone, Debug)]
+pub struct MmapRange {
+    size: usize,
+    file: Option<FileOffset>,
```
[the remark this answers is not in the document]
vireshk: I won't touch it for now :)

src/mmap_unix.rs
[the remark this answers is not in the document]
Ablu: Oh you are right. Got confused somewhere.

src/mmap_unix.rs
```diff
@@ -47,10 +47,55 @@ pub enum Error {
...
+}
+
+impl MmapRange {
+    /// Creates instance of the range with `size`.
+    pub fn new(size: usize) -> Self {
+        Self::new_with(size, None, None, None)
+    }
+
+    /// Creates instance of the range with multiple arguments.
+    pub fn new_with(
```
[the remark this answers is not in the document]
vireshk: named as with_file() now.

Ablu (approved):

src/atomic.rs
```diff
@@ -165,17 +168,18 @@ mod tests {
...
         assert_eq!(regions, iterated_regions);
         assert_eq!(mem.num_regions(), 2);
         assert!(mem.find_region(GuestAddress(0x1000)).is_some());
         assert!(mem.find_region(GuestAddress(0x10000)).is_none());
 
-        assert!(regions
-            .iter()
-            .map(|x| (x.0, x.1))
-            .eq(iterated_regions.iter().copied()));
+        assert!(regions.iter().eq(iterated_regions.iter()));
```
epilys, line 182: Isn't that guaranteed by the `assert_eq!(regions, iterated_regions);` above? (I'm not sure about the types here)

src/guest_memory.rs
```diff
@@ -331,7 +332,7 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = Error> {
 /// # #[cfg(feature = "backend-mmap")]
 /// # {
 /// # use std::sync::Arc;
-/// # use vm_memory::{GuestAddress, GuestAddressSpace, GuestMemory, GuestMemoryMmap};
+/// # use vm_memory::{ GuestAddress, GuestAddressSpace, GuestMemory, GuestMemoryMmap, GuestMmapRange, };
```
epilys, line 335:
```suggestion
/// # use vm_memory::{GuestAddress, GuestAddressSpace, GuestMemory, GuestMemoryMmap, GuestMmapRange};
```

src/mmap_windows.rs
```diff
@@ -61,6 +61,34 @@ pub const INVALID_HANDLE_VALUE: RawHandle = (-1isize) as RawHandle;
...
+#[derive(Clone, Debug)]
+pub struct MmapRange {
+    size: usize,
+    file: Option<FileOffset>,
+}
+
+impl MmapRange {
+    /// Creates instance of the range with `size`.
+    pub fn new(size: usize) -> Self {
+        Self::with_file(size, None)
```
epilys, line 74: Is this calling `mmap(2)` under the hood? If `size` is `0`, it would return `EINVAL`

src/mmap.rs
```diff
@@ -481,26 +504,12 @@ impl<B: NewBitmap> GuestMemoryMmap<B> {
...
+    /// Valid memory regions are specified as a slice of `GuestMmapRange` items sorted by Address.
+    pub fn from_ranges(ranges: &[GuestMmapRange]) -> result::Result<Self, Error> {
         Self::from_regions(
             ranges
-                .into_iter()
-                .map(|x| {
-                    GuestRegionMmap::from_range(x.borrow().0, x.borrow().1, x.borrow().2.clone())
-                })
+                .iter()
+                .map(|r| GuestRegionMmap::from_range(r))
```
epilys, line 512:
```suggestion
                .map(GuestRegionMmap::from_range)
```

src/mmap.rs
```diff
@@ -708,7 +727,14 @@ mod tests {
             regions_summary
                 .iter()
                 .map(|(region_addr, region_size)| {
-                    Arc::new(GuestRegionMmap::from_range(*region_addr, *region_size, None).unwrap())
+                    Arc::new(
+                        GuestRegionMmap::from_range(&GuestMmapRange::new(
```
epilys, line 731:
```suggestion                  
GuestRegionMmap::from_range(&GuestMmapRange::new(
```

src/mmap.rs
```diff
@@ -886,7 +912,8 @@ mod tests {
 
     #[test]
     fn slice_addr() {
-        let m = GuestRegionMmap::from_range(GuestAddress(0), 5, None).unwrap();
+        let m =
+            GuestRegionMmap::from_range(&GuestMmapRange::new(GuestAddress(0), 5, None)).unwrap();
```
epilys, line 916:
```suggestion
GuestRegionMmap::from_range(&GuestMmapRange::new(GuestAddress(0), 5, None)).unwrap();
```

src/mmap.rs
```diff
@@ -1029,8 +1073,9 @@ mod tests {
         f.set_len(0x400).unwrap();
 
         let start_addr = GuestAddress(0x0);
-        let guest_mem = GuestMemoryMmap::from_ranges(&[(start_addr, 0x400)]).unwrap();
-        let guest_mem_backed_by_file = GuestMemoryMmap::from_ranges_with_files(&[(
+        let guest_mem =
+            GuestMemoryMmap::from_ranges(&[GuestMmapRange::new(start_addr, 0x400, None)]).unwrap();
```
epilys, line 1077:
```suggestion
GuestMemoryMmap::from_ranges(&[GuestMmapRange::new(start_addr, 0x400, None)]).unwrap();
```

src/mmap.rs
```diff
@@ -1107,8 +1155,9 @@ mod tests {
         f.set_len(0x400).unwrap();
 
         let mut start_addr = GuestAddress(0x1000);
-        let gm = GuestMemoryMmap::from_ranges(&[(start_addr, 0x400)]).unwrap();
-        let gm_backed_by_file = GuestMemoryMmap::from_ranges_with_files(&[(
+        let gm =
+            GuestMemoryMmap::from_ranges(&[GuestMmapRange::new(start_addr, 0x400, None)]).unwrap();
```
epilys, line 1159:
```suggestion
GuestMemoryMmap::from_ranges(&[GuestMmapRange::new(start_addr, 0x400, None)]).unwrap();
```

src/mmap.rs
```diff
@@ -1138,8 +1187,10 @@ mod tests {
         let f = TempFile::new().unwrap().into_file();
         f.set_len(0x400).unwrap();
 
-        let gm = GuestMemoryMmap::from_ranges(&[(GuestAddress(0x1000), 0x400)]).unwrap();
-        let gm_backed_by_file = GuestMemoryMmap::from_ranges_with_files(&[(
+        let gm =
+            GuestMemoryMmap::from_ranges(&[GuestMmapRange::new(GuestAddress(0x1000), 0x400, None)])
```
epilys, line 1191:
```suggestion
GuestMemoryMmap::from_ranges(&[GuestMmapRange::new(GuestAddress(0x1000), 0x400, None)])
```

src/volatile_memory.rs
```diff
@@ -1261,24 +1270,62 @@ where
...
+            // the slices are created, and they never escape the VolatileSlices.
+            // ptr::add is safe because get_array_ref() validated that
+            // size_of::<T>() * self.len() fits in an isize.
+            unsafe {
+                *v = read_volatile(addr).0;
+                addr = addr.add(1);
+            };
+            i += 1;
+        }
+        i
```
epilys, line 1299:
```suggestion
        std::cmp::min(count, buf.len())
```

src/volatile_memory.rs
```diff
@@ -1261,24 +1270,62 @@ where
...
+        for v in buf.iter_mut().take(count) {
+            // SAFETY: read_volatile is safe because the pointers are range-checked when
+            // the slices are created, and they never escape the VolatileSlices.
+            // ptr::add is safe because get_array_ref() validated that
+            // size_of::<T>() * self.len() fits in an isize.
+            unsafe {
+                *v = read_volatile(addr).0;
+                addr = addr.add(1);
+            };
+            i += 1;
```
epilys, line 1297:
```suggestion
```

src/volatile_memory.rs
```diff
@@ -1261,24 +1270,62 @@ where
...
+            // the slices are created, and they never escape the VolatileSlices.
+            // ptr::add is safe because get_array_ref() validated that
+            // size_of::<T>() * self.len() fits in an isize.
+            unsafe { self.write_volatile(buf) };
+        }
+    }
+
+    // Returns number of elements read
+    unsafe fn read_volatile_aref(buf: &mut [T], mut addr: *const Packed<T>, count: usize) -> usize {
+        let mut i = 0;
```
epilys, line 1286:
```suggestion
```

src/volatile_memory.rs
```diff
@@ -1261,24 +1270,62 @@ where
...
+            // the slices are created, and they never escape the VolatileSlices.
+            // ptr::add is safe because get_array_ref() validated that
+            // size_of::<T>() * self.len() fits in an isize.
+            unsafe {
+                write_volatile(addr, Packed::<T>(v));
+                addr = addr.add(1);
+            };
         }
+
+        addr as usize - start as usize
```
epilys, line 1317: Don't know if this would be safe, the rust docs for this method have a "Safety" section to see if it would apply. Not sure about this. https://doc.rust-lang.org/core/primitive.pointer.html#method.offset_from

```suggestion
        addr.offset_from(start).unsigned_abs()
```

src/mmap_unix.rs
```diff
@@ -505,7 +580,13 @@ mod tests {
         let buf1 = [1u8, 2, 3, 4, 5];
 
         f.write_all(buf1.as_ref()).unwrap();
-        let r = MmapRegion::from_file(FileOffset::new(f, offset as u64), buf1.len()).unwrap();
+        let range = MmapRange::with_file(
+            buf1.len(),
+            Some(FileOffset::new(f, offset as u64)),
```
epilys, line 585: `as ` conversions are considered bad practice because if later the type of e.g. `offset` changes and becomes larger, it would be silently truncated.

src/atomic.rs
```diff
@@ -165,17 +168,18 @@ mod tests {
         }
 
         for region in mem.iter() {
-            iterated_regions.push((region.start_addr(), region.len() as usize));
+            iterated_regions.push(GuestMmapRange::new(
+                region.start_addr(),
+                region.len() as usize,
```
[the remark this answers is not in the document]
vireshk: Hmm, since this was already present before this series, maybe doing all such changes at once in a separate pull request would be better ?

src/atomic.rs
vireshk: Yes, you are right.

src/mmap_windows.rs
vireshk: No, this is just creating a structure with arguments.

src/mmap.rs
vireshk: Not sure I understand this one. The indentation is required here.

src/mmap.rs
vireshk: same here.

src/mmap.rs
vireshk: same.

src/volatile_memory.rs
vireshk: What about this instead ?

`addr.offset_from(start) as usize`, here `start` is the initial value of `addr`.

src/volatile_memory.rs
vireshk: `addr.offset_from(start)` gives number of elements and not bytes.

src/mmap_unix.rs
vireshk: I am not touching it for now, since it was already there. But I do understand what you are saying.

src/atomic.rs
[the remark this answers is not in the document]
epilys: Yeah! If you want you can make an issue out of it by clicking on "Reference in new issue" on the "..." menu and resolve this comment.

src/volatile_memory.rs
epilys: oh my bad, thought it was u8 pointers.

src/volatile_memory.rs
```diff
@@ -955,22 +956,20 @@ where
         &self.bitmap
     }
 
-    /// Does a volatile write of the value `v` to the address of this ref.
     #[inline(always)]
-    pub fn store(&self, v: T) {
+    unsafe fn write_volatile(&self, addr: *mut Packed<T>, v: T) {
         // SAFETY: Safe because we checked the address and size when creating this VolatileRef.
-        unsafe { write_volatile(self.addr, Packed::<T>(v)) };
+        unsafe { write_volatile(addr, Packed::<T>(v)) };
```
jiangliu, line 962: The function is marked as `unsafe fn`, so is `unsafe` within the function body redundant?

src/mmap.rs
epilys: Hm that's bizarre, I don't know/remember why I posted those (maybe github was showing me the wrong indentation?) In any case sorry for the noise @vireshk

src/volatile_memory.rs
vireshk: Found same issue at few other places, fixed all of those too. Thanks.

epilys: It is (for now!) But that's gonna change in the future: https://github.com/rust-lang/rust/issues/71668

In any case, now it's only a lint thus not wrong or invalid

epilys (approved):

jiangliu (dismissed):

JonathanWoollett-Light:
On [volatile_memory: Remove redundant unsafe blocks](https://github.com/rust-vmm/vm-memory/pull/235/commits/57693f17a53f483417fe492ea7ec356ad0fdeeca) while the lint name doesn't spring to my mind immediately I beleive there is one. Are we not currently warning on this? Would it make sense to add this in this commit?

vireshk:
> On [volatile_memory: Remove redundant unsafe blocks](https://github.com/rust-vmm/vm-memory/pull/235/commits/57693f17a53f483417fe492ea7ec356ad0fdeeca) while the lint name doesn't spring to my mind immediately I beleive there is a lint for checking for `unsafe` blocks within `unsafe` blocks. Are we not currently warning on this? Would it make sense to add this in this commit?

@epilys did mention about another lint which can be of use here: https://github.com/rust-lang/rust/issues/71668 

I couldn't find a lint which will warn if `unsafe` block is added to a `unsafe` fn though.

epilys:
@JonathanWoollett-Light @vireshk the lint for nested `unsafe` blocks is [`clippy::multiple_unsafe_ops_per_block`](https://rust-lang.github.io/rust-clippy/master/index.html#/multiple_unsafe_ops_per_block) the link for blocks in `unsafe` functions does not seem to be stable yet: `unsafe_block_in_unsafe_fn` so we can ignore it for now

roypat (changes requested):
Alright, I've gone through all the commits. 075e37c42ae414f97452f439e5530153c3dea58a is fine, but for the others I have some high level questions/comments. 
1. 57693f17a53f483417fe492ea7ec356ad0fdeeca: As others have already pointed out, there is https://rust-lang.github.io/rfcs/2585-unsafe-block-in-unsafe-fn.html, which is an accepted rust-lang RFC and means that going forwards, not having the unsafe blocks will be at the very least a warning. We should maybe consider enabling this lint ahead of time for this crate, but that's out of scope here. Having the unsafe blocks even inside unsafe functions is quite valuable imo, because it immediately tells you what parts of a function is actually performing unsafe operations, and also forces you to accurately document the safety invariants (by means of `#![deny(clippy::undocumented_unsafe_blocks)]`). In that sense, I don't think this commit is a change we want to do. 
2. 4d73f17e97763b108d5ab958d49ba4466ea44594: The call stack of these methods is starting to be very deep, which makes understanding the code pretty complex (this is already a big problem in this crate, but I'd like to not make it worse). In this PR all the new methods have exactly one call-site. I think the context for these changes is that in the follow up, you'll want to call these from different cfg'd contexts. However, as Andreea noted on that PR, you can mark individual lines of code with `#[cfg(...)]` instead of whole methods. With that, the number of call-sites should stay at 1 I think, and these functions can be re-inlined, so this commit isn't necessary. 
3. f3a9fc57133b5419ea6a9264ffe4c18535cd08ab: The changes in this commit are not motivated in the commit message. I'm not sure why this is required, and it adds cognitive similar complexity as the previous commit. I see that in the follow up PR, it is this structure that VolatileSlice gains a reference to, but why can't we just reuse `MmapRegion` there? The type parameter shouldn't cause problems, since VolatileSlice has it, too. 
4. 6cfe4178350a09175eab79d54751c8f4f944f174: I understand that this is based on a concern Andreea had on the follow up PR. My main concern is the same as the previous two commits, that we start introducing too many layers of abstraction here. The structures added here are a new layer on top of MmapRegionBuilder. The diff of this commit touches mostly on test code, so I think what happened at some point in the past is that someone did not want to use the builder in test code, so they added the tuple-based construction methods as shortcuts. Any actual new information that is used during construction of MmapRegions should be integrated with the MmapRegionBuilder (e.g. the two new Xen fields from the follow up PR should become #[cfg(xen)] fields in the build, with associated setters). Outside of test code, the builder should always be used to construct mmap regions. In that sense, I think the implementation how it was previous to this commit is fine (it should probabl be moved into a test module, but I'd like to keep breaking changes to the public API to a minimum). However, this problem will resurface when it comes to testing the Xen code, as using builders is probably overly verbose. I propose we revisit when it comes to that point, and then add a subset of the changes introduced in the this PR (having a struct that encapsules the fields of the tuple, and a utility method that uses the build to construct a GuestMmapRegion from the tuple-struct) in the Xen test module. The existing methods can be then marked as `#[cfg(not(xen))]`.

Sorry if this is a lot (or if I got some things wrong - cross-referencing with the other PR was difficult at time). The main themes in my feedback are keeping hierarchies flat, and keeping the changes to the public API for non-xen users minimal.  Before I do another of this PR/the follow up, could you try addressing @fandree's [comment](https://github.com/rust-vmm/vm-memory/pull/231#pullrequestreview-1449145413) in the other PR about minimizing the scope of cfg blocks? I think that will make review a lot easier, and should bring down the size of the diffs quite a bit too. Thanks!

vireshk:
Thanks for the review Patrick, really appreciate it !

> 1. [57693f1](https://github.com/rust-vmm/vm-memory/commit/57693f17a53f483417fe492ea7ec356ad0fdeeca): I don't think this commit is a change we want to do.

Dropped.

> 2. [4d73f17](https://github.com/rust-vmm/vm-memory/commit/4d73f17e97763b108d5ab958d49ba4466ea44594): With that, the number of call-sites should stay at 1 I think, and these functions can be re-inlined, so this commit isn't necessary.

It may look messy without that, but I am okay with dropping it. Dropped.

> 3. [f3a9fc5](https://github.com/rust-vmm/vm-memory/commit/f3a9fc57133b5419ea6a9264ffe4c18535cd08ab): The changes in this commit are not motivated in the commit message.

I am not sure I understood that.

> I'm not sure why this is required.

There are a couple of reasons for that:
- One is that this `struct Mmap` is used by mmap_xen.rs file (renamed as `MmapUnix` there) as well, for all the xen variants.
- Also the way it is currently implemented doesn't look very clean or Rust-ish to me (though I am still a newbie). The object (mapped address) here is allocated from `MmapRegionBuilder` structure and freed by `MmapRegion` 's drop() method, conditionally. A cleaner way (from my understanding of Rust) normally is to allocate objects is in a way that the resource get freed automatically by the drop method, normally unconditionally. It is okay to pass the object from builder to region, like instance of `Mmap` now, but the mapped raw address is not really an end object which should be used like this.
- And yes `VoltaileSlice` could have used `MmapRegion` as well, but `MmapInfo`, which is `Mmap` for non-xen cases, is used in `mmap_unix.rs` as well to write some common code and so I reused the same type in `VolatileSlice` too.

> 4. [6cfe417](https://github.com/rust-vmm/vm-memory/commit/6cfe4178350a09175eab79d54751c8f4f944f174): I understand that this is based on a concern Andreea had on the follow up PR. My main concern is the same as the previous two commits, that we start introducing too many layers of abstraction here. The structures added here are a new layer on top of MmapRegionBuilder. The diff of this commit touches mostly on test code, so I think what happened at some point in the past is that someone did not want to use the builder in test code, so they added the tuple-based construction methods as shortcuts. Any actual new information that is used during construction of MmapRegions should be integrated with the MmapRegionBuilder (e.g. the two new Xen fields from the follow up PR should become #[cfg(xen)] fields in the build, with associated setters).

Yeah, this is exactly what I have done for `MmapRegionBuilder`.

> Outside of test code, the builder should always be used to construct mmap regions.

Hmm, not really I guess. `MmapRegionBuilder` is a Unix only concept. The common interfaces provided by this crate (in `mmap.rs`) are `GuestRegionMmap` and `GuestMemoryMmap` instead and all the end users should use them. For example `MmapRegionBuilder` isn't there for Windows based implementation. If you see the `vhost` crate too, it uses these two interfaces only and not the builder. And so we required new structures for the parameters of these two structures, which is what was suggested by Andreea. Which eventually gets used in `mmap_unix.rs` as well.

> Sorry if this is a lot (or if I got some things wrong - cross-referencing with the other PR was difficult at time).

I am fine with whatever changes I need to make to get over with this work. I just want to get over this soon :)

> The main themes in my feedback are keeping hierarchies flat, and keeping the changes to the public API for non-xen users minimal.

Sure.

> Before I do another of this PR/the follow up, could you try addressing @andreeaflorescu's [comment](https://github.com/rust-vmm/vm-memory/pull/231#pullrequestreview-1449145413) in the other PR about minimizing the scope of cfg blocks? I think that will make review a lot easier, and should bring down the size of the diffs quite a bit too. Thanks!

Hmm, I already did a lot of that and the diff is really small now comparatively (from where I started).

Since there is a lot of code to review here, I would really prefer to get this pull request merged before we start reviewing the other one, which introduces Xen based stuff. I understand that you had to look at the other pull requests too, to get an insight on how things are getting used eventually and it is not super easy.

My idea behind this separate pull request was to try and make the review process easier and shorter and merge code that is already reviewed, instead of waiting for the rest of it to be ready as well.

I have repushed this branch now after removing the two commits and making the bitmap change you suggested.

Thanks.

vireshk:
> > 2. [4d73f17](https://github.com/rust-vmm/vm-memory/commit/4d73f17e97763b108d5ab958d49ba4466ea44594): With that, the number of call-sites should stay at 1 I think, and these functions can be re-inlined, so this commit isn't necessary.
> 
> It may look messy without that, but I am okay with dropping it. Dropped.

Now that I have rebased [xen/mmapv2](https://github.com/rust-vmm/vm-memory/pull/231) over the new changes (and pushed), it doesn't look that messy afterall. :)

roypat:
> > 3. [f3a9fc5](https://github.com/rust-vmm/vm-memory/commit/f3a9fc57133b5419ea6a9264ffe4c18535cd08ab): The changes in this commit are not motivated in the commit message.
> 
> I am not sure I understood that.

I meant that the commit message only explains _what_ the commit is doing, but not why. Got it now though, thanks for the explanation!

> > I'm not sure why this is required.
> 
> There are a couple of reasons for that:
> 
> - One is that this `struct Mmap` is used by mmap_xen.rs file (renamed as `MmapUnix` there) as well, for all the xen variants.
> 
>  - Also the way it is currently implemented doesn't look very clean or Rust-ish to me (though I am still a newbie). The object (mapped address) here is allocated from `MmapRegionBuilder` structure and freed by `MmapRegion` 's drop() method, conditionally. A cleaner way (from my understanding of Rust) normally is to allocate objects is in a way that the resource get freed automatically by the drop method, normally unconditionally. It is okay to pass the object from builder to region, like instance of `Mmap` now, but the mapped raw address is not really an end object which should be used like this.

Ah, I think this is actually a pretty common pattern in rust. All the structures that take ownership over raw file descriptors (e.g. File::from_raw_fd) come to mind :)

>   - And yes `VoltaileSlice` could have used `MmapRegion` as well, but `MmapInfo`, which is `Mmap` for non-xen cases, is used in `mmap_unix.rs` as well to write some common code and so I reused the same type in `VolatileSlice` too.

Okay, but for unix, we don't need anything extra in `VolatileSlice` - the additional field should be `#[cfg(xen)]`, and just contain the xen specific type. And then at this point, since only the Xen specific type uses this, it's fine to inline its few fields, I think.

> > 4. [6cfe417](https://github.com/rust-vmm/vm-memory/commit/6cfe4178350a09175eab79d54751c8f4f944f174): I understand that this is based on a concern Andreea had on the follow up PR. My main concern is the same as the previous two commits, that we start introducing too many layers of abstraction here. The structures added here are a new layer on top of MmapRegionBuilder. The diff of this commit touches mostly on test code, so I think what happened at some point in the past is that someone did not want to use the builder in test code, so they added the tuple-based construction methods as shortcuts. Any actual new information that is used during construction of MmapRegions should be integrated with the MmapRegionBuilder (e.g. the two new Xen fields from the follow up PR should become #[cfg(xen)] fields in the build, with associated setters).
> 
> Yeah, this is exactly what I have done for `MmapRegionBuilder`.
> 
> > Outside of test code, the builder should always be used to construct mmap regions.
> 
> Hmm, not really I guess. `MmapRegionBuilder` is a Unix only concept. The common interfaces provided by this crate (in `mmap.rs`) are `GuestRegionMmap` and `GuestMemoryMmap` instead and all the end users should use them. For example `MmapRegionBuilder` isn't there for Windows based implementation. If you see the `vhost` crate too, it uses these two interfaces only and not the builder. And so we required new structures for the parameters of these two structures, which is what was suggested by Andreea. Which eventually gets used in `mmap_unix.rs` as well.

Mh, yeah, right now it is, but it doesnt have to stay that way. Instead of adding shared abstraction on top of it, we can just extend the builder to cover both xen and unix, with setters/fields being appropriately cfg'd and then maybe different .build methods. Because right now, what used to be `MmapRegionBuilder::build() -> libc::map()` turned into `MmapRegionBuilder::build() -> MmapRegionBuilder::mmap() -> MmapInfo::with_checks() -> Mmap::new() -> libc::map()` on unix, and `MmapRegionBuilder::build() -> MmapRegionBuilder::mmap() [but a different one] -> MmapXen::mmap() -> MmapXenTrait::mmap() -> MmapXenGrant::mmap_range() -> Mmap::new() -> libc::mmap()` on xen. The cognitive complexity of this call stack is too much, especially given how similarly called everything is :(. And then `MmapRange` adds another layer on top of all of this (and the on-the-fly mapping in VolatileSlice does not even use this specific call stack, so I'm struggling a bit to see where this is used - if we don't use this for the mapping in the VolatileSlice, then why are we making this change?). Intuitively, I'd say 1 intermediary max to get from `build` to `libc::mmap` is something we should strive for here. If this means a little bit of code will be duplicated then that's fine. 

> > Sorry if this is a lot (or if I got some things wrong - cross-referencing with the other PR was difficult at time).
> 
> I am fine with whatever changes I need to make to get over with this work. I just want to get over this soon :)
> 
> > The main themes in my feedback are keeping hierarchies flat, and keeping the changes to the public API for non-xen users minimal.
> 
> Sure.
> 
> > Before I do another of this PR/the follow up, could you try addressing @andreeaflorescu's [comment](https://github.com/rust-vmm/vm-memory/pull/231#pullrequestreview-1449145413) in the other PR about minimizing the scope of cfg blocks? I think that will make review a lot easier, and should bring down the size of the diffs quite a bit too. Thanks!
> 
> Hmm, I already did a lot of that and the diff is really small now comparatively (from where I started).
> 
> Since there is a lot of code to review here, I would really prefer to get this pull request merged before we start reviewing the other one, which introduces Xen based stuff. I understand that you had to look at the other pull requests too, to get an insight on how things are getting used eventually and it is not super easy.

Right, yeah, this was my main problem - I needed to cross-ref between two git checkouts, which was a bit tedious at times xP

> My idea behind this separate pull request was to try and make the review process easier and shorter and merge code that is already reviewed, instead of waiting for the rest of it to be ready as well.
> 
> I have repushed this branch now after removing the two commits and making the bitmap change you suggested.
> 
> Thanks.

vireshk:
> > * And yes `VoltaileSlice` could have used `MmapRegion` as well, but `MmapInfo`, which is `Mmap` for non-xen cases, is used in `mmap_unix.rs` as well to write some common code and so I reused the same type in `VolatileSlice` too.
> 
> Okay, but for unix, we don't need anything extra in `VolatileSlice` - the additional field should be `#[cfg(xen)]`, and just contain the xen specific type. And then at this point, since only the Xen specific type uses this, it's fine to inline its few fields, I think.

For unix we used to have `PhantomData<&'a T>`, which we can avoid now with a single generic field here. Also since this is set as `Option`, it is set to `None` for the non-xen case.

For xen, I need a new parameter to all the `::with_bitmap()` methods. Making this field available for just Xen, would mean a lot of duplicated code for these functions, unless I am missing something here.

vireshk:
> > Hmm, not really I guess. `MmapRegionBuilder` is a Unix only concept. The common interfaces provided by this crate (in `mmap.rs`) are `GuestRegionMmap` and `GuestMemoryMmap` instead and all the end users should use them.
>
> Mh, yeah, right now it is, but it doesnt have to stay that way. Instead of adding shared abstraction on top of it, we can just extend the builder to cover both xen and unix, with setters/fields being appropriately cfg'd and then maybe different .build methods.

This is already done in my other patchset. We have a method named `with_mmap_info()`, which is used to do that exactly.

The problem is that we need something for `GuestRegionMmap` and `GuestMemoryMmap` as well and that's where the Range based new structures come in as the parameters list became really big, which was suggested to me by Andreea and @Ablu to reduce the Xen specific code btw, which eventually looks great to me at least.

> Because right now, what used to be `MmapRegionBuilder::build() -> libc::map()` turned into `MmapRegionBuilder::build() -> MmapRegionBuilder::mmap() -> MmapInfo::with_checks() -> Mmap::new() -> libc::map()` on unix,

I can merge the `Mmapinfo::with_checks() and Mmap::new()`, they are kept separate to make the code efficient and skip an `if` comparison though.

The new structure `Mmap` came out of earlier reviews only, and dropping that would mean going back to duplicating code now.

> and `MmapRegionBuilder::build() -> MmapRegionBuilder::mmap() [but a different one] -> MmapXen::mmap() -> MmapXenTrait::mmap() -> MmapXenGrant::mmap_range() -> Mmap::new() -> libc::mmap()` on xen.

Right.

> The cognitive complexity of this call stack is too much, especially given how similarly called everything is :(.

This is bound to happen when the code is required to do more and handle more complex cases, I don't how this can be avoided. I am sure that initially we won't have wanted to keep separate files like `mmap.rs` and `mmap_unix.rs`, but we ended up doing that as we wanted to support different operating systems, etc. These abstractions will come in if want to support more with less code.

> And then `MmapRange` adds another layer on top of all of this (and the on-the-fly mapping in VolatileSlice does not even use this specific call stack, so I'm struggling a bit to see where this is used -

This call stack provides extra information, i.e. `mmap_flags` and `mmap_data`, which is used by Xen mapping routines, while the mapping is made. For foreign mapping, it is done right from this call stack as `MmapRegionBuilder::build()` ends up mapping the memory. For grant mapping though, the information is saved in the structure and used later when the real mapping happens via `VolatileSlice`. This information is also used to know which mapping type Xen supports, foreign or grant, and so it is used at `build()` all the time.

> if we don't use this for the mapping in the VolatileSlice, then why are we making this change?).

We do use it :)

> Intuitively, I'd say 1 intermediary max to get from `build` to `libc::mmap` is something we should strive for here. If this means a little bit of code will be duplicated then that's fine.

This new strucuture is really useful to free the resources (`munmap()`) when the instance goes out of scope. This is required for each Xen mapping model as well as Unix. I can remove this strucure if you like and start duplicating the same code, but that would be going backwards and then I will have to implement a `drop()` trait for many structures which is avoided right now with this.

vireshk:
Replaced by: https://github.com/rust-vmm/vm-memory/pull/241

PR #241, opened 2023-06-23
Xen/mmapv3

vireshk:
Add support for Xen specific memory mapping

Xen allows two memory mapping mechanisms:
- Foreign mapping: entire guest space is mapped at once.
- Grant mapping: guest controls what backend is allowed to map.

Both these mechanisms required additional mapping support, for example
an ioctl() needs to be issued (with arguments passed from backend),
along with mmap().

Also for grant memory mapping, the mapping may or may not be done in
advance and may be required to be done on-demand.

Add memory mapping support for both these cases, under a new feature
"xen".

We can't use the existing `as_ptr()` API, as the memory may not be mapped
already. Mark the existing API available to non-xen implementations only.

Also add new implementations, ptr_guard() and ptr_guard_mut(), for both xen
and non-xen platforms, which can be used to map memory and then get a
pointer to it, work on it, and unmap the memory automatically.

This also adds few custom buidkite testcases to test Xen based implementation.

Signed-off-by: Viresh Kumar [viresh.kumar@linaro.org](mailto:viresh.kumar@linaro.org)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,33 @@ The detailed design of the `vm-memory` crate can be found [here](DESIGN.md).
 - Arch: x86, AMD64, ARM64
 - OS: Linux/Unix/Windows
 
+### Xen support
+
+Supporting Xen requires special handling while mapping the guest memory and
+hence a separate feature is provided in the crate: `xen`. Mapping the guest
+memory for Xen requires an `ioctl()` to be issued along with `mmap()` for the
+memory area. The arguments for the `ioctl()` are received via the `vhost-user`
+protocol's memory region area.
+
+Xen allows two different mapping models: `Foreign` and `Grant`.
+
+In `Foreign` mapping model, the entire guest address space is mapped at once, in
+advance. In `Grant` mapping model, the memory for few regions, like those
+representing the virtqueues, is mapped in advance. The rest of the memory
+regions are mapped (partially) only while accessing the buffers and the same is
+immediately deallocated after the buffer is accessed. Hence, special handling
+for the same in `VolatileMemory.rs`.
+
+In order to still support standard Unix memory regions, for special regions and
+testing, the Xen specific implementation here allows a third mapping type:
+`MmapXenFlags::UNIX`. This performs standard Unix memory mapping and the same is
+used for all tests in this crate.
+
+It was decided by the `rust-vmm` maintainers to keep the interface simple and
+build the crate for either standard Unix memory mapping or Xen, and not both.
+
+Xen is only supported for Unix platforms.
+
 ## Usage
 
 Add `vm-memory` as a dependency in `Cargo.toml`
--- a/src/volatile_memory.rs
+++ b/src/volatile_memory.rs
@@ -37,6 +37,12 @@ use crate::atomic_integer::AtomicInteger;
 use crate::bitmap::{Bitmap, BitmapSlice, BS};
 use crate::{AtomicAccess, ByteValued, Bytes};
 
+#[cfg(all(feature = "backend-mmap", feature = "xen", unix))]
+use crate::mmap_xen::{MmapXen as MmapInfo, MmapXenSlice};
+
+#[cfg(not(feature = "xen"))]
+type MmapInfo = std::marker::PhantomData<()>;
+
 use copy_slice_impl::{copy_from_volatile_slice, copy_to_volatile_slice};
 
 /// `VolatileMemory` related errors.
@@ -115,7 +121,13 @@ pub trait VolatileMemory {
         let slice = self.get_slice(offset, size_of::<T>())?;
         // SAFETY: This is safe because the pointer is range-checked by get_slice, and
         // the lifetime is the same as self.
-        unsafe { Ok(VolatileRef::with_bitmap(slice.addr, slice.bitmap)) }
+        unsafe {
+            Ok(VolatileRef::with_bitmap(
+                slice.addr,
+                slice.bitmap,
+                slice.mmap,
+            ))
+        }
     }
 
     /// Returns a [`VolatileArrayRef`](struct.VolatileArrayRef.html) of `n` elements starting at
@@ -136,7 +148,14 @@ pub trait VolatileMemory {
         let slice = self.get_slice(offset, nbytes as usize)?;
         // SAFETY: This is safe because the pointer is range-checked by get_slice, and
         // the lifetime is the same as self.
-        unsafe { Ok(VolatileArrayRef::with_bitmap(slice.addr, n, slice.bitmap)) }
+        unsafe {
+            Ok(VolatileArrayRef::with_bitmap(
+                slice.addr,
+                n,
+                slice.bitmap,
+                slice.mmap,
+            ))
+        }
     }
 
     /// Returns a reference to an instance of `T` at `offset`.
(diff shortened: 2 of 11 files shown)


jiangliu:
Really hope `GrantTable` is available for kvm too:)

README.md
```diff
@@ -21,6 +21,33 @@ The detailed design of the `vm-memory` crate can be found [here](DESIGN.md).
...
+immediately deallocated after the buffer is accessed. Hence, special handling
+for the same in `VolatileMemory.rs`.
+
+In order to still support standard Unix memory regions, for special regions and
+testing, the Xen specific implementation here allows a third mapping type:
+`MmapXenFlags::UNIX`. This performs standard Unix memory mapping and the same is
+used for all tests in this crate.
+
+It was decided by the `rust-vmm` maintainers to keep the interface simple and
+build the crate for either standard Unix memory mapping or Xen, and not both.
```
jiangliu, line 47: +1

vireshk:
> Really hope `GrantTable` is available for kvm too:)

:)

I haven't really followed the pKvm discussions, since it is going to be about secure guests, they will surely need something similar. Either something exactly like Xen, or they may fix the mappable areas somewhere (which may have penalties in form on no ZeroCopy).

src/volatile_memory.rs
```diff
@@ -219,13 +235,57 @@ impl<'a> From<&'a mut [u8]> for VolatileSlice<'a, ()> {
...
+            };
+
+            let slice = match mmap {
+                Some(mmap) => mmap.mmap_slice(addr, prot, len),
+                None => MmapSlice::raw(addr),
+            };
+
+            (slice.addr(), slice)
+        };
+
```
roypat, line 267: I think we can simplify a bit more here. Because really, the "guard" is `MmapSlice`: Its what unmaps the memory when it gets dropped. So this can simply be a utility function that returns a `MmapSlice` and then at the guarded callsites, we use `MmapSlice::addr` instead of `AsPtrGuard::as_ptr`.

src/volatile_memory.rs
```diff
@@ -949,27 +1060,36 @@ where
...
     #[inline(always)]
     pub fn store(&self, v: T) {
-        // SAFETY: Safe because we checked the address and size when creating this VolatileRef.
-        unsafe { write_volatile(self.addr, Packed::<T>(v)) };
-        self.bitmap.mark_dirty(0, size_of::<T>())
+        self.ptr_guard_mut(|ptr| {
+            // SAFETY: Safe because we checked the address and size when creating this VolatileRef.
+            unsafe { write_volatile(ptr as *mut Packed<T>, Packed::<T>(v)) };
+            self.bitmap.mark_dirty(0, self.len())
+        })
```
roypat, line 1067: Nice!

roypat (commented):
Looks good! I didn't really look too deeply into the mmap_xen.rs code because I'm lacking Xen expertise, but everything else I'm happy with approving! It's just missing a Changelog entry for what's added (Xen support, the new API for accessing pointers to volatile slices is all I think?) and we should be good to go :)

src/volatile_memory.rs
vireshk: > I think we can simplify a bit more here. Because really, the "guard" is `MmapSlice`: Its what unmaps the memory when it gets dropped. So this can simply be a utility function that returns a `MmapSlice` and then at the guarded callsites, we use `MmapSlice::addr` instead of `AsPtrGuard::as_ptr`.

I am not sure I understood it properly. `MmapSlice` is a Xen thing, and doesn't exist otherwise. And so the call sites do need to use another abstraction, like `AsPtrGuard`. What am I missing ?

vireshk:
> Looks good! I didn't really look too deeply into the mmap_xen.rs code because I'm lacking Xen expertise, but everything else I'm happy with approving! It's just missing a Changelog entry for what's added (Xen support, the new API for accessing pointers to volatile slices is all I think?) and we should be good to go :)

Thanks. I have updated the changelog and also added two documentation examples (for foreign and grant mappings) in Xen's `MmapRegion::from_range()` helper.

src/volatile_memory.rs
Ablu: @vireshk asked me to test this with virtiofsd again and I fear that this won't work nicely. One use case where virtiofsd needs the raw pointer is in order to issue `writev`/`readv` syscalls. These essentially receive a list of pointers + lens and will do scattered/gather read/writes. So given a list of slices, there needs to be a way to turn it into a list of pointers. With the previously sketched guard API that was doable by just keeping a list of guards. Now, that no longer works without nesting calls very deeply (and the code probably would not be pretty).

I understand that this was done because you had some concerns about the old "guard" not really guarding something. However, I am not sure how much safer the new proposal is. One could still, for example, accidentally access outside of the allowed length (or call .ptr_guard twice, nested and create a duplicated map). Do you have an example where you think the FnOnce solution is a better compared to the guard? Personally, I mostly understand it such that both cases still have unsafe inside of them but the FnOnce just makes the unsafe verification _slightly_ easier?

Of course, we could provide extra API to provide (and map) a list of pointers. But then we are introducing further new API variants at the cost of the just slightly easier unsafety verification. Also, where would we introduce that API on?

FWIW: The virtiofsd use-case could _maybe_ also be solvable by just providing a Read/Write impl somewhere that supports the Rust vectored read/write API. _Maybe_ something like https://github.com/rust-vmm/vm-virtio/pull/33 could be combined with the vectored read/write variants of `Read`/`Write` to solve virtiofsd needs. But the `writev`/`readv` examples are probably good hints that there are similar scenarios hidden somewhere that would require multiple slices to be mapped at the same time.

roypat: Right, I see. I was mainly suggesting this because it looked simpler than the guard approach, but I agree that unless vm-memory provides better read(v)/write(v) abstractions this would make that quite complicated (fwiw I do have some sketches for something like that, but that's not useful at this time ofc). I kinda wanted to avoid the need for a `MappedVolatileSlice` thingy but I suppose there's no way around here.

roypat:
> > Looks good! I didn't really look too deeply into the mmap_xen.rs code because I'm lacking Xen expertise, but everything else I'm happy with approving! It's just missing a Changelog entry for what's added (Xen support, the new API for accessing pointers to volatile slices is all I think?) and we should be good to go :)
> 
> Thanks. I have updated the changelog and also added two documentation examples (for foreign and grant mappings) in Xen's `MmapRegion::from_range()` helper.

I think you might have forgotten to push the changes? :sweat_smile:

vireshk:
> I think you might have forgotten to push the changes? sweat_smile

Maybe not :)

https://github.com/rust-vmm/vm-memory/pull/241/commits/65f2c8dc481fd1b3af4778025e35df9579b71c14

roypat:
> > I think you might have forgotten to push the changes? sweat_smile
> 
> Maybe not :)
> 
> [65f2c8d](https://github.com/rust-vmm/vm-memory/commit/65f2c8dc481fd1b3af4778025e35df9579b71c14)

ah whoops missed that, so sorry! Am still not seeing the changelog though :thinking:

vireshk:
> ah whoops missed that, so sorry! Am still not seeing the changelog though thinking

My mistake. I though you were asking me to update the Pull request's commit message :(, only now realized you meant CHANGELOG.md :)

Done now.

src/volatile_memory.rs
vireshk: Okay, now that everyone agrees, I have moved back to the earlier version in somewhat better shape with less code. Hope it looks fine now.

roypat (approved):

Ablu:
/me is still investigating some test failures with virtiofsd. May very likely be in my setup. Will report once I figured it out.

Ablu:
Tested OK with virtiofsd + small patch [1] on Xen [2] with grant mappings [3] on 6.1 series kernel [4].

Errors were my own.

[1] https://gitlab.com/ablu/virtiofsd/-/commit/9466396862dde30f3b692c6e9eacbc89044eb874
[2] https://github.com/vireshk/xen/commit/acb25f01ba63877eb960aed711fb08ed667d8314
[3] `virtio = [ "type=virtio,device26,transport=mmio,grant_usage=1" ]`
[4] 6.1.24-yocto-standard

Ablu (approved):

vireshk:
> Tested OK with virtiofsd + small patch [1] on Xen [2] with grant mappings [3] on 6.1 series kernel [4].

Thanks Erik.

PR #244, opened 2023-07-03
volatile_memory: Deprecate .as_ptr() interfaces

vireshk:
The `.as_ptr()` interface can't be used with the `xen` feature. Mark them deprecated with a note to use the new interfaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [Unreleased]
+
+### Deprecated
+
+- [[#244]](https://github.com/rust-vmm/vm-memory/pull/241) Deprecate volatile
+  memory's `as_ptr()` interfaces. The new interfaces to be used instead are:
+  `ptr_guard()` and `ptr_guard_mut()`.
+
 ## [v0.12.0]
 
 ### Added
--- a/src/volatile_memory.rs
+++ b/src/volatile_memory.rs
@@ -355,6 +355,10 @@ impl<'a, B: BitmapSlice> VolatileSlice<'a, B> {
     /// Returns a pointer to the beginning of the slice. Mutable accesses performed
     /// using the resulting pointer are not automatically accounted for by the dirty bitmap
     /// tracking functionality.
+    #[deprecated(
+        since = "0.12.1",
+        note = "Use `.ptr_guard()` or `.ptr_guard_mut()` instead"
+    )]
     #[cfg(not(all(feature = "xen", unix)))]
     pub fn as_ptr(&self) -> *mut u8 {
         self.addr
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## [Unreleased]
+
+### Deprecated
+
+- [[#244]](https://github.com/rust-vmm/vm-memory/pull/241) Deprecate volatile
+  memory's `as_ptr()` interfaces.
```
Ablu, line 8: Maybe include a short note of using `.ptr_guard` instead?

src/volatile_memory.rs
```diff
@@ -1028,6 +1032,10 @@ where
     /// Returns a pointer to the underlying memory. Mutable accesses performed
     /// using the resulting pointer are not automatically accounted for by the dirty bitmap
     /// tracking functionality.
+    #[deprecated(
+        since = "0.12.0",
```
Ablu, line 1036: Shouldn't this be 0.13.0 then (as that would be the next release)? Or do you plan to include this in 0.12.0 (in that case the changelog entry is wrong)?

It may slightly confuse people that use 0.12.0 (which does not include the warning), but see the deprecation in the docs.

vireshk: Can we make this part of v0.12.1 instead ? Since this doesn't change the functionality at all ?

CHANGELOG.md
vireshk: Well, the commit itself with the deprecation message talks about all that. Not sure how worth it would be adding that here in `CHANGELOG.md`. I am fine without whatever people feel about it though.

src/volatile_memory.rs
Ablu: I thought it may annoy people since a new lint may show up. But https://doc.rust-lang.org/cargo/reference/semver.html#minor-introducing-new-lints suggests a minor bump. So 0.12.1 sounds correct :)

Ablu (approved):

jiangliu (approved):

andreeaflorescu:
This breaks the vm-virtio/vsock build. How can the new interface be used in that case?
See: https://github.com/rust-vmm/vm-virtio/pull/254
CC: @vireshk @roypat

vireshk:
> This breaks the vm-virtio/vsock build. How can the new interface be used in that case? See: [rust-vmm/vm-virtio#254](https://github.com/rust-vmm/vm-virtio/pull/254) CC: @vireshk @roypat

Hi @andreeaflorescu 

Sorry for the late reply. I was out of office for sometime. Is this issue resolved or you still need someone to look into this ?

vireshk:
> Sorry for the late reply. I was out of office for sometime. Is this issue resolved or you still need someone to look into this ?

Okay, fixed it.

https://github.com/rust-vmm/vm-virtio/pull/257

PR #245, opened 2023-07-13
mmap_xen: Don't drop the FileOffset while in use

vireshk:
The drop implementation of `struct MmapXenGrant` issues an Ioctl call, which uses the file descriptor.

Currently the FileOffset is getting dropped before that, and results in "Bad File Descriptor" error from the Ioctl. This commit fixes it by keeping a copy of the same in `struct MmapXenGrant`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,19 @@
 # Changelog
+## [Unreleased]
+
+### Added
+
+### Changed
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


roypat (approved):
Looks reasonable to me. `FileOffset` wraps the `File` into an `Arc`, so cloning indeed prevents dropping

jiangliu (approved):

PR #247, opened 2023-07-24
[RFC] `ReadVolatile` and `WriteVolatile` traits for efficient data transfer from/to guest memory

roypat:
### Summary of the PR

With https://github.com/rust-vmm/vm-memory/pull/217, various `read_from` and `write_to` methods across vm-memory
incurred a performance penalty due to an additional copy of all data
read and written. This is due to these functions operating on arbitrary
Read/Write implementations, which cannot generally be assumed to respect
the volatile semantics of guest memory. Additionally, they required
taking slices to guest memory, which was undefined behavior.

This patch series fixes those short-comings by introducing a pair of new
traits, `ReadVolatile` and `WriteVolatile` which are identical to the
standard library's `Read` and `Write` traits, with the difference being
that they operate on `VolatileSlice`s instead of `&mut [u8]`. Using
these traits will therefore not incur the performance penalty of https://github.com/rust-vmm/vm-memory/pull/217.
Additionally, this will make using the vm-memory library easier for
users familiar with std::io, as the APIs will be more familiar (in
particular, the unintuitive "swapping" of reader/writer src and dst of
the old `read_from`/`write_to` APIs is resolved).

Please consider this as a not-completely-polished proposal for which I am seeking feedback. The code is based on what we wrote in firecracker to mitigate the performance impact of upgrading to 0.10.0 as a result of #217.

### Open Questions
- this patch series does not attempt to implement vectored I/O via `readv`/`writev`. I think there is definitely a usecase for this, since both firecracker and virtio-fsd use vectored I/O in various places. It would be nice to upstream something similar to https://gitlab.com/virtio-fs/virtiofsd/-/blob/main/src/file_traits.rs to `ReadVolatile`/`WriteVolatile`. Maybe someone from that team could chime in?
- I'd honestly love to cut down the entire jungle of `read` and `write` functions in this crate, but considered that change to be somewhat out of scope here (with this PR only aiming to address the inability to transfer data efficiently between a file descriptor and guest memory). I feel however that deprecating all `read`/`write` functions would greatly simplify this library at the cost of users having to call `get_slice()`. Maybe via turning the `Bytes` trait into something that represents "`self` can be turned into a `VolatileSlice`", as `read` -> `read_volatile`, `read_slice` -> `read_exact_volatile`, `write` -> `write_volatile` and `write_slice` -> `write_all_volatile`. 

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -52,7 +52,9 @@ use std::sync::Arc;
 use crate::address::{Address, AddressValue};
 use crate::bitmap::{Bitmap, BS, MS};
 use crate::bytes::{AtomicAccess, Bytes};
+use crate::io::{ReadVolatile, WriteVolatile};
 use crate::volatile_memory::{self, VolatileSlice};
+use crate::GuestMemoryError;
 
 static MAX_ACCESS_CHUNK: usize = 4096;
 
@@ -665,6 +667,146 @@ pub trait GuestMemory {
         }
     }
 
+    /// Reads up to `count` bytes from an object and writes them into guest memory at `addr`.
+    ///
+    /// Returns the number of bytes written into guest memory.
+    ///
+    /// # Arguments
+    /// * `addr` - Begin writing at this address.
+    /// * `src` - Copy from `src` into the container.
+    /// * `count` - Copy `count` bytes from `src` into the container.
+    ///
+    /// # Examples
+    ///
+    /// * Read bytes from /dev/urandom (uses the `backend-mmap` feature)
+    ///
+    /// ```
+    /// # #[cfg(feature = "backend-mmap")]
+    /// # {
+    /// # use vm_memory::{Address, GuestMemory, Bytes, GuestAddress, GuestMemoryMmap};
+    /// # use std::fs::File;
+    /// # use std::path::Path;
+    /// #
+    /// # let start_addr = GuestAddress(0x1000);
+    /// # let gm = GuestMemoryMmap::<()>::from_ranges(&vec![(start_addr, 0x400)])
+    /// #    .expect("Could not create guest memory");
+    /// # let addr = GuestAddress(0x1010);
+    /// # let mut file = if cfg!(unix) {
+    /// let mut file = File::open(Path::new("/dev/urandom")).expect("Could not open /dev/urandom");
+    /// #   file
+    /// # } else {
+    /// #   File::open(Path::new("c:\\Windows\\system32\\ntoskrnl.exe"))
+    /// #       .expect("Could not open c:\\Windows\\system32\\ntoskrnl.exe")
+    /// # };
+    ///
+    /// gm.read_volatile_from(addr, &mut file, 128)
+    ///     .expect("Could not read from /dev/urandom into guest memory");
+    ///
+    /// let read_addr = addr.checked_add(8).expect("Could not compute read address");
+    /// let rand_val: u32 = gm
+    ///     .read_obj(read_addr)
+    ///     .expect("Could not read u32 val from /dev/urandom");
+    /// # }
+    /// ```
+    fn read_volatile_from<F>(&self, addr: GuestAddress, src: &mut F, count: usize) -> Result<usize>
+    where
+        F: ReadVolatile,
+    {
+        self.try_access(count, addr, |offset, len, caddr, region| -> Result<usize> {
+            // Check if something bad happened before doing unsafe things.
+            assert!(offset <= count);
+
+            let len = std::cmp::min(len, MAX_ACCESS_CHUNK);
+
+            let mut vslice = region.get_slice(caddr, len)?;
+
+            src.read_volatile(&mut vslice)
+                .map_err(GuestMemoryError::from)
+        })
+    }
+
+    /// Reads up to `count` bytes from the container at `addr` and writes them it into guest memory.
+    ///
+    /// Returns the number of bytes written into guest memory.
+    ///
+    /// # Arguments
+    /// * `addr` - Begin reading from this address.
+    /// * `dst` - Copy from guest memory to `dst`.
+    /// * `count` - Copy `count` bytes from guest memory to `dst`.
+    fn write_volatile_to<F>(&self, addr: GuestAddress, dst: &mut F, count: usize) -> Result<usize>
+    where
+        F: WriteVolatile,
+    {
+        self.try_access(count, addr, |offset, len, caddr, region| -> Result<usize> {
+            // Check if something bad happened before doing unsafe things.
+            assert!(offset <= count);
+
+            let len = std::cmp::min(len, MAX_ACCESS_CHUNK);
+            let vslice = region.get_slice(caddr, len)?;
+
+            // For a non-RAM region, reading could have side effects, so we
+            // must use write_all().
+            dst.write_all_volatile(&vslice)?;
+
+            Ok(len)
+        })
+    }
+
+    /// Reads exactly `count` bytes from an object and writes them into guest memory at `addr`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if `count` bytes couldn't have been copied from `src` to guest memory.
+    /// Part of the data may have been copied nevertheless.
+    ///
+    /// # Arguments
+    /// * `addr` - Begin writing at this address.
+    /// * `src` - Copy from `src` into guest memory.
+    /// * `count` - Copy exactly `count` bytes from `src` into guest memory.
+    fn read_exact_volatile_from<F>(
+        &self,
+        addr: GuestAddress,
+        src: &mut F,
+        count: usize,
+    ) -> Result<()>
+    where
+        F: ReadVolatile,
+    {
+        let res = self.read_volatile_from(addr, src, count)?;
+        if res != count {
+            return Err(Error::PartialBuffer {
+                expected: count,
+                completed: res,
+            });
+        }
+        Ok(())
+    }
+
+    /// Reads exactly `count` bytes from the container at `addr` and writes them into guest memory.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if `count` bytes couldn't have been copied from guest memory to `dst`.
+    /// Part of the data may have been copied nevertheless.
+    ///
+    /// # Arguments
+    /// * `addr` - Begin reading from this address.
+    /// * `dst` - Copy from guest memory to `dst`.
+    /// * `count` - Copy exactly `count` bytes from guest memory to `dst`.
+    fn write_all_volatile_to<F>(&self, addr: GuestAddress, dst: &mut F, count: usize) -> Result<()>
+    where
+        F: WriteVolatile,
+    {
+        let res = self.write_volatile_to(addr, dst, count)?;
+        if res != count {
+            return Err(Error::PartialBuffer {
+                expected: count,
+                completed: res,
+            });
+        }
+        Ok(())
+    }
+
     /// Get the host virtual address corresponding to the guest address.
     ///
     /// Some [`GuestMemory`](trait.GuestMemory.html) implementations, like `GuestMemoryMmap`,
@@ -985,8 +1127,6 @@ mod tests {
     #[cfg(feature = "backend-mmap")]
     use crate::GuestAddress;
     #[cfg(feature = "backend-mmap")]
-    use std::io::Cursor;
-    #[cfg(feature = "backend-mmap")]
     use std::time::{Duration, Instant};
 
     use vmm_sys_util::tempfile::TempFile;
@@ -1026,7 +1166,7 @@ mod tests {
         let count: usize = 0x20;
         assert_eq!(
             0x20_usize,
-            mem.read_from(offset, &mut Cursor::new(&image), count)
+            mem.read_volatile_from(offset, &mut image.as_slice(), count)
                 .unwrap()
         );
     }
@@ -1181,19 +1321,24 @@ mod tests {
         assert!(mem.write_obj(obj, addr).is_ok());
         assert!(mem.read_obj::<ZeroSizedStruct>(addr).is_ok());
 
-        assert_eq!(mem.read_from(addr, &mut Cursor::new(&image), 0).unwrap(), 0);
-
-        assert!(mem
-            .read_exact_from(addr, &mut Cursor::new(&image), 0)
-            .is_ok());
-
         assert_eq!(
-            mem.write_to(addr, &mut Cursor::new(&mut image), 0).unwrap(),
+            mem.read_volatile_from(addr, &mut image.as_slice(), 0)
+                .unwrap(),
             0
         );
 
         assert!(mem
-            .write_all_to(addr, &mut Cursor::new(&mut image), 0)
+            .read_exact_volatile_from(addr, &mut image.as_slice(), 0)
+            .is_ok());
+
+        assert_eq!(
+            mem.write_volatile_to(addr, &mut image.as_mut_slice(), 0)
+                .unwrap(),
+            0
+        );
+
+        assert!(mem
+            .write_all_volatile_to(addr, &mut image.as_mut_slice(), 0)
             .is_ok());
     }
 
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -856,6 +856,7 @@ impl<T: GuestMemory + ?Sized> Bytes<GuestAddress> for T {
     where
         F: Read,
     {
+        #[allow(deprecated)] // this function itself is deprecated
         let res = self.read_from(addr, src, count)?;
         if res != count {
             return Err(Error::PartialBuffer {
@@ -949,6 +950,7 @@ impl<T: GuestMemory + ?Sized> Bytes<GuestAddress> for T {
     where
         F: Write,
     {
+        #[allow(deprecated)] // this function itself is deprecated
         let res = self.write_to(addr, dst, count)?;
         if res != count {
             return Err(Error::PartialBuffer {
--- a/src/io.rs
+++ b/src/io.rs
@@ -1,0 +1,272 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//! Module containing versions of the standard library's [`Read`] and [`Write`] traits compatible
+//! with volatile memory accesses.
+
+use crate::bitmap::BitmapSlice;
+use crate::volatile_memory::copy_slice_impl::{copy_from_volatile_slice, copy_to_volatile_slice};
+use crate::{VolatileMemoryError, VolatileSlice};
+use std::io::ErrorKind;
+use std::os::fd::AsRawFd;
+
+/// A version of the standard library's [`Read`] trait that operates on volatile memory instead of
+/// slices
+///
+/// This trait is needed as rust slices (`&[u8]` and `&mut [u8]`) cannot be used when operating on
+/// guest memory [1].
+///
+/// [1]: https://github.com/rust-vmm/vm-memory/pull/217
+pub trait ReadVolatile {
+    /// Tries to read some bytes into the given [`VolatileSlice`] buffer, returning how many bytes
+    /// were read.
+    ///
+    /// The behavior of implementations should be identical to [`Read::read`]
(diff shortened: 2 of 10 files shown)


src/guest_memory.rs
```diff
@@ -665,6 +667,146 @@ pub trait GuestMemory {
...
+    /// gm.read_from_volatile(addr, &mut file, 128)
+    ///     .expect("Could not read from /dev/urandom into guest memory");
+    ///
+    /// let read_addr = addr.checked_add(8).expect("Could not compute read address");
+    /// let rand_val: u32 = gm
+    ///     .read_obj(read_addr)
+    ///     .expect("Could not read u32 val from /dev/urandom");
+    /// # }
+    /// ```
+    fn read_from_volatile<F>(&self, addr: GuestAddress, src: &mut F, count: usize) -> Result<usize>
```
Ablu, line 711: This is not really reading from a volatile, but _into_ a volatile, right?

src/io.rs
```diff
@@ -0,0 +1,443 @@
...
+    fn read_volatile<B: BitmapSlice>(
+        &mut self,
+        buf: &mut VolatileSlice<B>,
+    ) -> Result<usize, VolatileMemoryError>;
+
+    /// Tries to fill the given [`VolatileSlice`] buffer by reading from `self` returning an error
+    /// if insufficient bytes could be read.
+    ///
+    /// The default implementation is identical to that of [`Read::read_exact`]
+    fn read_exact_volatile<B: BitmapSlice>(
```
Ablu, line 33: Who would use this? This sounds like some code always expects things to fit into a single VolatileSlice. We got quite a few cases where consumers of this library are not considering that descriptors can be sliced up arbitrarily and do not necessarily just have one slice that fits what one would expect.

Ablu (commented):
We definitely need some kind of abstraction like this! I am just worrying a bit about the right level to introduce it... This implements the traits on `VolatileSlice`. I wonder whether it would not be more useful to implement this on a higher level? We only have such crazy number of read/write calls because we lack read/write abstraction on a descriptor chain. Also, if we would implement these traits on a descriptor chain level of abstraction, vectored read and writes could be implemented fairly straight forward.

So: Do we _need_ to have this on this abstraction layer? Or do we only have it here because we lack a higher-level abstraction layer? My feeling is that we _will_ need an abstraction on descriptor-chain level. Mostly for preventing people to use these low-level APIs incorrectly, but it may also allow to solve two problems at once: Zero-overhead access and a lot simpler and less error-prone API. https://github.com/rust-vmm/vm-virtio/pull/33 proposed such an API and may be a natural abstraction level to add these "0-copy" variant traits.

Do we have a use-case that we cannot solve with such a higher-level API?

roypat:
> We definitely need some kind of abstraction like this! I am just worrying a bit about the right level to introduce it... This implements the traits on `VolatileSlice`. I wonder whether it would not be more useful to implement this on a higher level? We only have such crazy number of read/write calls because we lack read/write abstraction on a descriptor chain. Also, if we would implement these traits on a descriptor chain level of abstraction, vectored read and writes could be implemented fairly straight forward.
> 
> So: Do we _need_ to have this on this abstraction layer? Or do we only have it here because we lack a higher-level abstraction layer? My feeling is that we _will_ need an abstraction on descriptor-chain level. Mostly for preventing people to use these low-level APIs incorrectly, but it may also allow to solve two problems at once: Zero-overhead access and a lot simpler and less error-prone API. [rust-vmm/vm-virtio#33](https://github.com/rust-vmm/vm-virtio/pull/33) proposed such an API and may be a natural abstraction level to add these "0-copy" variant traits.
> 
> Do we have a use-case that we cannot solve with such a higher-level API?

APIs for transfering data/from guest memory are currently used outside of virtio stacks. For example, linux-loader uses these to copy the kernel to guest memory [[1]], and firecracker uses them to save guest memory to disk for snapshot purposes [[2]]. Neither of these could be solved if these traits only lived in vm-virtio. So in that sense, I think it makes sense to have some Read/Write abstractions at the vm-memory level. Having them at `VolatileSlice` level made sense to me, since it is essentially the guest memory version of `&[u8]`. Do you see any case where not having the functions from this PR at a high abstraction level will pose a problem? 

In a sense, this PR does not add anything new - it simply provides replacements for the APIs that got broken by #217, and deprecates the broken ones. 

[1]: https://github.com/rust-vmm/linux-loader/blob/617ebf72d1e02023f391731c1678c6e7f8c265fd/src/loader/x86_64/elf/mod.rs#L284
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/memory_snapshot.rs#L49

As for vectored reads and writes, I definitely agree that they are a lot more naturally expressible at the virtio descriptor chain level (we actually have an implementation similar to what you describe in firecracker [[3]]). I actually was not able to come up with a nice solution here for how to offer them at the vm-memory level without restricting what the virtio level is able to do (for example, if we only offer `write_vectored_volatile` with `&[VolatileSlice]` as an argument, then `write_vectored_volatile` would have to allocate a vector of `iovec` as `VolatileSlice` is not ABI compatible with `writev`. But the caller will already have allocated a vector of `VolatileSlice` from a descriptor chain, whereas if we directly did vectored writes at the virtio level, only a single allocation would be needed. But a vm-memory-level API that takes &[iovec] is also not very useful because then the caller would be responsible for updating for example the dirty bitmap, and the function would have to be unsafe, so its just `libc::writev` in a trenchcoat). Hence the "RFC" status of this PR. If there's no way to offer them at the vm-memory level, then we simply won't do it. However, I think it'd be great if we could share some of these primitives between the different virtio stack implementations out there, so if we can express it outside of the virtio-framework, then vm-memory would be the place to do that. 

As for why we have all the read/write, I think its mainly convenience. People did not want to convert everything to VolatileSlice before being able to call .read/.write, so everything that could be converted to VolatileSlice instead gained a .read/.write method that does the conversion and then delegates. Over time it just got kinda out of control because the semantics/naming between the various functions got out of sync.

[3]: https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/devices/virtio/iovec.rs

src/mmap.rs
```diff
@@ -1186,8 +1190,8 @@ mod tests {
                 assert_eq!(value, 0x0090_5a4d);
             }
 
-            let mut sink = Vec::new();
-            gm.write_all_to(addr, &mut sink, mem::size_of::<u32>())
+            let mut sink = vec![0; mem::size_of::<u32>()];
+            gm.write_all_to_volatile(addr, &mut sink.as_mut_slice(), mem::size_of::<u32>())
```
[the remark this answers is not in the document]
roypat: Ahh, true, keep forgetting that exists, thanks!

src/mmap.rs
```diff
@@ -273,6 +273,7 @@ impl<B: Bitmap> Bytes<MemoryRegionAddress> for GuestRegionMmap<B> {
         F: Read,
     {
         let maddr = addr.raw_value() as usize;
+        #[allow(deprecated)] // function itself is deprecated
```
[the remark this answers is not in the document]
roypat: The function is an implementation of a deprecated trait function. Rust doesn't allow you to re-tag it as deprecated though (it'll give a "needless `#[deprecated]` warnings), which is slightly annoying >.>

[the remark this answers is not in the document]
Ablu: Ah, makes sense!

Ablu (commented):
> For example, linux-loader uses these to copy the kernel to guest memory [[1](https://github.com/rust-vmm/linux-loader/blob/617ebf72d1e02023f391731c1678c6e7f8c265fd/src/loader/x86_64/elf/mod.rs#L284)], and firecracker uses them to save guest memory to disk for snapshot purposes [[2](https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/memory_snapshot.rs#L49)].

Makes sense. That were the use-cases that I did not have on my radar :).

Will re-review with that perspective tomorrow.

src/volatile_memory.rs
```diff
@@ -2044,16 +2043,15 @@ mod tests {
         } else {
             File::open(Path::new("c:\\Windows\\system32\\ntoskrnl.exe")).unwrap()
         };
-        assert!(s.read_exact_from(2, &mut file, size_of::<u32>()).is_err());
-        assert!(s
-            .read_exact_from(core::usize::MAX, &mut file, size_of::<u32>())
-            .is_err());
```
Ablu, line 2064: This test case was dropped, right? Is it no longer needed?

src/volatile_memory.rs
```diff
@@ -2062,13 +2060,12 @@ mod tests {
             assert_eq!(value, 0x0090_5a4d);
         }
 
-        let mut sink = Vec::new();
-        assert!(s.write_all_to(1, &mut sink, size_of::<u32>()).is_ok());
-        assert!(s.write_all_to(2, &mut sink, size_of::<u32>()).is_err());
-        assert!(s
-            .write_all_to(core::usize::MAX, &mut sink, size_of::<u32>())
-            .is_err());
-        format!("{:?}", s.write_all_to(2, &mut sink, size_of::<u32>()));
```
Ablu, line 2085: Same here?

Ablu (commented):
nothing else jumped to my eye :)

src/volatile_memory.rs
roypat: Yes, these test cases were checking the out-of-bounds checks in read_exact_from, however using `ReadVolatile::read_exact_volatile` does not take offset/count parameters and instead expects the `VolatileSlice` being passed to have the correct offset/size. This moves the bounds checks into the `.get_slice()` call, which is a separate function with separate tests (e.g. its impossible to replicate this test scenario with the new APIs)

src/volatile_memory.rs
roypat: Yup, same as above

src/io.rs
```diff
@@ -279,3 +279,165 @@ impl ReadVolatile for &[u8] {
...
+    use vmm_sys_util::tempfile::TempFile;
+
+    #[test]
+    fn test_read_volatile() {
+        let test_cases = [
+            (vec![1u8, 2], [1u8, 2, 0, 0, 0]),
+            (vec![1, 2, 3, 4], [1, 2, 3, 4, 0]),
+            // ensure we don't have a buffer overrun
+            (vec![5, 6, 7, 8, 9], [5, 6, 7, 8, 0]),
+        ];
```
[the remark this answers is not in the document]
roypat: Done!

src/guest_memory.rs
roypat: fixed :)

jiangliu:
How about also implementing `ReadVolatile`/`WriteVolatile` for `OwnedFd`/`BorrowedFd`?

roypat:
> How about also implementing `ReadVolatile`/`WriteVolatile` for `OwnedFd`/`BorrowedFd`?

Ahh, yes, I forgot about that! Originally I wanted to implement these traits for all `T: AsRawFd` (which would cover `OwnedFd`/`BorrowedFd`, and is slightly more general than `T: Into<Ownedfd>`). There was some issue of overlapping impls on firecracker's side, but I just found a different solution on our side for that, so let me update this patch to have the general impls (should also make the patch slightly smaller!)

roypat:
> How about also implementing `ReadVolatile`/`WriteVolatile` for `OwnedFd`/`BorrowedFd`? Otherwise LGTM!

Done, please have another look :)

jiangliu (approved):

jiangliu (dismissed):

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 **/*.rs.bk
 Cargo.lock
+.idea
```
JonathanWoollett-Light, line 4: I don't think this should be included.

src/guest_memory.rs
```diff
@@ -665,6 +667,146 @@ pub trait GuestMemory {
...
     }
 
+    /// Reads up to `count` bytes from an object and writes them into guest memory at `addr`.
+    ///
+    /// Returns the number of bytes written into guest memory.
+    ///
+    /// # Arguments
+    /// * `addr` - Begin writing at this address.
+    /// * `src` - Copy from `src` into the container.
+    /// * `count` - Copy `count` bytes from `src` into the container.
```
JonathanWoollett-Light, line 677: To match [std](https://doc.rust-lang.org/std/index.html)::[ptr](https://doc.rust-lang.org/std/ptr/index.html)::[write_volatile](https://doc.rust-lang.org/std/ptr/fn.write_volatile.html#) I would prefer the arguments be named:
```suggestion
    /// # Arguments
    /// * `dest` - Begin writing at this address.
    /// * `src` - Begin reading at this address.
    /// * `count` - Copy this many bytes.
```

src/guest_memory.rs
```diff
@@ -665,6 +667,146 @@ pub trait GuestMemory {
...
+    /// let read_addr = addr.checked_add(8).expect("Could not compute read address");
+    /// let rand_val: u32 = gm
+    ///     .read_obj(read_addr)
+    ///     .expect("Could not read u32 val from /dev/urandom");
+    /// # }
+    /// ```
+    fn read_volatile_from<F>(&self, addr: GuestAddress, src: &mut F, count: usize) -> Result<usize>
+    where
+        F: ReadVolatile,
+    {
```
JonathanWoollett-Light, line 714: Why do we use `read_volatile_from` over `read_volatile`?

src/guest_memory.rs
```diff
@@ -665,6 +667,146 @@ pub trait GuestMemory {
...
+        self.try_access(count, addr, |offset, len, caddr, region| -> Result<usize> {
+            // Check if something bad happened before doing unsafe things.
+            assert!(offset <= count);
```
JonathanWoollett-Light, line 717: Shouldn't this be return as a `Result::Err` variant?

src/guest_memory.rs
```diff
@@ -665,6 +667,146 @@ pub trait GuestMemory {
...
+        F: WriteVolatile,
+    {
+        let res = self.write_volatile_to(addr, dst, count)?;
+        if res != count {
+            return Err(Error::PartialBuffer {
+                expected: count,
+                completed: res,
+            });
+        }
+        Ok(())
```
JonathanWoollett-Light, line 807: Nit: I find it simply to interpret like this (doing the negation is a little mental overhead when reading it quickly):
```suggestion
        let res = self.write_volatile_to(addr, dst, count)?;
        if res == count {
            Ok(())
        }
        else {
            Err(Error::PartialBuffer {
                expected: count,
                completed: res,
            })
        }
```

src/guest_memory.rs
```diff
@@ -1179,19 +1321,24 @@ mod tests {
...
+        assert_eq!(
+            mem.read_volatile_from(addr, &mut image.as_slice(), 0)
+                .unwrap(),
+            0
+        );
 
         assert!(mem
-            .read_exact_from(addr, &mut Cursor::new(&image), 0)
+            .read_exact_volatile_from(addr, &mut image.as_slice(), 0)
             .is_ok());
```
JonathanWoollett-Light, line 1332: Nit: It is more rigorous to match results exactly (also applies to all other cases where you've edited the assertion)
```suggestion
        asser_eq!(mem
            .read_exact_from(addr, &mut Cursor::new(&image), 0)
            .read_exact_volatile_from(addr, &mut image.as_slice(), 0),
            Ok(0)
        );
```

src/guest_memory.rs
roypat: Yeah, this assert has already confused me. I think it should be a `debug_assert!`, since `try_access` should really already verify this (and indeed correctly return an `Err`). I kept it like this because the logic is copied from the deprecated functions.

src/guest_memory.rs
roypat: This function is written to exactly mirror the corresponding standard library function, so I'd keep it as is

.gitignore
roypat: I've dropped this commit from this PR

jiangliu (approved):

src/guest_memory.rs
JonathanWoollett-Light: This is fair.

JonathanWoollett-Light (approved):

Ablu:
@roypat: The Rust `Write` trait has a `flush()`. That this does not seem to provide. Also, trying to port the `virtio-console` from `vm-virtio` to this would benefit from some kind of Rust Read/Write <> VolatileRead/Write adapter... Would that be in scope of vm-memory? How would you propose to handle the update of vm-memory in vm-virtio?

roypat:
> @roypat: The Rust `Write` trait has a `flush()`. That this does not seem to provide. Also, trying to port the `virtio-console` from `vm-virtio` to this would benefit from some kind of Rust Read/Write <> VolatileRead/Write adapter... Would that be in scope of vm-memory? How would you propose to handle the update of vm-memory in vm-virtio?

Mhh, for `flush`, I think we can add that to `VolatileWrite` and in the implementations just have it forward it to `Write`'s flush implementation. I don't think there's anything volatile-memory-specific that needs to happen there. I actually also noticed that right now this is lacking a `VolatileRead`/`VolatileWrite` implementation for `Cursor` that is needed for some linux-loader tests >.>.

As for adapters for `Read`/`Write`, I'd like to avoid that because it comes with all the same problems that the old `read_*` and `write_*` family of functions had: Since `Read` and `Write` dont know about volatile memory semantics, every operation would need to be buffered in hypervisor memory before it can be forwarded to the traits. For `vm-console`, I think the only other thing that would be needed to avoid this is a `VolatileWrite` implementation for `std::io::Stdout`, which could be generated using the `RawFd` macros in this PR. 

Tbh, I probably was a bit overzealous with deprecating the old functions here, if you want I'll be happy to just revert that part of this PR and make a patch release.

Ablu:
> Mhh, for `flush`, I think we can add that to `VolatileWrite` and in the implementations just have it forward it to `Write`'s flush implementation. I don't think there's anything volatile-memory-specific that needs to happen there. I actually also noticed that right now this is lacking a `VolatileRead`/`VolatileWrite` implementation for `Cursor` that is needed for some linux-loader tests >.>.

Right... it probably does not make a lot of sense to support these...

> As for adapters for `Read`/`Write`, I'd like to avoid that because it comes with all the same problems that the old `read_*` and `write_*` family of functions had: Since `Read` and `Write` dont know about volatile memory semantics, every operation would need to be buffered in hypervisor memory before it can be forwarded to the traits. For `vm-console`, I think the only other thing that would be needed to avoid this is a `VolatileWrite` implementation for `std::io::Stdout`, which could be generated using the `RawFd` macros in this PR.

Do you maybe want to draft a PR for that? Currently this is blocking the updates to latest vm-memory for a lot of crates.

> Tbh, I probably was a bit overzealous with deprecating the old functions here, if you want I'll be happy to just revert that part of this PR and make a patch release.

Well, eventually we want to deprecate them anyway I guess? So we might as well just do it now and fix the code that depends on the old API.

roypat:
> > Mhh, for `flush`, I think we can add that to `VolatileWrite` and in the implementations just have it forward it to `Write`'s flush implementation. I don't think there's anything volatile-memory-specific that needs to happen there. I actually also noticed that right now this is lacking a `VolatileRead`/`VolatileWrite` implementation for `Cursor` that is needed for some linux-loader tests >.>.
> 
> Right... it probably does not make a lot of sense to support these...

I think it might make sense, to avoid people having to do `Write + VolatileWrite` bounds, maybe?

> > As for adapters for `Read`/`Write`, I'd like to avoid that because it comes with all the same problems that the old `read_*` and `write_*` family of functions had: Since `Read` and `Write` dont know about volatile memory semantics, every operation would need to be buffered in hypervisor memory before it can be forwarded to the traits. For `vm-console`, I think the only other thing that would be needed to avoid this is a `VolatileWrite` implementation for `std::io::Stdout`, which could be generated using the `RawFd` macros in this PR.
> 
> Do you maybe want to draft a PR for that? Currently this is blocking the updates to latest vm-memory for a lot of crates.

Please have a look at https://github.com/rust-vmm/vm-memory/pull/256 :)

> > Tbh, I probably was a bit overzealous with deprecating the old functions here, if you want I'll be happy to just revert that part of this PR and make a patch release.
> 
> Well, eventually we want to deprecate them anyway I guess? So we might as well just do it now and fix the code that depends on the old API.

Sure!

PR #255, opened 2023-10-03
mmap_xen: Update to bitflags 2.4.0

vireshk:
Clippy gives this warning currently:

error: &-masking with zero
   --> src/mmap_xen.rs:433:1
    |
433 | / bitflags! {
434 | |     /// Flags for the Xen mmap message.
435 | |     pub struct MmapXenFlags: u32 {
436 | |         /// Standard Unix memory mapping.
...   |
446 | |     }
447 | | }
    | |_^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#bad_bit_mask
    = note: `#[deny(clippy::bad_bit_mask)]` on by default
    = note: this error originates in the macro `__impl_bitflags` which comes from the expansion of the macro `bitflags` (in Nightly builds, run with -Z macro-backtrace for more info)

Fix it.

### Summary of the PR

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/mmap_xen.rs
+++ b/src/mmap_xen.rs
@@ -432,6 +432,7 @@ impl Drop for MmapUnix {
 // Bit mask for the vhost-user xen mmap message.
 bitflags! {
     /// Flags for the Xen mmap message.
+    #[derive(Copy, Clone, Debug)]
     pub struct MmapXenFlags: u32 {
         /// Standard Unix memory mapping.
         const UNIX = 0x0;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ xen = ["backend-mmap", "bitflags", "vmm-sys-util"]
 [dependencies]
 libc = "0.2.39"
 arc-swap = { version = "1.0.0", optional = true }
-bitflags = { version = "1.0", optional = true }
+bitflags = { version = "2.4.0", optional = true }
 thiserror = "1.0.40"
 vmm-sys-util = { version = "0.11.0", optional = true }
 


stefano-garzarella:
I don't know if it's a real bug or not, but if you want to keep the previous values, you can switch to bitflags 2.* and you shouldn't see the warning anymore.

stefano-garzarella:
Not for this PR, but I was looking at the code, and I'm a bit suspicious of:
```
    /// Is standard Unix memory.
    pub fn is_unix(&self) -> bool {
        self.bits() == Self::UNIX.bits()
    }

```

Can a Unix memory have `NO_ADVANCE_MAP` flag set?
If it is the case `is_unix()` will return `false`, so we may need a mask there.

vireshk:
> Not for this PR, but I was looking at the code, and I'm a bit suspicious of:
> 
> ```
>     /// Is standard Unix memory.
>     pub fn is_unix(&self) -> bool {
>         self.bits() == Self::UNIX.bits()
>     }
> ```
> 
> Can a Unix memory have `NO_ADVANCE_MAP` flag set? If it is the case `is_unix()` will return `false`, so we may need a mask there.

Unix can't have `NO_ADVANCE_MAP` flag. Only Grant mapping can have that.

stefano-garzarella:
> Unix can't have `NO_ADVANCE_MAP` flag. Only Grant mapping can have that.

Oh I see, I was confused by the negative flag!

src/mmap_xen.rs
```diff
@@ -432,6 +432,7 @@ impl Drop for MmapUnix {
 // Bit mask for the vhost-user xen mmap message.
 bitflags! {
     /// Flags for the Xen mmap message.
+    #[derive(Clone, Debug)]
```
stefano-garzarella, line 435: We can derive also `Copy` and avoid the next change.

stefano-garzarella (approved):

roypat (approved):

PR #258, opened 2023-10-31
Use non-zero page size

JonathanWoollett-Light:
This avoids potential divide by zero error.

Fixes https://github.com/rust-vmm/vm-memory/issues/268

### Summary of the PR

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/bitmap/backend/atomic_bitmap.rs
+++ b/src/bitmap/backend/atomic_bitmap.rs
@@ -3,6 +3,7 @@
 
 //! Bitmap backend implementation based on atomic integers.
 
+use std::num::NonZeroUsize;
 use std::sync::atomic::{AtomicU64, Ordering};
 
 use crate::bitmap::{Bitmap, RefSlice, WithBitmapSlice};
@@ -17,14 +18,14 @@ use crate::mmap::NewBitmap;
 pub struct AtomicBitmap {
     map: Vec<AtomicU64>,
     size: usize,
-    page_size: usize,
+    page_size: NonZeroUsize,
 }
 
 #[allow(clippy::len_without_is_empty)]
 impl AtomicBitmap {
     /// Create a new bitmap of `byte_size`, with one bit per page. This is effectively
     /// rounded up, and we get a new vector of the next multiple of 64 bigger than `bit_size`.
-    pub fn new(byte_size: usize, page_size: usize) -> Self {
+    pub fn new(byte_size: usize, page_size: NonZeroUsize) -> Self {
         let mut num_pages = byte_size / page_size;
         if byte_size % page_size > 0 {
             num_pages += 1;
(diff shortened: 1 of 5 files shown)


src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -225,7 +226,7 @@ mod tests {
 
     #[test]
     fn test_bitmap_impl() {
-        let b = AtomicBitmap::new(0x2000, 128);
+        let b = AtomicBitmap::new(0x2000, DEFAULT_PAGE_SIZE);
```
Ablu, line 229: nit: The helper variable probably could have been introduced in a separate commit.

Ablu (commented):
This is breaking API and should have a CHANGELOG entry. While I see no problem with the approach, I wonder whether this could be reasoned a bit better in the commit message.

My immediate questions are: Where did this cause major problems? Why is making the user go through the NonZeroUsize conversion worth it?

Is this the only place where we deal with page_sizes? :thinking: Should we convert them all?

JonathanWoollett-Light:
> My immediate questions are: Where did this cause major problems? Why is making the user go through the NonZeroUsize conversion worth it?

Without this it is possible to trigger a divide by zero error, see https://github.com/rust-vmm/vm-memory/blob/main/src/bitmap/backend/atomic_bitmap.rs#L28

Ablu:
> > My immediate questions are: Where did this cause major problems? Why is making the user go through the NonZeroUsize conversion worth it?
> 
> Without this it is possible to trigger a divide by zero error, see https://github.com/rust-vmm/vm-memory/blob/main/src/bitmap/backend/atomic_bitmap.rs#L28

Sure! I was just wondering where that became a problem in "practise" :).

JonathanWoollett-Light:
> Sure! I was just wondering where that became a problem in "practise" :).

I understand your point, in practice I cannot see this occurring, but it is possible so it should be mitigated.

src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -136,38 +137,35 @@ impl Bitmap for AtomicBitmap {
...
+            GetSystemInfo(&mut sysinfo as LPSYSTEM_INFO);
+            sysinfo.dwPageSize
+        };
 
         // The `unwrap` is safe to use because the above call should always succeed on the
         // supported platforms, and the size of a page will always fit within a `usize`.
-        AtomicBitmap::new(len, usize::try_from(page_size).unwrap())
+        AtomicBitmap::new(
+            len,
+            NonZeroUsize::try_from(usize::try_from(page_size).unwrap()).unwrap(),
```
jiangliu, line 167: This is not a good sign, we change a divided by zero into panic:)

JonathanWoollett-Light: What do you think of returning `Result` or `Option` from this function with `NonZeroUsize::try_from(usize::try_from(page_size).unwrap())?` ?

Is there another approach you would propose?

Ablu: I think this is a good example why one could argue that this may not be worth the effort... Realistically, people will replace a division by zero (where Rust would panic) with accidental unsafety or manual panics. Not sure what is gained there. The only alternative is forcing Result<> into paths that are usually very unlikely to fail, cluttering the code... And all of that at the cost of breaking existing API and forcing people to update their code.

I got no strong opinion on this, but after still waiting for crates to update after the recent vm-memory API breaks, I am a bit worried...

JonathanWoollett-Light: I understand this, I am not strongly opinionated on pushing for this change.

roypat: > This is not a good sign, we change a divided by zero into panic:)
A divide by zero is already a panic in rust, both in release and debug builds.

I also don't think using `Result` or `Option` here is a good idea, for the reasons @Ablu mentions. Passing in zero for page size is fundamentally wrong, so replacing the implicit "divide by zero" panic with one that's a bit more informative (e.g. "attempt to pass zero for page size") or using a type-system level approach as in this PR would be my preferred approach.

src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -136,38 +137,35 @@ impl Bitmap for AtomicBitmap {
...
-            // It's safe to call this method as the pointer is based on the address
-            // of the previously initialized `sysinfo` object.
-            unsafe { GetSystemInfo(&mut sysinfo as LPSYSTEM_INFO) };
-
-            page_size = sysinfo.dwPageSize;
-        }
+            let mut sysinfo: SYSTEM_INFO = std::mem::zeroed();
+            GetSystemInfo(&mut sysinfo as LPSYSTEM_INFO);
+            sysinfo.dwPageSize
+        };
```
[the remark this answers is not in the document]
JonathanWoollett-Light: I've re-worked it to work more inline with this.

roypat (approved):

src/bitmap/backend/atomic_bitmap.rs
andreeaflorescu: Isn't updating the AtomicBitmap::new to take as a parameter NonZeroUsize also a breaking change? Why is this better than returning a Result?

roypat: It is, yeah. I greatly prefer it over the `Result` approach though, because `NonZeroUsize` already prevents you at compile time from passing in nonsense.

src/bitmap/backend/atomic_bitmap.rs
JonathanWoollett-Light: I think this doesn't warrant an additional commit.

src/bitmap/backend/atomic_bitmap.rs
Ablu: I am still a bit worried that this is a lot of code churn for very little gain... But I have no strong opinion here.

Are other crates within rust-vmm using the changed public API? Should we prepare some PRs to demonstrate how those need to be adjusted?

Ablu: > I think this is outside the scope of this PR. I don't think any PR that introduces a breaking change should also require PR's to all other rust-vmm crates using breaking change.
> 
> I would suggest if we wanted to move towards this pattern we should move towards a more monorepo-like design to support it.

We brought this up during the last rust-vmm sync since the lack of such investigation caused us to get stuck with some state where we were unable to update vm-memory for a couple of months. The echo there was that we should encourage people to submit PRs that demonstrate the scale of changes that are required on the rest of the rust-vmm ecosystem. IMHO it plays a critical role in deciding the impact of breaking changes and helps with timing the release.

jiangliu (approved):

src/bitmap/backend/atomic_bitmap.rs
roypat: I ran `cargo test` on vm-virtio with a `crates-io.patch` directive for vm-memory pointing this PR and everything passed, so I think we are fine here! Maybe that is something we can add to the CI in vm-memory as an optional check?

JonathanWoollett-Light: I've created https://github.com/rust-vmm/vm-memory/issues/271 to track this.

PR #260, opened 2023-11-06
Check `try_access` callback length

JonathanWoollett-Light:
### Summary of the PR

Check the length returned by the callback in `try_access` does not exceed the address range.

Fixes: https://github.com/rust-vmm/vm-memory/issues/269

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -77,6 +77,14 @@ pub enum Error {
     /// Host virtual address not available.
     #[error("Guest memory error: host virtual address not available")]
     HostAddressNotAvailable,
+    /// The length returned by the callback passed to `try_access` is outside the address range.
+    #[error(
+        "The length returned by the callback passed to `try_access` is outside the address range."
+    )]
+    CallbackOutOfRange,
+    /// The address to be read by `try_access` is outside the address range.
+    #[error("The address to be read by `try_access` is outside the address range")]
+    GuestAddressOverflow,
 }
 
 impl From<volatile_memory::Error> for Error {
@@ -646,15 +654,15 @@ pub trait GuestMemory {
                 Ok(0) => return Ok(total),
                 // made some progress
                 Ok(len) => {
-                    total += len;
-                    if total == count {
-                        break;
-                    }
+                    total = match total.checked_add(len) {
+                        Some(x) if x < count => x,
+                        Some(x) if x == count => return Ok(x),
+                        _ => return Err(Error::CallbackOutOfRange),
+                    };
                     cur = match cur.overflowing_add(len as GuestUsize) {
-                        (GuestAddress(0), _) => GuestAddress(0),
-                        (result, false) => result,
-                        (_, true) => panic!("guest address overflow"),
-                    }
+                        (x @ GuestAddress(0), _) | (x, false) => x,
+                        (_, true) => return Err(Error::GuestAddressOverflow),
+                    };
                 }
                 // error happened
                 e => return e,


src/guest_memory.rs
```diff
@@ -646,10 +651,11 @@ pub trait GuestMemory {
...
                 Ok(len) => {
-                    total += len;
-                    if total == count {
-                        break;
-                    }
+                    total = match total.checked_add(len) {
+                        Some(x) if x == count => break,
+                        Some(x) if x < count => x,
+                        _ => return Err(Error::CallbackOutOfRange),
+                    };
```
roypat, line 658: In the case of `x == count`, we still need to update `total` before breaking out of the loop, as `total` will be used to report how many bytes were transferred, and without updating it the final transfer chunk will not be considered. That is why the CI is failing. 
```suggestion
                    match total.checked_add(len) {
                        Some(x) if x <= count => {
                            total = x;
                            if total == count {
                                break
                            }
                        }
                        _ => return Err(Error::CallbackOutOfRange)
                    }
```

roypat (changes requested):

src/guest_memory.rs
```diff
@@ -77,6 +77,11 @@ pub enum Error {
     /// Host virtual address not available.
     #[error("Guest memory error: host virtual address not available")]
     HostAddressNotAvailable,
+    /// The length returned by the callback passed to `try_access` is outside the address range.
+    #[error(
+        "The length returned by the callback passed to `try_access` is outside the address range."
+    )]
+    CallbackOutOfRange,
```
roypat, line 84: Lets reuse `Error::InvalidGuestAddress` here, so that this can be a patch release

JonathanWoollett-Light: Updated to use this approach.

src/guest_memory.rs
```diff
@@ -646,10 +646,15 @@ pub trait GuestMemory {
...
+                    match total.checked_add(len) {
+                        Some(x) if x <= count => {
+                            total = x;
+                            if total == count {
+                                break;
+                            }
+                        },
+                        _ => return Err(Error::InvalidBackendAddress),
+                    };
                     cur = match cur.overflowing_add(len as GuestUsize) {
```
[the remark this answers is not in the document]
Ablu: Hm... Looking at it, it _looks_ like it would panic before or immediately after the above code would panic. Should we maybe write a test for this? We can mock fake regions, right?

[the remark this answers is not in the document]
JonathanWoollett-Light: Ideally it shouldn't ever panic. It should always try to return an error that can be properly handled and propagated.

It should not panic below.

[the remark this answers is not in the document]
Ablu: > Ideally it shouldn't ever panic. It should always try to return an error that can be properly handled and propagated.
> 
> It should not panic below.

Full ACK. But that's why I think both additions / checks should probably be touched. And I think a test would help a lot here.

src/guest_memory.rs
```diff
@@ -646,10 +646,15 @@ pub trait GuestMemory {
...
-                        break;
-                    }
+                    match total.checked_add(len) {
+                        Some(x) if x <= count => {
+                            total = x;
+                            if total == count {
+                                break;
+                            }
+                        },
+                        _ => return Err(Error::InvalidBackendAddress),
```
andreeaflorescu, line 656: I find this slightly awkward since the Backend Address is not actually invalid, but rather the access size. I think the major release problem is giving the wrong incentives in this case. I would prefer us having it as it was initially with a new error type, but I am also fine with disagreeing and committing on this one.

Ablu: While I suggested the error code, I got no opinion on this. Happy with either way.

JonathanWoollett-Light: I've reverted it back to using a new specific error code.

roypat (dismissed):

roypat (approved):

roypat (approved):

src/guest_memory.rs
```diff
@@ -646,15 +654,15 @@ pub trait GuestMemory {
...
+                        _ => return Err(Error::CallbackOutOfRange),
+                    };
                     cur = match cur.overflowing_add(len as GuestUsize) {
-                        (GuestAddress(0), _) => GuestAddress(0),
-                        (result, false) => result,
-                        (_, true) => panic!("guest address overflow"),
-                    }
+                        (x @ GuestAddress(0), _) | (x, false) => x,
+                        (_, true) => return Err(Error::GuestAddressOverflow),
+                    };
```
Ablu, line 665: I think the `overflowing_add` would probably better be a `checked_add` too, but I guess that can be fixed separately.

src/guest_memory.rs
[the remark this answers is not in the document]
JonathanWoollett-Light: I've updated the PR to avoid the panic.

src/guest_memory.rs
```diff
@@ -646,15 +654,15 @@ pub trait GuestMemory {
...
                 Ok(len) => {
-                    total += len;
-                    if total == count {
-                        break;
-                    }
+                    total = match total.checked_add(len) {
+                        Some(x) if x < count => x,
+                        Some(x) if x == count => return Ok(x),
+                        _ => return Err(Error::CallbackOutOfRange),
+                    };
```
Ablu, line 661: What do you think about (untested)?
```suggestion
                    total = total.checked_add(len).ok_or_else(|| Error::CallbackOutOfRange)?;
                    if total == count {
                        break;
                    };
```
I find that slightly easier to tread and it leaves a single (successful) exit point from the function.

src/guest_memory.rs
```diff
@@ -646,15 +654,15 @@ pub trait GuestMemory {
...
                 // made some progress
                 Ok(len) => {
-                    total += len;
-                    if total == count {
-                        break;
-                    }
+                    total = match total.checked_add(len) {
+                        Some(x) if x < count => x,
+                        Some(x) if x == count => return Ok(x),
+                        _ => return Err(Error::CallbackOutOfRange),
```
Ablu, line 660: I would also be fine with returning GuestAddressOverflow, saving one error code. But no real opinion here.

Ablu (approved):
LGTM, with or without nitpick fixed.

src/guest_memory.rs
```diff
@@ -77,6 +77,14 @@ pub enum Error {
...
     #[error("Guest memory error: host virtual address not available")]
     HostAddressNotAvailable,
+    /// The length returned by the callback passed to `try_access` is outside the address range.
+    #[error(
+        "The length returned by the callback passed to `try_access` is outside the address range."
+    )]
+    CallbackOutOfRange,
+    /// The address to be read by `try_access` is outside the address range.
+    #[error("The address to be read by `try_access` is outside the address range")]
+    GuestAddressOverflow,
```
andreeaflorescu, line 87: Following the InvalidGuestAddress variant, these two should also specify which address is invalid. This should make problems easier to debug.

src/guest_memory.rs
andreeaflorescu: What is the reason for using `overflowing_add` instead of `checked_add`? And why would we return x if it overflows to 0. Wouldn't that also be an invalid access?

src/guest_memory.rs
andreeaflorescu: We can use `ok_or` because we're not calling any function, we're using a constant, so no need to use the `ok_or_else`. Otherwise, I also find the code much easier to read with your proposal @Ablu.

andreeaflorescu (commented):
Sorry for the really late review :(

src/guest_memory.rs
JonathanWoollett-Light: These errors occur due to out of bounds addition, so they cannot return the invalid out of bounds value. Should they contain 2 values that added make the out of bounds value? This seems over-complicated.

src/guest_memory.rs
JonathanWoollett-Light: I do not know, I have not investigated all the reasoning behind this. For now I'm preserving previous behavior, I think further improvements can be made outside of this PR.

src/guest_memory.rs
JonathanWoollett-Light: I lightly disagree here on the styling, I think this is a nit so we are okay to merge either way?

roypat: But isnt that just reverting the logic back to what it was before the PR (e.g. without fixing #269)? E.g. we need to return `Error::CallbackOutOfRange` if `total + len > count`, not just if it overflows.

src/guest_memory.rs
andreeaflorescu: I was more thinking about using the `addr` that was passed to try_access. That should be enough for gettung started with debugging. We can add it later if needed.

src/guest_memory.rs
andreeaflorescu: Oh, I didn't see that's how it was before. This should indeed be fixed in a follow-up PR.

src/guest_memory.rs
JonathanWoollett-Light: I think this is a nit that can be in a follow-up, is this okay?

src/guest_memory.rs
Ablu: > But isnt that just reverting the logic back to what it was before the PR (e.g. without fixing #269)? E.g. we need to return `Error::CallbackOutOfRange` if `total + len > count`, not just if it overflows.

Ah... I misread the ticket and was thinking we wanted to avoid a usize overflow here. Makes sense. Though I still think that `if total >= count` would be easier on my brain.

PR #270, opened 2023-11-20
atomic_bitmap: add capability to reset bits range

russell-islam:
Currently, there is no implementation to reset a single bit or bit range for AtomicBitmap struct. This patch adds necessary function to reset a specific bit or a bit range. This use case is needed for Cloud-Hypervisor in managing page cache for Microsoft Hypervisor related to SEV-SNP guest. Rather than implementing a new one , it is helpful to add feature to this struct.

### Summary of the PR

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/bitmap/backend/atomic_bitmap.rs
+++ b/src/bitmap/backend/atomic_bitmap.rs
@@ -61,6 +61,14 @@ impl AtomicBitmap {
     /// is for the page corresponding to `start_addr`, and the last bit that we set corresponds
     /// to address `start_addr + len - 1`.
     pub fn set_addr_range(&self, start_addr: usize, len: usize) {
+        self.set_reset_addr_range(start_addr, len, true);
+    }
+
+    // Set/Reset a range of `len` bytes starting at `start_addr`
+    // reset parameter determines whether bit will be set/reset
+    // if set is true then the range of bits will be set to one,
+    // otherwise zero
+    fn set_reset_addr_range(&self, start_addr: usize, len: usize, set: bool) {
         // Return early in the unlikely event that `len == 0` so the `len - 1` computation
         // below does not underflow.
         if len == 0 {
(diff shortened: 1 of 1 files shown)


src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -79,6 +79,47 @@ impl AtomicBitmap {
...
+        // by ignoring pages at invalid addresses.
+        let last_bit = start_addr.saturating_add(len - 1) / self.page_size;
+        for n in first_bit..=last_bit {
+            if n >= self.size {
+                // Attempts to reset bits beyond the end of the bitmap are simply ignored.
+                break;
+            }
+            self.map[n >> 6].fetch_and(!(1 << (n & 63)), Ordering::SeqCst);
+        }
+    }
```
00xc, line 121: I think you could factor out this code from `set_addr_range` to a function that takes a `bool` and sets the range to the specified value. This would avoid repeating the same code twice.

roypat: Seconded. Otherwise, this looks good

russell-islam: Sorry, I was away to do some other priority. Thanks for the suggestion. I will update the PR accordingly.

russell-islam: @00xc @roypat  Please take another look  if this is okay.

roypat (approved):

jinankjain (approved):
LGTM

src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -76,8 +84,37 @@ impl AtomicBitmap {
...
+                self.map[n >> 6].fetch_or(1 << (n & 63), Ordering::SeqCst);
+            }
+        }
+    }
+
+    /// Reset a range of `len` bytes starting at `start_addr`. The first bit set in the bitmap
+    /// is for the page corresponding to `start_addr`, and the last bit that we set corresponds
+    /// to address `start_addr + len - 1`.
+    pub fn reset_addr_range(&self, start_addr: usize, len: usize) {
+        self.set_reset_addr_range(start_addr, len, true);
```
00xc, line 99: Just a nitpick, but it's a bit counter-intuitive in my opinion that the third argument being set to `true` means that the bits are set to zero, I would expect the opposite. Otherwise LGTM.

russell-islam: @00xc  Fixed now

roypat (approved):

russell-islam:
@00xc  Could you please take another look into this PR?

00xc (approved):

russell-islam:
@bonzini @jiangliu

russell-islam:
Pinging again: @alexandruag @bonzini @jiangliu

JonathanWoollett-Light (approved):

PR #274, opened 2023-12-06
Drop Default requirement on ByteValued

Ablu:
Since zeros are a valid value for any ByteValued type, we can use that instead of relying on a Default implementation (which may not be auto-derivable for some array types).

Further optimizations would be possible if we would have access to a pointer value into the source data. Then we could also use ptr::copy_nonoverlapping and avoid pre-initialization altogether. But changing the API just for that seems a bit excessive.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - [[#266](https://github.com/rust-vmm/vm-memory/pull/266)] Derive `Debug` for several
   types that were missing it.
+- [[#274]] Drop `Default` as requirement for `ByteValued`.
 
 ## [v0.13.1]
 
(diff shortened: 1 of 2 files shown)


roypat (approved):
Yeah, this makes sense to me, I've thought about it before. Maybe, given that there's been issues with implementing this trait, and we have https://github.com/rust-vmm/vm-memory/issues/246, long term it might make sense for us to just use the [zerocopy](https://docs.rs/zerocopy/latest/zerocopy) crate here :thinking: Although one would have to check if it can deal with volatile memory

bonzini (approved):

Issue #225, opened 2023-04-18
Tests fail on 32-bit


plugwash:
In debian we recently updated the vm-memory crate to version 0.10.0, our CI workers ran the tests (with previous versions the tests were not run due to dependency issues) and this revealed serveral failures. All of them appeared to be issues in the testsuite rather than in the crate itself. Some of them affected all 32-bit architectures  we tested on where as others only affected a subset. Three of them seem to just be issues in the tests.

1. "test_bytes" failed to build on all 32-bit architectures tested with "the trait `bytes::AtomicAccess` is not implemented for `u64`", I changed "1u64" to "1usize" to fix this.
2. An out of memory failure in benches/main.rs, this only seems to happen if the "backend-mmap" feature is enabled. I fixed this by using a lower value of REGION_SIZE on 32-bit architectures in benches/mmap/mod.rs
3. check_byte_valued_type assumed that all types tested would either have natural alignment or would have an alignment of 1. This is not the case for u64/i64 on most 32-bit architectures. I fixed the test to take account of this.

There is also a fourth issue but it is one where I will not blame you if you declare it "out of scope". This fourth issue can only be encountered with a patched vmm-sys-util crate, as that crate currently fails to build at all on architectures that do not support AtomicU64.

src/bitmap/backend/atomic_bitmap.rs will not build on architectures that do not have AtomicU64. The most popular 32-bit architectures do have AtomicU64 but some of the less popular ones do not. This code is included when the "backend-atomic" feature is enabled, it is also included when building tests regardless of the feature configuration. 

It seems to me that atomic_bitmap.rs could be made to use AtomicUsize instead of AtomicU64 but writing that modification goes beyond what I'm prepared to do as a distro maintainer.

So for now I prepared a patch that add and adjusts some attributes so it is at least possible to run most of the tests for other feature configurations on architectures that do not have AtomicU64.

https://salsa.debian.org/rust-team/debcargo-conf/-/blob/4b1f227286fa3da2468e98c85bd453c626555c47/src/vm-memory/debian/patches/fix-tests-32-bit.patch

https://salsa.debian.org/rust-team/debcargo-conf/-/blob/4b1f227286fa3da2468e98c85bd453c626555c47/src/vm-memory/debian/patches/skip-tests-atomic-u64.patch

mjt0k:
Now with vm-memory 0.12 release, it fails in a more difficult way on debian, notable on armel (which also don't have AtomicU64):

[...]

PR #275, opened 2023-12-18
Fail build on non 64-bit targets

Ablu:
As discussed on Slack and during sync meetings, we should make it explicit that we only support 64bit. While other architectures may be viable to support, currently we assume that usize == u64 and nobody expressed interest in 32bit support.

Closes: #225

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 ### Changed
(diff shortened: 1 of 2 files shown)


roypat (approved):

roypat:
Nice, thanks for picking this up!

rbradford (approved):

Ablu:
rebased

roypat (approved):

JonathanWoollett-Light (approved):

PR #278, opened 2024-01-04
Remove  `GuestMemoryIterator`, replace with GAT

roypat:
Use an implicit generic associated type to realize GuestMemory::iter(). This removes the pre-GAT workaround used here to get an iterator that borrows values with the same lifetime as the `&self` parameter to `.iter()`.

This is a follow up from a discussion between @Ablu and me on Slack. I was cleaning up my workspace and decided to just put it out here to get some comments. Note that this requires rustc 1.75.0 to compile.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Upcoming version
 
+### Changed
+
+- [[#278](https://github.com/rust-vmm/vm-memory/pull/278) Remove `GuestMemoryIterator` trait,
+  and instead have GuestMemory::iter() return `impl Iterator`.
+
 ## [v0.15.0]
 
 ### Added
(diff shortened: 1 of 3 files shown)


ShadowCurse (approved):

roypat:
@stefano-garzarella what do you think about this? I compiled both vm-virto and vhost workspaces against this without issues, and now that Rust 1.75.0 and impl-trait-in-trait have been out for a while, I think it's a nice simplification we can make

stefano-garzarella:
@roypat I'm not a vm-memory expert, but I like this simplification, especially if it does not require changes from crate users!

stefano-garzarella (approved):

rbradford (approved):

PR #285, opened 2024-03-19
Update documentation

jacksonbrim:
##Summary of the PR
Annotated modules in lib.rs to indicate their feature dependencies such that it is reflected in the docs, enhancing documentation clarity for users on docs.rs. 

##Building Documentation Locally:

To build the documentation locally with the specified feature flags, use the following command:

```
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --all-features --open
```

This command sets the necessary RUSTDOCFLAGS to include conditional compilation flags and features specific to our project, ensuring the generated documentation reflects the same level of detail and conditions as seen on docs.rs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,3 +44,4 @@ codegen-units = 1
 
 [package.metadata.docs.rs]
 all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,10 +15,10 @@
 //! without knowing the implementation details of the VM memory provider. Thus hypervisor
 //! components, such as boot loader, virtual device drivers, virtio backend drivers and vhost
 //! drivers etc, could be shared and reused by multiple hypervisors.
-
 #![deny(clippy::doc_markdown)]
 #![deny(missing_docs)]
 #![deny(missing_debug_implementations)]
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
 // We only support 64bit. Fail build when attempting to build other targets
 #[cfg(not(target_pointer_width = "64"))]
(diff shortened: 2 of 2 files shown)


roypat (commented):
Hi @jacksonbrim,
Thank you for your contribution! We've done a similar thing on some other rust-vmm crates (e.g. https://github.com/rust-vmm/linux-loader/pull/170/commits/26a0d86263882c1713ac1005199c9ff534d70fd8), so it'd be nice to keep the implementation consistent (e.g. use `doc_auto_cfg`) across the crates. Could you have a look at the comments below to see if they make sense? Thanks!

jacksonbrim:
> Hi @jacksonbrim, Thank you for your contribution! We've done a similar thing on some other rust-vmm crates (e.g. [rust-vmm/linux-loader@26a0d86](https://github.com/rust-vmm/linux-loader/commit/26a0d86263882c1713ac1005199c9ff534d70fd8)), so it'd be nice to keep the implementation consistent (e.g. use `doc_auto_cfg`) across the crates. Could you have a look at the comments below to see if they make sense? Thanks!

Yes, that looks much better.

roypat (approved):
Thank you!

Cargo.toml
```diff
@@ -44,3 +44,4 @@ codegen-units = 1
 
 [package.metadata.docs.rs]
 all-features = true
+rustdoc-args = ["--cfg", "docsrs", "--cfg", "feature=\"backend-mmap\"", "--cfg", "feature=\"backend-atomic\""]
```
JonathanWoollett-Light: M understanding is that `rustdoc-args` enables features in the rustdoc binary, so this only needs to be `rustdoc-args = ["--cfg", "docsrs"]`. The features `backend-mmap` etc. will be enabled by `all-features = true`.

JonathanWoollett-Light: I've seen the later commit removes these, I think in this case it would make sense to squash the commits in this PR.

jacksonbrim: Wait, we don't get paid by the commit? 

Squashed it. Thanks for the patience.

roypat (approved):

JonathanWoollett-Light (approved):

PR #296, opened 2024-09-02
Prepare 0.15.0 release

russell-islam:
### Summary of the PR

### Requirements

requirements:

  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@
 
 ## Upcoming version
 
+## [v0.15.0]
+
 ### Added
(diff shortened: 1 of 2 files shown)


roypat:
Thanks! It looks like we might be missing some changelog entries, could you add some for https://github.com/firecracker-microvm/firecracker/pull/4743 and https://github.com/rust-vmm/vm-memory/pull/270 (which is the change we're doing the release for in the first place, haha)

russell-islam:
> Thanks! It looks like we might be missing some changelog entries, could you add some for #285 and #270 (which is the change we're doing the release for in the first place, haha)

Please take another look now.

roypat (approved):
Looks good, thanks!

PR #299, opened 2024-10-01
atomic_bitmap: support enlarging the bitmap

russell-islam:
### Summary of the PR

This patch adds the support the enlarge the atomic bitmap.
 Some use cases in cloud-Hypervisor needs to enlarge the bitmap on-demand.

### Requirements

requirements:

  test.

### DIFF
--- a/src/bitmap/backend/atomic_bitmap.rs
+++ b/src/bitmap/backend/atomic_bitmap.rs
@@ -18,6 +18,7 @@ use crate::mmap::NewBitmap;
 pub struct AtomicBitmap {
     map: Vec<AtomicU64>,
     size: usize,
+    byte_size: usize,
     page_size: NonZeroUsize,
 }
 
@@ -33,6 +34,7 @@ impl AtomicBitmap {
         AtomicBitmap {
             map,
             size: num_pages,
+            byte_size,
             page_size,
         }
     }
@@ -117,6 +119,11 @@ impl AtomicBitmap {
         self.size
     }
 
+    /// Get the size in bytes i.e how many bytes the bitmap can represent, one bit per page.
+    pub fn byte_size(&self) -> usize {
+        self.byte_size
+    }
+
     /// Atomically get and reset the dirty page bitmap.
     pub fn get_and_reset(&self) -> Vec<u64> {
         self.map
@@ -144,6 +151,7 @@ impl Clone for AtomicBitmap {
         AtomicBitmap {
             map,
             size: self.size,
+            byte_size: self.byte_size,
             page_size: self.page_size,
         }
     }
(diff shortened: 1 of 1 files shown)


src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -37,6 +37,15 @@ impl AtomicBitmap {
         }
     }
 
+    /// Enlarge the vector with additionsl byte size, with one bit per page
+    /// Reserve the existing bits and filling out the new space with default.
```
roypat, line 41: nit: 
```suggestion
    /// Enlarge this bitmap with enough bits to track `additional_size` additional bytes at page granularity.
    /// New bits are initialized to zero.
```

russell-islam: Thank you. Fixed now.

roypat (approved):

src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -37,6 +37,15 @@ impl AtomicBitmap {
         }
     }
 
+    /// Enlarge this bitmap with enough bits to track `additional_size` additional bytes at page granularity.
+    /// New bits are initialized to zero.
+    pub fn enlarge(&mut self, additional_size: usize) {
```
rbradford, line 42: It's an interesting choice to make it the size to increase rather than the new size - does that mean the caller of the API has to track the size (or can it be obtained from this API?)

rbradford: Maybe add a `::size()` to allow access to the current size?

russell-islam: This API assumes the caller wants to enlarge the bitmap where new size may mean t reduce the size of the bitmap. The struct does not give the size in bytes but rather it give the len where len is the total size in bits. Could you please clarify if anything needed more here?

rbradford: You could generate an error if the size was smaller. My point is that you don't want the caller to have to track what they created the bitmap with originally or what the last enlargement they made was. Either you need to provide an API call so that the caller can work out the difference themselves or have it take the new value.

russell-islam:
@roypat @rbradford  Please take another look to see the changes make sense now!!

src/bitmap/backend/atomic_bitmap.rs
```diff
@@ -117,6 +128,11 @@ impl AtomicBitmap {
         self.size
     }
 
+    /// Get the size in bytes i.e how many bytes the bitmap can represent, one bit per page sage.
```
rbradford, line 131: Typo - last word in this comment?

rbradford (commented):
I think this is a good change.

rbradford (approved):

roypat (approved):

PR #307, opened 2024-11-12
Cleanup of deprecated functions + Behavioral Clarifications

roypat:
### Summary of the PR

This PR aim to clean up the API surface of the vm-memory crate in a way that is as unintrusive as possible. It does this in two ways
- Remove the myriad of deprecated functions we have accumulated over the years, and
- Improve documentation to capture some of the more quirky edge cases some functions have (completely ignoring the question of whether some of these should be changed for now)

Additionally, it fixes up a goof I made when I introduced the `ReadVolatile` and `WriteVolatile` traits in #247, where I put the `read_volatile_from` and friends funcions into the wrong trait. When it comes to how to implement these traits for `GuestMemoryMmap` and `VolatileSlice`, I've deferred to how the old `Read` and `Write` based functions were implemented for them.

I've compiled the vm-virtio workspace and linux-loader against this. All tests pass in both, only linux-loader needs a tiny change to import the `Bytes` trait now.

If one squints at the code a bit, one might ask whether all the `read`/`read_slice` functions in the `Bytes` trait could be default functions, implemented in terms of the volatile functions (since `&mut [u8]` implements the volatile access traits). I played around with this, but there are subtle functional and performance changes involved in that (although it would be a great simplification), which is why I'm holding off on doing anything there until a later PR.

### Requirements

requirements:

  test.


TimePrinciple:
With this change, the next release should be `0.17.0` I guess 🤔

roypat:
> With this change, the next release should be `0.17.0` I guess 🤔

Yup!

src/volatile_memory.rs
```diff
@@ -1011,19 +1011,7 @@ impl<B: BitmapSlice> VolatileMemory for VolatileSlice<'_, B> {
...
-            // the lifetime is the same as self.
-            unsafe {
-                VolatileSlice::with_bitmap(
-                    self.addr.add(offset),
-                    count,
-                    self.bitmap.slice_at(offset),
-                    self.mmap,
-                )
-            },
-        )
```
TimePrinciple, line 1015: If that's the case, would it be better to use a `macro` to expand code here instead of calling one another 🤔

src/guest_memory.rs
```diff
@@ -681,10 +681,7 @@ pub trait GuestMemory {
     where
         F: ReadVolatile,
     {
-        self.try_access(count, addr, |offset, len, caddr, region| -> Result<usize> {
-            // Check if something bad happened before doing unsafe things.
-            assert!(offset <= count);
-
```
TimePrinciple, line 687: How about completely remove the `offset` parameter, and change `try_access` accordingly.

In my view the only place might need the assertion is before the first execution of `f(total, len as usize, start, region)` in `try_access`, but since `total` defaults to 0 and `len` is an `usize`, the assertion there is pointless. The rest cases are well guarded in `Ok(len)` match arm inside `try_access`.

TimePrinciple (commented):
Really good work, some thoughts 😉

src/guest_memory.rs
roypat: The `offset` parameter is used in the `read` and `write` implementations in `Bytes<GuestAddress> for T: GuestMemory` :/

Well, technically, we cannot rely on the actual behavior of `try_access`, since it's a trait function, and some arbitrary implementation of `GuestMemory` could overwrite it with some absolute nonsense. The main point for me here is that even in that case, it will not cause unsoundness, because there's no unsafe code here, and all bounds are properly checked further down the stack.

src/volatile_memory.rs
roypat: I'd prefer keeping the duplicated code over a macro if we think that we don't want to have another layer of indirection here (which is a fair point, the indirection in this crate are already fairly complex as-is). But the compiler should be clever enough to figure out that it can inline the call to `subslice`, so no actual subroutine call should happen.

src/guest_memory.rs
TimePrinciple: > The `offset` parameter is used in the `read` and `write` implementations in `Bytes<GuestAddress> for T: GuestMemory` :/

Oh I didn't read that far, let's just remove the `assertion` here then 😉

src/volatile_memory.rs
```diff
@@ -538,6 +538,32 @@ impl<'a, B: BitmapSlice> VolatileSlice<'a, B> {
...
+    ///
+    /// ```rust
+    /// # use vm_memory::VolatileSlice;
+    /// let mut arr = [1, 2, 3, 4, 5];
+    /// let slice = VolatileSlice::from(arr.as_mut_slice());
+    ///
+    /// assert_eq!(slice.truncate(3).len(), 3);
+    /// assert_eq!(slice.truncate(10).len(), 5);
+    /// ```
+    pub fn truncate(&self, count: usize) -> VolatileSlice<'a, B> {
```
andreeaflorescu, line 541: Truncate somehow implies (to me) that we're updating self because of how `truncate` on Vec works: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.truncate.

Shouldn't we be able to obtain the same result by calling `subslice` with offset = 0 & count as the size we want to "truncate" to?

src/volatile_memory.rs
```diff
@@ -1651,6 +1662,28 @@ mod tests {
...
     const DEFAULT_PAGE_SIZE: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(0x1000) };
 
+    #[test]
+    fn test_compute_end_offset() {
+        let mut array = [1, 2, 3, 4, 5];
+        let slice = VolatileSlice::from(array.as_mut_slice());
+
+        // Iterate over all valid ranges, assert that they pass validation.
+        // This includes edge cases such as len = 0 and base = 5!
+        for len in 0..5 {
```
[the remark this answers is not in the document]
roypat: Yup! fixed :)

src/volatile_memory.rs
roypat: Argh, true. Naming is hard. Maybe `.truncated()` instead (e.g. with an extra 'd')?

> Shouldn't we be able to obtain the same result by calling subslice with offset = 0 & count as the size we want to "truncate" to?

It'd need to be count = min(size we want to be, length). I didn't like it because subslice does a million bounds checks that are all unnecessary for these arguments. But I should probably actually benchmark it/see if the compiler manages to optimize it away

roypat: Damn, criterion indeed says there's no difference between them. Guess rustc is better than I thought at optimizing xP

So my only argument is "it avoids .unwrap()-ing something that we know can never fail"

src/volatile_memory.rs
```diff
@@ -1011,19 +1011,7 @@ impl<B: BitmapSlice> VolatileMemory for VolatileSlice<'_, B> {
...
-            unsafe {
-                VolatileSlice::with_bitmap(
-                    self.addr.add(offset),
-                    count,
-                    self.bitmap.slice_at(offset),
-                    self.mmap,
-                )
-            },
-        )
+        self.subslice(offset, count)
```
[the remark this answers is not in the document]
roypat: Mhh, we could. But tbh, `slice.get_slice(start, count)` reads a bit weirdly (and conversely, so does `region.subslice(start, count)`) :/

CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
...
+
+### Changed
+
+- \[[#307](https://github.com/rust-vmm/vm-memory/pull/304)\] Move `read_volatile_from`, `read_exact_volatile_from`,
+  `write_volatile_to` and `write_all_volatile_to` functions from the `GuestMemory` trait to the `Bytes` trait.
+
+### Removed
+
+- \[[#307](https://github.com/rust-vmm/vm-memory/pull/304)\] Remove deprecated functions `Bytes::read_from`, `Bytes::read_exact_from`,
+  `Bytes::write_to`, `Bytes::write_all_to`, `GuestMemory::as_slice`, `GuestMemory::as_slice_mut`, `GuestMemory::with_regions`,
```
bonzini, line 17: I agree that the non-volatile read/write method are not sound for `Bytes`, but maybe `read_exact_from`/`write_all_to` could be added to `ByteValued`?

bonzini (commented):
I'm too out of the loop to approve anything, but it looks good.

The only comments I have are:
* you could add methods to read/write a ByteValued from an `impl Read`/`impl Write`
* `truncate` needs to be renamed, and I'd personally make it return `Result` but I'm not going to insiste

So the next release would be 1.0? :)

CHANGELOG.md
roypat: Yeah, that makes sense to me! There's some longer-term plans to eliminate `ByteValued` and switch to using zerocopy (https://github.com/rust-vmm/vm-memory/issues/246), which provides the same functionality. Their traits do actually implement the `read_exact_from`/`write_all_to` functionality you're referring to for free (e.g. https://docs.rs/zerocopy/latest/zerocopy/trait.IntoBytes.html#method.write_to_io).

roypat:
> I'm too out of the loop to approve anything, but it looks good.

Thanks for having a look anyway!

> The only comments I have are:
> 
>   - you could add methods to read/write a ByteValued from an `impl Read`/`impl Write`
>
>   - `truncate` needs to be renamed, and I'd personally make it return `Result` but I'm not going to insiste

see inline comments :) 
 
> So the next release would be 1.0? :)

slowly but surely working towards it at least, I hope :D

CHANGELOG.md
bonzini: Yes, it makes sense to make `ByteValued` a thin wrapper over zerocopy, wrapping FromBytes/IntoBytes/AsBytes and providing the implementation of the legacy vm-memory API.

But even if it's going to be a legacy API it will take time to phase it out from everywhere, so let's add the two to ByteValued anyway.

roypat: Mhh, the possibility of turning it into a wrapper for a phase-out period is certainly enticing... maybe some `zerocopy` feature that if enabled provides a blanket impl for ByteValued in terms of zerocopy, and causes rustc to warn on manual impls of the trait (if that's possible), and then later the default impl can be promoted to trait bounds, and even later to a removal of ByteValued altogether once everyone only uses zerocopy types anymore. I'll have to think about that, thanks!

For now, I've added the read/write variants as suggested, since they won't cause any harm indeed :) 
(and also fixed up two minor things I noticed about ByteValued while I was at it)

ShadowCurse: Since `read_exact_from`/`write_all_to` were already deprecated, I would not move them to another trait and keep them deleted. Looking at the impls: `write_all_to` is just indirection one liner, and `read_exact_from` is a constructor with incorrect name. Do we really need to keep them around?

andreeaflorescu (dismissed):

bonzini (changes requested):
Sorry, one more small request.

CHANGELOG.md
bonzini: They were deprecated because they're unsound in Bytes context, but they're useful nevertheless and make it easier to write correct code (for example avoiding that you use read/write instead of read_exact/write_all). The PR just moves them to the right place.

src/bytes.rs
```diff
@@ -116,6 +117,20 @@ pub unsafe trait ByteValued: Copy + Send + Sync {
...
+        writer.write_all(self.as_slice())
+    }
+
+    /// Constructs an instance of this [`ByteValued`] by reading from the given [`Read`] impl.
+    fn read_exact_from<R: Read>(mut reader: R) -> Result<Self, std::io::Error> {
+        // SAFETY: ByteValued objects must be assignable from arbitrary byte
+        // sequences and are mandated to be packed.
+        // Hence, zeroed memory is a fine initialization.
+        let mut result: Self= unsafe { MaybeUninit::<Self>::zeroed().assume_init() };
+        reader.read(result.as_mut_slice()).map(|_| result)
```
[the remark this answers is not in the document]
roypat: Ahh, yes, good catch! Fixed :)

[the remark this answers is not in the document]
roypat: Alright, also added `zeroed()`, and some tests for them as well! Please have another look

bonzini:
Thanks and sorry for the late feedback.

bonzini (approved):

PR #310, opened 2025-01-28
Cleanups for Clippy and Windows

bonzini:
### Summary of the PR

Fix some more issues from clippy and let vm-memory compile on Windows.

Putting these two together because they is really no functional change.

### Requirements

requirements:

  test.

### DIFF
--- a/src/volatile_memory.rs
+++ b/src/volatile_memory.rs
@@ -16,12 +16,15 @@
 //! doing with that hunk of memory.
 //!
 //! For the purposes of maintaining safety, volatile memory has some rules of its own:
+//!
 //! 1. No references or slices to volatile memory (`&` or `&mut`).
+//!
 //! 2. Access should always been done with a volatile read or write.
-//!    The First rule is because having references of any kind to memory considered volatile would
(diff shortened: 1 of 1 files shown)


src/volatile_memory.rs
```diff
@@ -312,15 +315,37 @@ pub struct PtrGuard {
...
 
+#[cfg(all(feature = "xen", unix))]
+impl PtrGuard {
+    const PROT_READ: i32 = libc::PROT_READ;
+    const PROT_WRITE: i32 = libc::PROT_WRITE;
+
+    fn map(
+        mmap: Option<&MmapInfo>,
+        addr: *mut u8,
+        prot: i32,
```
roypat, line 326: Maybe it's simpler to change the `prot` argument to just be `read: bool`, then inside the body of this function we do `let prot = if read { libc::PROT_READ } else { libc::PROT_WRITE }`? (or even just have that in the old `cfg`'d block).

bonzini: Sure.

stefano-garzarella (approved):

PR #312, opened 2025-02-10
Generify parts of mmap.rs

roypat:
### Summary of the PR

Mainly, add the concept of a `GuestRegionCollection`, which just manages a list of some `GuestMemoryRegion` impls. Functionality wise, it offers the same implementations as `GuestMemoryMmap`. As a result, turn `GuestMemoryMmap` into a type alias for `GuestRegionCollection` with a fixed `R = GuestRegionMmap`.

While we're at it, generalize the `impl Bytes for GuestRegionMmap` into `impl<R: GuestMemoryRegion> Bytes for R`, because really that impls didn't actually make use of any specifics of the `GuestRegionMmap` struct.

This PR is the backward-compatible subset of #317. In that context, it is a first step towards resolving the incompatibility between the unix and xen features (currently, the code in `mmap.rs` only works if exactly one of the modules `mmap_windows`, `mmap_xen` and `mmap_unix` exist, since they all define roughly the same types which are then imported by the `mmap` module without renames). However, short-term is is also a required step for enabling things like https://github.com/rust-vmm/vm-memory/issues/315 without necessarily requiring changes in core vm-memory code.

### Requirements

requirements:

  test.

### DIFF
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -408,7 +408,9 @@ pub trait GuestMemory {
     fn num_regions(&self) -> usize;
 
     /// Returns the region containing the specified address or `None`.
-    fn find_region(&self, addr: GuestAddress) -> Option<&Self::R>;
+    fn find_region(&self, addr: GuestAddress) -> Option<&Self::R> {
+        self.iter().find(|region| addr >= region.start_addr() && addr <= region.last_addr())
+    }
 
     /// Gets an iterator over the entries in the collection.
     ///
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -405,7 +405,9 @@ pub trait GuestMemory {
     type R: GuestMemoryRegion;
 
     /// Returns the number of regions in the collection.
-    fn num_regions(&self) -> usize;
+    fn num_regions(&self) -> usize {
+        self.iter().count()
+    }
 
     /// Returns the region containing the specified address or `None`.
     fn find_region(&self, addr: GuestAddress) -> Option<&Self::R> {
--- a/src/guest_memory.rs
+++ b/src/guest_memory.rs
@@ -50,10 +50,11 @@ use std::sync::atomic::Ordering;
 use std::sync::Arc;
 
 use crate::address::{Address, AddressValue};
-use crate::bitmap::{Bitmap, BS, MS};
+use crate::bitmap::MS;
 use crate::bytes::{AtomicAccess, Bytes};
 use crate::io::{ReadVolatile, WriteVolatile};
 use crate::volatile_memory::{self, VolatileSlice};
+use crate::GuestMemoryRegion;
 
 /// Errors associated with handling guest memory accesses.
 #[allow(missing_docs)]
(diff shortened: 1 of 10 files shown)


src/guest_memory.rs
```diff
@@ -586,6 +587,168 @@ pub trait GuestMemory {
...
+    ///
+    /// * `regions` - The vector of `Arc` regions.
+    ///               The regions shouldn't overlap and they should be sorted
+    ///               by the starting address.
+    pub fn from_arc_regions(
+        regions: Vec<Arc<R>>,
+    ) -> std::result::Result<Self, GuestRegionCollectionError> {
+        if regions.is_empty() {
+            return Err(GuestRegionCollectionError::NoMemoryRegion);
+        }
```
ShadowCurse, line 674: Is it really an error to provide no regions?

src/guest_memory.rs
```diff
@@ -586,6 +587,168 @@ pub trait GuestMemory {
...
+    pub fn insert_region(
+        &self,
+        region: Arc<R>,
+    ) -> std::result::Result<GuestRegionCollection<R>, GuestRegionCollectionError> {
+        let mut regions = self.regions.clone();
+        regions.push(region);
+        regions.sort_by_key(|x| x.start_addr());
+
+        Self::from_arc_regions(regions)
+    }
```
ShadowCurse, line 705: This should just take `&mut self` and mutate in place. The check for region intersection can be moved into a separate function and used here and in the `from_arc_regions`

src/guest_memory.rs
```diff
@@ -586,6 +587,168 @@ pub trait GuestMemory {
...
+
+    /// Remove a region from the [`GuestRegionCollection`] object and return a new `GuestRegionCollection`
+    /// on success, together with the removed region.
+    ///
+    /// # Arguments
+    /// * `base`: base address of the region to be removed
+    /// * `size`: size of the region to be removed
+    pub fn remove_region(
+        &self,
```
ShadowCurse, line 714: Same here, why not `&mut self`

src/guest_memory.rs
```diff
@@ -586,6 +587,168 @@ pub trait GuestMemory {
...
+    ) -> std::result::Result<(GuestRegionCollection<R>, Arc<R>), GuestRegionCollectionError> {
+        if let Ok(region_index) = self.regions.binary_search_by_key(&base, |x| x.start_addr()) {
+            if self.regions.get(region_index).unwrap().len() == size {
+                let mut regions = self.regions.clone();
+                let region = regions.remove(region_index);
+                return Ok((Self { regions }, region));
+            }
+        }
+
+        Err(GuestRegionCollectionError::NoMemoryRegion)
```
ShadowCurse, line 726: Is this an error to remove a non existing region? Maybe return type should be `(Self, Option<Arc<R>>)` where optional is the removed region if it existed?

src/guest_memory.rs
```diff
@@ -586,6 +587,168 @@ pub trait GuestMemory {
...
+    }
+
+    fn find_region(&self, addr: GuestAddress) -> Option<&R> {
+        let index = match self.regions.binary_search_by_key(&addr, |x| x.start_addr()) {
+            Ok(x) => Some(x),
+            // Within the closest region with starting address < addr
+            Err(x) if (x > 0 && addr <= self.regions[x - 1].last_addr()) => Some(x - 1),
+            _ => None,
+        };
+        index.map(|x| self.regions[x].as_ref())
```
ShadowCurse, line 744: How many regions there can possibly be? I think a simple for loop where we check if address is in the region will be as fast in general case and more readable.

src/guest_memory.rs
```diff
@@ -586,6 +587,168 @@ pub trait GuestMemory {
     }
 }
 
+/// Errors that can occur when creating a memory map.
+#[derive(Debug, thiserror::Error)]
+pub enum GuestRegionCollectionError {
```
ShadowCurse, line 592: This whole `GuestRegionCollection` looks like it can be moved into it's own module. Also maybe rename to just `Regions`? I assume there are no other regions other than memory regions.

src/guest_memory.rs
roypat: This code is just moved from mmap.rs without changes. There might be improvements that can be made to the actual API, but that's out of scope here I'd say, because ideally we'd keep backwards compatibility here.

src/guest_memory.rs
roypat: This code is just moved from mmap.rs without changes. There might be improvements that can be made to the actual API, but that's out of scope here I'd say, because ideally we'd keep backwards compatibility here.

src/guest_memory.rs
roypat: This code is just moved from mmap.rs without changes. There might be improvements that can be made to the actual API, but that's out of scope here I'd say, because ideally we'd keep backwards compatibility here.

src/guest_memory.rs
roypat: This code is just moved from mmap.rs without changes. There might be improvements that can be made to the actual API, but that's out of scope here I'd say, because ideally we'd keep backwards compatibility here.

src/guest_memory.rs
roypat: This one is also just moved from mmap.rs, so I also err'd on the side of keeping things as they are. The default implementation for this trait function indeed just uses a simple loop.

src/guest_memory.rs
roypat: Mh, I'd be open to having it in its own module. What should we call it though?

ShadowCurse: Just `regions.rs`?

src/guest_memory.rs
ShadowCurse: ok

src/guest_memory.rs
ShadowCurse: ok

src/guest_memory.rs
ShadowCurse: ok

src/guest_memory.rs
ShadowCurse: ok

src/guest_memory.rs
ShadowCurse: ok

src/guest_memory.rs
roypat: done

ShadowCurse (approved):

roypat:
Hey @bonzini, since you chimed in over at #321 the other day, any thoughts on this one? :o

bonzini:
> While we're at it, generalize the impl Bytes for GuestRegionMmap into impl<R: GuestMemoryRegion> Bytes for R, because really that impls didn't actually make use of any specifics of the GuestRegionMmap struct.

I would need to double check that. IIRC if QEMU was ever going to use vm-memory, which actually could happen now, it needed something custom to handle access to an assigned device's BAR.

roypat:
> > > While we're at it, generalize the impl Bytes for GuestRegionMmap into impl<R: GuestMemoryRegion> Bytes for R, because really that impls didn't actually make use of any specifics of the GuestRegionMmap struct.
> > 
> > 
> > I would need to double check that. IIRC if QEMU was ever going to use vm-memory, which actually could happen now, it needed something custom to handle access to an assigned device's BAR.
> 
> Ah, that's good to know, thanks! I can rework it, so that an additional empty marker trait is required to be implemented, e.g. `GuestMemoryRegionBytes: GuestMemoryRegion` and then only have `impl<R: GuestMemoryRegionBytes> Bytes for R { ... }`?

Done :)

roypat:
> > The error type handling is a bit wack, but this is needed to preserve
> > backwards compatibility
> 
> Do we really want to keep backwards compatibility at all costs? I am afraid that we'll end up with even more technical debt. How hard would it be to update to the new vm-memory version if we were NOT to keep the backwards compatibility?

Honestly, the backward compatibility went out the window a few revisions ago... I'll just make this change and clean it up properly. We already have quite a few breaking changes accumulated on `main` anyway, and this should be a fairly minor one.

src/guest_memory.rs
```diff
@@ -408,7 +408,9 @@ pub trait GuestMemory {
     fn num_regions(&self) -> usize;
 
     /// Returns the region containing the specified address or `None`.
-    fn find_region(&self, addr: GuestAddress) -> Option<&Self::R>;
+    fn find_region(&self, addr: GuestAddress) -> Option<&Self::R> {
```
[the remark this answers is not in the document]
roypat: Mainly because it's possible, and it means that downstream impls of the trait need to do less (in my case, I was implementing a mock for some testing and got a bit annoyed at having to unneccessarily write out things that I didn't very much care about in that scenario). Looking at it now, `num_regions` can also get a default impl, so I'll add that too haha

andreeaflorescu (approved):

ShadowCurse (approved):

PR #316, opened 2025-03-14
refactor: Move `NewBitmap` trait to `bitmap` module

roypat:
There is really no reason (at least that I can see) for this to be defined in `mmap.rs`, and for it to be gated behind the backend-mmap feature.

While we're at it, add a comment explaining why this cannot be part of the normal `Bitmap` trait.

Add a public reexport from mmap.rs anyway for backwards compat reasons. No functional change.

### Summary of the PR

### Requirements

requirements:

  test.

### DIFF
--- a/src/bitmap/backend/atomic_bitmap.rs
+++ b/src/bitmap/backend/atomic_bitmap.rs
@@ -6,10 +6,7 @@
 use std::num::NonZeroUsize;
 use std::sync::atomic::{AtomicU64, Ordering};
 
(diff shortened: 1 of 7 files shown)


ShadowCurse (approved):

stefano-garzarella (approved):
> There is really no reason (at least that I can see) for this to be defined in mmap.rs, and for it to be gated behind the backend-mmap feature.

The only reason I can see can be because it was used only with backend-mmap feature, but yeah LGTM

ShadowCurse (approved):

stefano-garzarella (approved):

PR #317, opened 2025-03-14
[RFC] Make mmap_unix and mmap_xen compatible 

roypat:
This PR enables the use of structs from the modules `mmap_unix` and `mmap_xen` at the same time, resolving their mutual exclusivity. Particularly, one can now use both xen and non-xen backends at the same time (e.g. no longer is there a need to make a decision at compile time).

If the xen feature is enabled, all unit tests run for both xen and non-xen backends.

The windows bits are only compile tested in CI.

There's probably a lot that can be improved still, e.g. there's more duplication between unix and windows support now than probably needed, but as a proposal it's good enough I think. I also probably messed up re-exports every now and then, but the changes to the vhost crate to support this do not look too bad: https://github.com/rust-vmm/vhost/pull/325

### Requirements

requirements:

  test.


roypat:
@epilys @stsquad please have a look, this is essentially what I was trying to describe over in #314 the other day

PR #319, opened 2025-03-21
Use stat instead of seek to determine file size

roypat:
When constructing a MmapRegion that is backed by a `File`, we sanity check that the size of the file passed in is sufficiently large given the offset / length requested of mmap. Do this check using stat instead of seek, as stat is supported on more file descriptors (particularly, it is supported on guest_memfd, while seek isnt).

### Summary of the PR

### Requirements

requirements:

  test.

### DIFF
--- a/src/mmap.rs
+++ b/src/mmap.rs
@@ -14,7 +14,6 @@
 
 use std::borrow::Borrow;
 #[cfg(unix)]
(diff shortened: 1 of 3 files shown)


roypat:
ugh nevermind this doesnt work for sparse files :(

PR #327, opened 2025-05-30
[DRAFT] I/O virtual memory (IOMMU) support

XanClic:
### Summary of the PR

This MR adds support for an IOMMU, and thus I/O virtual memory handling.

#### New Memory Trait: `IoMemory`

Handling I/O virtual memory requires a new interface to access guest memory:
`GuestMemory` does not allow specifying the required access permissions, which
is necessary when working with MMU-guarded memory.

We could add memory access methods with such a permissions parameter to
`GuestMemory`, but I prefer to provide a completely new trait instead.  This
ensures that users will only use the interface that actually works when working
with (potentially) I/O virtual memory, i.e.:
- They must always specify the required permissions,
- They cannot (easily) directly access the memory regions, because doing so
  generally assumes that regions are long, continuous, and any address in a
  given range will be in the same memory region.  This is absolutely no longer
  the case with virtual memory, which is heavily fragmented into pages.

That is, adding a new trait (`IoMemory`) allows to catch a lot of potential
mistakes at compile time, which I feel is much better than finding out at
runtime that some place forgot to specify the access permissions.

Unfortunately, this is an incompatible change because we need to decide on a
single guest memory trait that we expect users to primarily use: We can only
have one blanket implementation of e.g. `Bytes`, and this MR changes that
blanket implementation to be on `IoMemory` instead of `GuestMemory` because we
want to prefer `IoMemory` with its permissions-including interface.

While this MR does provide a blanket implementation of `IoMemory` for all
`GuestMemory`, Rust isn’t fully transitive here, so just because we have a
blanket `impl IoMemory for GuestMemory` and a blanket `impl Bytes for IoMemory`
doesn’t really implicitly give us an `impl Bytes for GuestMemory`.

What this means can be seen in virtio-queue (in vm-virtio): It uses trait bounds
like `M: GuestMemory` only, but then expects to be able to use the `Bytes`
trait.  This is no longer possible, the trait bound must be extended to
`M: GuestMemory + Bytes` or replaced by `M: IoMemory` (the latter is what we
want).

#### Guest Address Type

Another consideration is that I originally planned to introduce new address
types.  `GuestAddress` currently generally refers to a guest physical address
(GPA); but we now also need to deal with I/O virtual addresses (IOVAs), and an
IOMMU generally doesn’t translate those into GPAs, but VMM user space addresses
(VUAs) instead, so now there’s three kinds of addresses.  Ideally, all of those
should get their own type; but I felt like:
- This would require too many changes from our users, and
- You don’t even know whether the address you use on an `IoMemory` object is an
  IOVA or a GPA.  It depends on whether the IOMMU is enabled or not, which is
  generally a runtime question.

Therefore, I kept `GuestAddress` as the only type, and it may refer to any of
the three kinds of addresses (GPAs, IOVAs, VUAs).

#### Async Accesses

I was considering whether to also make memory accesses optionally `async`.  The
vhost-user IOMMU implementation basically needs two vhost-user socket roundtrips
per IOTLB miss, which can make guest memory accesses quite slow.  An `async`
implementation could allow mitigating that.

However, I decided against it (for now), because this would also require
extensive changes in all of our consuming crates to really be useful: Anything
that does a guest memory access should then be `async`.

I think if we want to add this functionality later, it should be possible in a
compatible manner.

#### vm-virtio

Implementation: https://gitlab.com/hreitz/vm-virtio/-/commits/iommu

As stated above, places that bind `M: GuestMemory` but expect the `Bytes` trait
to also be implemented need to be changed to `M: GuestMemory + Bytes` or
`M: IoMemory`.  I opted for the latter approach, and basically replaced all
`GuestMemory` instances by `IoMemory`.

(That is what we want because dropping `GuestMemory` in favor of `IoMemory`
ensures that all vm-virtio crates can work with virtual memory.)

#### vhost

Implementation: https://gitlab.com/hreitz/vhost/-/commits/iommu

Here, the changes that updating vm-memory necessitates are quite marginal, and
have a similar cause: But instead of requiring the `Bytes` trait, it’s the
`GuestAddressSpace` trait.  The resolution is the same: Switch from requiring
`GuestMemory` to `IoMemory`.

The rest of the commits concerns itself with implementing `VhostUserIommu` and
allowing users to choose to use `IommuMemory<GuestMemoryMmap, VhostUserIommu>`
instead of only `GuestMemoryMmap`.

#### virtiofsd (as one user)

Implementation: https://gitlab.com/hreitz/virtiofsd-rs/-/commits/iommu

This is an example of an actual user.  Updating all crates to IOMMU-supporting
versions actually does not require any changes to the code, but enabling the
'iommu' feature does: This feature makes the vhost-user-backend crate require
the `VhostUserBackend::Memory` associated type (because associated type defaults
are not stable yet), so this single line of code must be added (which sets the
type to `GuestMemoryMmap<BitmapMmapRegion>`).

Actually enabling IOMMU support is then a bit more involved, as it requires
switching away from `GuestMemoryMmap` to `IommuMemory` again.

However, to me, this shows that end users working with concrete types do not
seem to be affected by the incompatible `IoMemory` change until they want to opt
in to it.  That’s because `GuestMemoryMmap` implements both `GuestMemory` and
`IoMemory` (thanks to the blanket impl), so can transparently be used wherever
the updated crates expect to see an `IoMemory` type.

### Why a Draft?

I did not write unit tests yet.  I assume the design will be something that will
need some discussion, so I wanted to publish my state before I fully finalize
it.

### Requirements

requirements:

  test. *(not done, which is why this is a draft)*


XanClic:
Pushed an update without actual changes (just re-committing the top commit) to trigger a CI re-run. This time, only the coverage check failed (as expected).

XanClic:
cc @germag

src/guest_memory.rs
```diff
@@ -591,17 +592,62 @@ impl<T: GuestMemory + ?Sized> Bytes<GuestAddress> for T {
...
+
+        let completed = self.try_access(
+            expected,
+            addr,
+            |offset, len, region_addr, region| -> Result<usize> {
+                assert_eq!(offset, 0);
+                if len < expected {
+                    return Err(Error::PartialBuffer {
+                        expected,
+                        completed: len,
```
bonzini, line 622: `completed` should be 0, since you didn't read anything. So here you could also return `Ok(0)` ("no more data") and let the `if` below return `Error::PartialBuffer`.

src/guest_memory.rs
```diff
@@ -591,17 +592,62 @@ impl<T: GuestMemory + ?Sized> Bytes<GuestAddress> for T {
...
```
bonzini, line 652: Same as above.

src/io_memory.rs
```diff
@@ -0,0 +1,183 @@
...
+//!
+//! In addition, any access to virtual memory must be annotated with the intended access mode (i.e.
+//! reading and/or writing).
+
+use crate::guest_memory::Result;
+use crate::{bitmap, GuestAddress, GuestMemory, MemoryRegionAddress, VolatileSlice};
+
+/// Permissions for accessing virtual memory.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Permissions {
```
bonzini, line 22: What about defining this like:

```rust
pub struct Permissions(u8);
impl Permissions {
    pub const No: Permissions = Permissions(0);
    pub const Read: Permissions = Permissions(1);
    pub const Write: Permissions = Permissions(2);
    pub const ReadWrite: Permissions = Permissions(3);
}
```

and implementing `&` and `|` as just a bitwise operation on `x.0`? (Basically reinventing the bitflags crate to avoid introducing a new dependency)

src/io_memory.rs
```diff
@@ -0,0 +1,183 @@
...
+    ) -> Result<VolatileSlice<bitmap::MS<Self::PhysicalMemory>>>;
+
+    /// If this virtual memory is just a plain `GuestMemory` object underneath without an IOMMU
+    /// translation layer in between, return that `GuestMemory` object.
+    fn physical_memory(&self) -> Option<&Self::PhysicalMemory> {
+        None
+    }
+}
+
+impl<M: GuestMemory> IoMemory for M {
```
bonzini, line 147: I'd consider just changing `GuestMemory`. Is anyone using `try_access` directly? It's not private, but almost so.

src/iommu.rs
```diff
@@ -172,6 +176,22 @@ pub struct IotlbFails {
     pub access_fails: Vec<IovaRange>,
 }
 
+/// [`IoMemory`] type that consists of an underlying [`GuestMemory`] object plus an [`Iommu`].
+///
+/// The underlying [`GuestMemory`] is basically the physical memory, and the [`Iommu`] translates
+/// the I/O virtual address space that `IommuMemory` provides into that underlying physical address
+/// space.
+#[derive(Debug, Default)]
+pub struct IommuMemory<M: GuestMemory, I: Iommu> {
```
bonzini, line 186: I am not sure I understand methods such as `inner_replaced`. Maybe what you want is a `GuestAddressSpace` here, and `physical_memory()` returns a `GuestAddressSpace::T`?

Maybe with that change you don't need the `Arc<I>` anymore, but you can just use `&I`?

src/io_memory.rs
```diff
@@ -87,6 +93,8 @@ impl std::ops::BitAnd for Permissions {
 pub trait IoMemory {
     /// Underlying `GuestMemory` type.
     type PhysicalMemory: GuestMemory;
+    /// Dirty bitmap type for tracking writes to the IOVA address space.
+    type Bitmap: Bitmap;
```
bonzini, line 97: I am not sure I like the design here. The IOVA address space, being virtual, can have aliases. Could you still store the bitmap in the low-level memory region, but with accesses going through IOMMU translation?

What does vhost-user require? Are dirty bitmap accesses done in IOVA or GPA space?

bonzini (commented):
Overall this is good stuff. It also has a lot of overlap with the experimental changes that were necessary in order to use vm-memory in QEMU, which is a very good thing.

I have two main questions:

* do we really need an `IoMemory` or can we change `GuestMemory`?
* if we need an `IoMemory`, I think the `PhysicalMemory: GuestMemory` type instead be an `AS: GuestAddressSpace`? (and likewise for `physical_memory()`?

I'm marking this as "request changes" because there would be changes either way.

src/io_memory.rs
bonzini: (and also reinventing it to hide the underlying implementation of the fake enum)

src/io_memory.rs
XanClic: I did consider that, and I put the reasoning why I decided not to do so (and indeed failed to do so) into the opening comment: The most immediate reason is that anyone accessing memory needs to specify the required access permissions. Yes, most accesses go through the `Bytes` interface, i.e. already specify the intended access, but not all of them:

Which gets me to the second question: There are users (like descriptor_utils used by virtiofsd or vm-virtio/virtio-queue) that currently use `get_slice()`. `get_slice()` is dangerous because it cannot cope with fragmented memory, which becomes especially apparent when using paged virtual memory. So in fact my current draft replaces those `get_slice()` usages by `try_access()`.

We could of course replace `get_slice()` by something like `get_slices() -> impl Iterator<Item = VolatileSlice>`, but really, that wouldn’t offer more than the `try_access()` we already have, just replace the interior iterator by an exterior one. (Though I’m open to the idea.)

Besides the access mode, there’s also the general idea I outlined in the original comment that `GuestMemory` is quite open about how it is structured into a limited set of `GuestMemoryRegion` objects internally. For virtual memory, that’s no longer applicable: Pages and guest memory regions are not only conceptually different, access to single pages is also quite useless. (Whereas `GuestMemoryRegion` is fundamental to how the vhost code constructs the `GuestMemory` from the information it gets from the front-end, so it must be visible.)

If we continued to use `GuestMemory`, I’m afraid it will be impossible to verify at compile time that all the accesses are sound. For example, a user could try to access virtual memory’s `GuestMemoryRegion` objects, and that would only return an error at runtime.

src/iommu.rs
XanClic: As discussed, yep, I’ll try making it a `GuestAddressSpace` instead to have the user handle the physical memory object management.

(And I’ll see whether I can make it `I` instead of `Arc<I>`.)

src/io_memory.rs
bonzini: Yes, that's pretty convincing. The one thing I'd change however is to always expose `IoMemory` and only put `IommuMemory`/`Iommu`/`Iotlb` under the new feature.

This way, the `Bytes<GuestAddress>` implementation can simply move to `IoMemory`.

src/iommu.rs
bonzini: Yep, something like `IommuMemory<M: GuestAddressSpace, I: Deref<Target = Iommu>>` would do. Then the `Clone` implementation can be conditional on having `Clone` on both `M` and `I`.

The main disadvantage would be an extra `clone()` in `try_access()`. But it could be optimized with a method like

```
    fn snapshot(&self) -> IommuMemory<&M::M, &I> {
         IommuMemory {
              inner: &*self.inner.memory(),
              iommu: &*self.iommu,
              enabled_iommu: self.enabled_iommu
        }
    }

File: src/io_memory.rs (hunk 3 above)
Comment (XanClic) [reply src/io_memory.rs:22]: I’m not dead-set; for bitfields, such a struct with constants can make more sense. On the other hand, I like using an `enum` if anything just to check that `match` is exhaustive (e.g. when we need to translate `Permissions` into `VhostAccess`).

FWIW, starting from `opt-level = 2`, this does get optimized properly:
```
   0:   48 89 f8                mov    rax,rdi
   3:   48 21 f0                and    rax,rsi
   6:   c3                      ret
```

As a compromise, I could do `repr(u8)` with `bitand(x, y) { ((x as u8) | (y as u8)).try_into().unwrap()` and an accompanying `TryFrom<u8>` implementation. Even on `opt-level = 0`, that would yield:
```
   0:   48 89 f8                mov    rax,rdi
   3:   48 89 44 24 f0          mov    QWORD PTR [rsp-0x10],rax
   8:   48 89 74 24 f8          mov    QWORD PTR [rsp-0x8],rsi
   d:   48 21 f0                and    rax,rsi
  10:   c3                      ret
```
(`opt-level = 1` then also removes the stack accesses, as above)

File: src/guest_memory.rs (hunk 2 above)
Comment (XanClic) [reply src/guest_memory.rs:652]: Indeed.

File: src/io_memory.rs (hunk 3 above)
Comment (bonzini) [reply src/io_memory.rs:22]: > On the other hand, I like using an enum if anything just to check that match is exhaustive (e.g. when we need to translate Permissions into VhostAccess).

Ah, you're right. Exhaustiveness is important.

Let's go for the repr, but with a private panicking `fn from_repr(x: u8 -> Self)` instead of the TryFrom trait.

File: src/io_memory.rs (hunk 6 above)
Comment (bonzini) [reply src/io_memory.rs:97]: > You mean putting the responsibility on the try_access() callback to dirty a bitmap slice given to it if it has written anything?

No, I confused `try_access` and `get_slice` sorry. For IoMemory, dirtying is already done it in `try_access` itself; for Iommu the slice could be returned by `translate`, that is it would be part of the Iotlb entry?

File: src/io_memory.rs (hunk 4 above)
Comment (bonzini) [reply src/io_memory.rs:147]: I think you'd get it just by importing Bytes...

The problem with moving IoMemory to the new feature is that features are additive. You can't make the iommu feature pick whether Bytes<> is added to GuestMemory or IoMemory.

Unrelated to this, maybe we should bite the bait and add a prelude...

Comment (bonzini) [reply src/io_memory.rs:147]: Testcase:

lib.rs:

```rust
pub trait GuestMemory {}
pub trait IoMemory {}

impl<T: GuestMemory> IoMemory for T {}

pub trait HelloWorld {
    fn hello_world(&self);
}

impl<T: IoMemory> HelloWorld for T {
     fn hello_world(&self) {}
}
```

test.rs:
```rust
use lib::{GuestMemory, HelloWorld};

pub struct Foo;
impl GuestMemory for Foo {}

fn main() {
    Foo.hello_world();
}
```

```
$ rustc --crate-type=rlib lib.rs --edition=2021
$ rustc test.rs --extern lib=liblib.rlib --edition=2021
```

roypat:
> > I don’t like keeping GuestMemory because I think actually none of its current methods will work with virtual memory, and so I would prefer adding a new interface that makes it clear that these aren’t available at compile time.
> 
> Yes, GuestMemory is a nice interface for implementation but clients should switch to `IoMemory`.

Different question about this: Do we really need the `IoMemory` trait, or could consumers just switch to `IommuMemory<M: GuestMemory, I: Iommu>` (e.g. be parametric by `M` and `I` instead of `T: IoMemory`, and for "no iommu" systems we can have a no-op implementation of `Iommu`)? That would also side-step all the problems of conflicting default implementations (although I'm not sure if the compiler will be clever enough to optimize the no-op iommu setup as well as with the special impl of IoMemory for GuestMemory :/)

XanClic:
> could consumers just switch to IommuMemory<M: GuestMemory, I: Iommu> (e.g. be parametric by M and I instead of T: IoMemory, and for "no iommu" systems we can have a no-op implementation of Iommu)?

Doesn’t sound bad, but would be a more “radical change”: Intermediate users like virtio-queue currently use memory types via `M: GuestMemory`. Converting them to `M: IoMemory` is straightforward and allows their users to continue to use non-virtual memory types like `GuestMemoryMmap` with no change needed if they don’t need IOMMU functionality.

If we change patterns like `foo<M: Deref<Target: GuestMemory>>(mem: M)` to `foo<M: Deref<Target = IommuMemory>>(mem: M)`, all users will need to be changed to use `IommuMemory`.

src/io_memory.rs
bonzini: > <IommuMemory as IoMemory>::get_slice() has the VolatileSlice do the work. For this, it replaces the VolatileSlice’s internal bitmap slice (which is initially set by the GuestMemoryRegion) by the right slice in the IOVA space.

To clarify, I was saying it should not be the `IommuMemory` that decides the address space used by the bitmap - whether IOVA space as vhost wants or GPA space as everyone else (?) wants. I thought `Iommu` would return the slice in the `Iotlb` object, but I agree that probably `Iommu` isn't the right place either.

That said I don't care much, because as long as `IommuMemory` is hidden behind a feature, it's just a convenience or a sample implementation (and QEMU probably will not use `IommuMemory`, only `IoMemory`). Maybe just add a comment that says "`NOTE: IommuMemory` can only be used if the dirty bitmap tracks accesses in IOVA space".

roypat:
> > And having less traits in vm-memory would be a win for maintainability imo.
> 
> In theory yes, in practice it's appearing that there are some design decisions that are made by `IommuMemory` that are good for vhost but not for VMMs.
> 
> In general, vm-memory uses traits because VMMs may have their own memory backends. Considering that QEMU is able to use vm-memory designed 5 years ago almost-unmodified, even without IOMMU support (and hopefully, unmodified once there is `IoMemory`), this worked out well and I'd rather keep it.

My concern mainly stems from feedback like [this](https://github.com/rust-vmm/vm-virtio/pull/278#issuecomment-1886565130), where the existing traits were already a big struggle for people trying to use vm-memory. That said, I'm not fundamentally opposed to adding a `IoMemory` trait, I'm mostly not a fan of the blanket impls. Let me bother you with some more clarifying questions:
- Currently, we have `impl<M: GuestMemory> IoMemory for M`, so from that I'll assume that implementing both IoMemory and GuestMemory for the same type is not a use case (because with the blanket impl, it's not possible). Does that sound right?
- If so, then the blanket impl is for compat reasons (e.g. so we can pass a `M: GuestMemory` where a `M: IoMemory` is requested)?

In this case, IMO it'd be less complex to have the compat layer be done through a new type instead of a blanket impl (e.g. add `struct Compat<M: GuestMemory>(M)` with `impl<M: GuestMemory> IoMemory for Compat<M>` (crates upgrading `M: GuestMemory` bounds to `M: IoMemory` bounds could even provide this shim themselves by keeping the old functions that take `M: GuestMemory` around, and internally dispatching to the new versions taking `IoMemory`). It'd also allow us to solve the `Bytes` dilemma of "implementing for IoMemory or GuestMemory": We could instead just make `Bytes` a super trait of both, and then instead of blanket impls, we can just provide macros that for specific types generate a `Bytes` impl in terms of `GuestMemory`/`IoMemory` (yeah, conceptually weird to implement a trait in terms of a subtrait, but [works](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=dc9ae5bdc148a97a4a8dc3d6391c1700)). This would mean people implementing their own backend will have to add one additional line of code, and code that is generic over GuestMemory/IoMemory can continue using the Bytes impl as today due to the supertrait bound (whereas just moving it over to IoMemory would actually force some code changes in Firecracker, because we're using GuestMemoryMmap, which does not/cannot implement IoMemory, so would lose its Bytes impl).

Paolo, do you have any pointers towards code for how QEMU would like to use this, so it's a bit easier for me to wrap my head around everything?  `git grep`-ing for "vm_memory"  in QEMU doesn't show anything (but probably am missing something) :(

src/iommu.rs
XanClic: After trying to implement it, I’m not sure how to make it work with the slice lifetimes. Both `try_access()` and `get_slice()` return[1] objects with the same lifetime as `self`, which only works if `self` owns the memory. With `GuestAddressSpace`, the lifetime would be limited to the `memory()` snapshot we can take.

The simplest way to do this would be for the caller to take that snapshot and pass it into `get_slice()`/`try_access()`. At that point, I suppose we could completely remove any reference to the `GuestMemory` from `IommuMemory` altogether. I wouldn’t like that very much, though.

Besides my personal preferences, I also don’t know how to make it work at least with users like [`descriptor_utils::Reader`](https://github.com/rust-vmm/vm-virtio/blob/433129f632fdb452f8b08b92de7288e7e3364f1d/virtio-queue/src/descriptor_utils.rs#L160) (and `Writer`): That `new()` function collects slices in memory[2], and returns those in an object whose lifetime is bound to that of its memory parameter. If we used `GuestAddressSpace` for the physical memory, someone in `descriptor_utils` would need to create a snapshot, store it somewhere, and the slices’ lifetimes would be bound to that snapshot’s lifetime. I don’t know how to express that, other than to have the `Reader::new()` caller in turn create that snapshot and pass it into that function. That may indeed be doable, but it doesn’t really feel worth it to me…

So at this point, I’d prefer for `IommuMemory` to own the inner physical memory, and consider `IoMemory` a snapshot like `GuestMemory` is a snapshot.

---

[1] In case of `try_access()`, “return” means passing it to the closure.

[2] Note that currently it uses `GuestMemory::find_region()` + `GuestMemoryRegion::get_slice()` – that will need to be changed into an `IoMemory::try_access(|_, len, region_addr, region| region.get_slice(region_addr, len))`, basically.

src/io_memory.rs
```diff
@@ -0,0 +1,252 @@
...
+impl<M: GuestMemory> IoMemory for M {
+    type PhysicalMemory = M;
+    type Bitmap = <M::R as GuestMemoryRegion>::B;
+
+    fn range_accessible(&self, addr: GuestAddress, count: usize, _access: Permissions) -> bool {
+        if count <= 1 {
+            <M as GuestMemory>::address_in_range(self, addr)
+        } else if let Some(end) = addr.0.checked_add(count as u64 - 1) {
+            <M as GuestMemory>::address_in_range(self, addr)
+                && <M as GuestMemory>::address_in_range(self, GuestAddress(end))
```
roypat, line 156: Mh, what if there's a hole in the physical address range (MMIO gap for example)? I think this might need to be some `try_access` magic (just `.get_slice()` won't cut it in case of the consecutive memory being split into different memslots I think?)

XanClic: Sure! I think most callers didn’t care so far (which is why `GuestMemory` only has `address_in_range()`, callers just used to test the end address), but it would definitely be better to do it right.

bonzini:
> crates upgrading M: GuestMemory bounds to M: IoMemory bounds could even provide this shim themselves by keeping the old functions that take M: GuestMemory around, and internally dispatching to the new versions taking IoMemory

The "problem" with this is the lack of function overloading in Rust. It would mean changing the whole API depending on whether you want the IOMMU indirection or not.

> Paolo, do you have any pointers towards code for how QEMU would like to use this, so it's a bit easier for me to wrap my head around everything? git grep-ing for "vm_memory" in QEMU doesn't show anything (but probably am missing something) :(

For now we have only an RFC implementation [here](https://lore.kernel.org/qemu-devel/aJynpEWNfbrTeirQ@intel.com/T/). But you can see [that it needs to apply some changes to vm-memory](https://lore.kernel.org/qemu-devel/20250807123027.2910950-11-zhao1.liu@intel.com/) in order to add a direction (read/write) argument to `try_access`.

Because QEMU has its backend in the C code, the natural thing for QEMU is to forgo `GuestMemory` completely and just go directly from `IoMemory` to `GuestMemoryRegion`s.

roypat:
> Yeah, that would be nice... and maybe we can do it?!? Almost all users use GuestMemoryMmap instead of GuestMemory, and because most access goes through Bytes anyway, the only thing to change would be an occasional use statement to add GuestMemoryBackend (to access methods like to_region_addr etc.).

mh, you're right, grepping Firecracker and cloud-hypervisor for GuestMemory doesn't show _much_ more than imports and a handful of functions generic over GuestMemory. And the latter shouldn't even need to be changed, at worst they'll need a new import for `GuestMemoryBackend` (or whatever we want to call it).

> For now I'd say the two main contendents are IoMemory and GuestMemoryView. But a big inversion is actually quite appealing as the final change towards 1.0 - in that case I'd keep IoMemory and focus later on the feasibility of using GuestMemory for @XanClic's new trait.

Potentially, we could have less churn for downstream if we do the rename straight away. But only if downstream would upgrade to the new interfaces straight away, instead of dealing with the rename by first updating trait bounds to be `GuestMemoryBackend`, and only later upgrading. Either way works for me though.

Maybe we should start having a GitHub milestone for 1.0 tasks.

> The "problem" with this is the lack of function overloading in Rust. It would mean changing the whole API depending on whether you want the IOMMU indirection or not.

grr, true. Forgot about that as well.

> For now we have only an RFC implementation [here](https://lore.kernel.org/qemu-devel/aJynpEWNfbrTeirQ@intel.com/T/). But you can see [that it needs to apply some changes to vm-memory](https://lore.kernel.org/qemu-devel/20250807123027.2910950-11-zhao1.liu@intel.com/) in order to add a direction (read/write) argument to try_access.
>
> Because QEMU has its backend in the C code, the natural thing for QEMU is to forgo GuestMemory completely and just go directly from IoMemory to GuestMemoryRegions.

Thanks, this is really helpful as a reference!

bonzini:
> Potentially, we could have less churn for downstream if we do the rename straight away. But only if downstream would upgrade to the new interfaces straight away

Gere's an idea. We could have (almost) simultaneously:
* a 0.(x+1) or 1.0 release with the new names
* the last 0.(x) release that just imports from 0.(x+1), and re-exports everything with GuestMemoryBackend renamed to GuestMemory (and the 0.(x+1) GuestMemory dropped completely from the public API)

This guarantees binary compatibility between the old and new API. I think we should do it, it's really now or never.

bonzini:
Can you place somewhere an extra commit to do the great rename GuestMemory->GuestMemoryBackend, IoMemory->GuestMemory? This way we can see how much breakage it causes.

XanClic:
> Can you place somewhere an extra commit to do the great rename GuestMemory->GuestMemoryBackend, IoMemory->GuestMemory? This way we can see how much breakage it causes.

Sure – I’ll can also show what this would mean for the vm-virtio and vhost crates.

roypat:
> > Potentially, we could have less churn for downstream if we do the rename straight away. But only if downstream would upgrade to the new interfaces straight away
> 
> Gere's an idea. We could have (almost) simultaneously:
> 
>     * a 0.(x+1) or 1.0 release with the new names
> 
>     * the last 0.(x) release that just imports from 0.(x+1), and re-exports everything with GuestMemoryBackend renamed to GuestMemory (and the 0.(x+1) GuestMemory dropped completely from the public API)
> 
> 
> This guarantees ~binary compatibility between~ across-crates coexistence of the old and new API. I think we should do it, it's really now or never.

Ahh, the semver trick, I forgot about that! Yes, let's do that for when it comes to the next release, I like that

XanClic:
> > Can you place somewhere an extra commit to do the great rename GuestMemory->GuestMemoryBackend, IoMemory->GuestMemory? This way we can see how much breakage it causes.
> 
> Sure – I’ll can also show what this would mean for the vm-virtio and vhost crates.

@bonzini: Here are some repos (renaming commits marked as “[preliminary]”):
* https://gitlab.com/hreitz/vm-memory/-/commits/iommu-renamed
* https://gitlab.com/hreitz/vm-virtio/-/commits/iommu-renamed
* https://gitlab.com/hreitz/vhost/-/commits/iommu-renamed
* https://gitlab.com/hreitz/virtiofsd-rs/-/commits/iommu-renamed

There’s a comment on [one of the vm-memory commits](https://gitlab.com/hreitz/vm-memory/-/commit/fcd6c6f9b7295792b53030dc249639a291ccecfe) about the [`guest_memory` module documentation](https://docs.rs/vm-memory/latest/vm_memory/guest_memory/index.html) mentioning “memory consumers (virtio device drivers, vhost drivers and boot loaders etc)”, specifically that these were in-VM users and so should probably continue to use `GuestMemory` (then `GuestMemoryBackend`), whereas `IoMemory` were only for users outside the VM, which would imply that maybe the names should stay as they are.

But reading it again, I’m no longer so sure. I’m bringing this up so someone can make it clear to me :)
* “vhost drivers” are not in-VM, but specifically what I mention as outside VM. Don’t know how I could confuse that. Either of the names `IoMemory` and `GuestMemory` make sense to me here.
* “boot loaders” should indeed largely continue to use physical memory. Renaming `IoMemory` to `GuestMemory` will require them to either use `GuestMemoryBackend` instead (clunky, and probably exactly what renaming is supposed to prevent) or specify the access mode and use `get_slices()`, for no functional reason. Not great, not terrible.
* “virtio device drivers” probably actually do need virtual memory support for IOMMU-enabled devices, so I’m just wrong saying these need access to physical memory. They do need access to the (potentially virtual) device memory address space. On the other hand, it’d be weird to then call this `GuestMemory` – it isn’t, it specifically is device memory. So we’d want to switch them to the new trait, but the name `IoMemory` would fit better than `GuestMemory` still, imho.

So all in all, I understand the switch makes sense to nudge all users to change their code so they can work with virtual memory if that becomes necessary; and the new trait still allows using all existing (physical) types, so it would all remain `GuestMemory`, nothing would *have to* use `GuestMemoryBackend`.

But on the other hand, as far as I can see, in all cases where we need to be able to access virtual memory it seems to be about I/O device memory, so the name `IoMemory` absolutely makes sense for it, and distinguishes it from all of the guest physical address space, which makes sense to continue to name `GuestMemory`.

bonzini:
> “boot loaders” should indeed largely continue to use physical memory

I'm not sure about that. How to allocate memory is a choice of the VMM, not the boot loader. Suppose a board has RAM at 0..2G and 4G..8G. Whether to allocate it as a single mmap or two should not be the boot loader's choice. If two (which is what something like firecracker would do) you can use `GuestMemory` in the boot loader. If one you could use your new trait to split the single backend into two regions, and only show an `IoMemory` to the boot loader.

A boot loader should also make sure to prepare the guest's memory view in such a way that the guest, after view, sees exactly what was prepared by the boot loader. Take as an example the PC's ability to remap the memory below 1MB to the corresponding area below 4GB. If a boot loader uses `GuestMemory`, it may set up the memory in a way that seems correct, but doesn't match the guest's view. The boot loader must then be able to interact with the chipset to ensure the guest's view is correct; passing it an `IoMemory` *may* make the boot loader code more complex, but it will also prevent this kinds of bugs.

Also, CPUs can have their own memory view, which is not the same as either the pure memory backend and not the same as any device. For example on x86 the local APIC is only visible to the CPUs.

So we could call your `IoMemory` something like `GuestMemoryView`, but I prefer to go all in and, as you say, nudge towards the transition.

Anyhow, looking at `vm-virtio` the changes are pretty small (beyond dropping `get_slice` and `get_host_address`, which of course are the largest part):

* changes to bitmaps (`B`->`Bitmap`):

```
-        <<M as GuestMemory>::R as GuestMemoryRegion>::B: WithBitmapSlice<'a, S = B>,
+        <M as GuestMemory>::Bitmap: WithBitmapSlice<'a, S = B>,
```

* adding permissions:

```
-        Ok(get_single_slice(mem, addr, length)?
+        Ok(get_single_slice(mem, addr, length, access)?
```

bonzini:
Also, given the intention to rename everything I would not create a separate io_memory module for IoMemory, and would instead create an iommu module for IommuMemory and its friends.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - \[[#311](https://github.com/rust-vmm/vm-memory/pull/311)\] Allow compiling without the ReadVolatile and WriteVolatile implementations
 - \[[#312](https://github.com/rust-vmm/vm-memory/pull/312)\] `GuestRegionContainer`, a generic container of `GuestMemoryRegion`s, generalizing `GuestMemoryMmap` (which 
   is now a type alias for `GuestRegionContainer<GuestRegionMmap>`).
+- \[[#327](https://github.com/rust-vmm/vm-memory/pull/327)\] I/O virtual memory support via `IoMemory`, `IommuMemory`, and `Iommu`/`Iotlb`
```
bonzini, line 10: This should be a new section of the CHANGELOG.md file (no need to update immediately).

src/iommu.rs
```diff
@@ -330,3 +351,198 @@ impl TryFrom<Range<u64>> for IovaRange {
...
+            use_iommu,
+        }
+    }
+
+    /// Create a new version of `self` with the underlying physical memory replaced.
+    ///
+    /// Note that the inner `Arc` reference to the IOMMU is cloned, i.e. both the existing and the
+    /// new `IommuMemory` object will share an IOMMU instance.  (The `use_iommu` flag however is
+    /// copied, so is independent between the two instances.)
+    pub fn inner_replaced(&self, inner: M) -> Self {
```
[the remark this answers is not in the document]
XanClic: Yes, that’s the pattern.

(In practice, `inner_replaced()` is called through a helper trait (`VirtualMemory`) that vhost-user-backend implements on both `IommuMemory` and `GuestMemoryMmap`, which provides a `with_physical_memory()` method that uses `inner_replaced()` for `IommuMemory` and just returns the new memory for `GuestMemoryMmap`.)

Why do you want to add that method?

Also, it isn’t generic across `IoMemory`, so it could only be used specifically for `IommuMemory`. I suppose we could add a trait `Foo<M: IoMemory>` with this method, and implement it for multiple `GuestMemoryExclusiveGuards<M>`. Then, vhost-user-backend could drop that `with_physical_memory()` method (but still needs the trait anyway). That alone doesn’t seem worth the effort here to me, but that’s why I’m wondering why you would like to have that method.

src/iommu.rs
```diff
@@ -330,3 +351,198 @@ impl TryFrom<Range<u64>> for IovaRange {
...
+            iommu: Arc::clone(&self.iommu),
+            use_iommu: self.use_iommu,
+        }
+    }
+
+    /// Enable or disable the IOMMU.
+    ///
+    /// Disabling the IOMMU switches to pass-through mode, where every access is done directly on
+    /// the underlying physical memory.
+    pub fn set_iommu_enabled(&mut self, enabled: bool) {
```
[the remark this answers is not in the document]
XanClic: Sure!

src/mmap/mod.rs
```diff
@@ -79,6 +85,16 @@ impl<B: Bitmap> GuestRegionMmap<B> {
...
             })
     }
+
+    /// Return a reference to the inner `Arc<MmapRegion>` (as opposed to
+    /// [`.deref()`](Self::deref()), which does not reference the `Arc`).
+    ///
+    /// The returned reference can be cloned to construct a new `GuestRegionMmap` with a different
+    /// base address (e.g. when switching between memory address spaces based on the guest physical
+    /// address vs. the VMM userspace virtual address).
+    pub fn get_mapping(&self) -> &Arc<MmapRegion<B>> {
```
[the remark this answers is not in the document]
XanClic: Sure!

src/guest_memory.rs
```diff
@@ -723,6 +724,157 @@ impl<T: GuestMemory + ?Sized> Bytes<GuestAddress> for T {
...
+    fn physical_memory(&self) -> Option<&Self::PhysicalMemory> {
+        None
+    }
+}
+
+/// Iterates over [`VolatileSlice`]s that together form an I/O memory area.
+///
+/// Returned by [`IoMemory::get_slices()`].
+pub trait IoMemorySliceIterator<'a, B: BitmapSlice>:
+    Iterator<Item = Result<VolatileSlice<'a, B>>> + Sized
```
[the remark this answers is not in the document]
bonzini: I think so, because you want any other implementations to be consistent. It's just a marker.

src/iommu.rs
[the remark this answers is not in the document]
bonzini: > Why do you want to add that method?

Because right now `GuestMemoryAtomic` focuses on replacing the whole `GuestMemory`, but there is a usecase for replacing the backend only.

> Also, it isn’t generic across IoMemory, so it could only be used specifically for IommuMemory

Yes, it's an extra `impl` only for `IommuMemory`.

> That alone doesn’t seem worth the effort here to me, but that’s why I’m wondering why you would like to have that method

More for clarity/documentation than anything else. `inner_replaced` was a bit mysterious (maybe it's just the name, `with_replaced_backend` or anything along those lines would help me too).

I see now that it can't be `pub(crate)` though!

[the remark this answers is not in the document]
XanClic: > there is a usecase for replacing the backend only.

True, but if we wanted to really provide for that use case, we’d need a generic `IoMemory` implementation (at least for all types in vm-memory).

> `inner_replaced` was a bit mysterious (maybe it's just the name, `with_replaced_backend` or anything along those lines would help me too).

Sure, I’d be more than happy to (and also rename `.inner()` to `.get_backend()`). I’ll definitely do this.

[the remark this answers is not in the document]
bonzini: Ok, let's do the rename and agree to disagree on the rest. :)

bonzini (approved):
This looks good to me now!

likebreath:
Late for the discussion. Just saw it from #362. I still need to find time to look into the changes closer. One high-level question: does the design expect VMMs to use the added interfaces around `iommu` and `iotlb`? 

I assume no, as VMM should always assume it is interacting with the Guest Physical Address space and use its own virtual IOMMU (say virtio-iommu and vSMMU) to handle address translation and access control, right?

bonzini:
> I assume no, as VMM should always assume it is interacting with the Guest Physical Address space and use its own virtual IOMMU (say virtio-iommu and vSMMU) to handle address translation and access control, right?

If the VMM implements an IOMMU, it will have to implement IoMemory. It is unlikely that it can reuse IommuMemory because the way it handles dirty bitmaps is not very suitable outside vhost_user.

PR #331, opened 2025-07-01
fix up #[cfg(...)] related to test code that uses dirty bitmaps

roypat:
Currently, vm-memory doesn't compile on all feature combinations: `cargo check --no-default-features --features backend-bitmap` does not compile because in `bitmap/backend/atomic_bitmap.rs` we try to use the `libc` crate to determine page size at some point. `libc` is an optional dependency and `backend-bitmap` does not enable it. Fixing this for production code is, of course, a one-liner in `Cargo.toml`. However, nothing is ever that straightforward. Turns out our _test_ code use some `#[cfg(test)]` directives to bypass cargo features, and always compile in the `bitmap/backend/mod.rs` module. This means `cargo test --no-default-features` is still broken. Instead of hacking around this by simply declaring libc an unconditional dev-dependency, ensure that tests that depend on bitmaps being available have the correct feature cfgs on them, and then drop the `cfg(test)` stuff.

There's a lot more cleanup that can be done here (all functions in `bitmap/mod.rs` probably do not belong there), but this is all I'm willing to do for today.

### Requirements

requirements:

  test.


roypat:
@andreeaflorescu @ShadowCurse sorry, it didnt cleanly rebase on top of main after merging #312, could you reapprove?

ShadowCurse (approved):

PR #338, opened 2025-08-12
Make GuestAddressSpace always cloneable

bonzini:
### Summary of the PR

The idea of `GuestAddressSpace` is that you would distribute as many references to it as needed.  However, `GuestAddressSpace` itself was not always cloneable because `GuestMemoryAtomic` uses `#[derive(Clone)]` which in turn requires *all type parameters* to be cloneable.

However, that is not a useful requirement for `GuestMemoryAtomic`, which is essentially an `Arc` wrapper. Fix that by implementing `Clone` by hand; this allows `GuestAddressSpace` to become a subtrait of `Clone` as well.

### Requirements

requirements:

  test.


roypat (dismissed):
Seems like `rustfmt` is demanding an addition comma, but functionally looks good!

roypat (approved):

bonzini:
Ping @ShadowCurse :)

PR #359, opened 2025-10-22
fix clippy warning about hidden life time

phi-nguyendp:
### Summary of the PR
Replace explicit lifetime names with the inferred lifetime placeholder '_ in return type paths to avoid "hiding a lifetime that's elided elsewhere is confusing" from clippy

### Requirements

requirements:

- [ x] All commits in this PR have Signed-Off-By trailers (with
- [ x] All added/changed functionality has a corresponding unit/integration
  test.
- [ x] All added/changed public-facing functionality has entries in the "Upcoming 
- [ x] Any newly added `unsafe` code is properly documented.


XanClic (changes requested):
I think this only fixes the warnings that appear when compiling without any non-default features enabled – there are more places that appear when enabling those features (each of backend-bitmap, backend-mmap, backend-atomic, xen).

PR #362, opened 2025-11-07
Rename GuestMemory->GuestMemoryBackend, IoMemory->GuestMemory 

bonzini:
### Summary of the PR

As discussed in #327, make the main interface to vm-memory the one that includes permissions and lets fewer aspects of the  implementation through.

Based on #361.

### Requirements

requirements:

  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Upcoming version
+## 0.18.0
 
 ### Changed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "vm-memory"
-version = "0.17.1"
+version = "0.18.0"
 description = "Safe abstractions for accessing the VM physical memory"
 keywords = ["memory"]
 categories = ["memory-management"]


src/guest_memory.rs
```diff
@@ -22,23 +22,23 @@
 //!   region.
 //! - [`GuestMemoryRegion`](trait.GuestMemoryRegion.html): represent a continuous region of guest's
 //!   physical memory.
-//! - [`GuestMemory`](trait.GuestMemory.html): represent a collection of `GuestMemoryRegion`
+//! - [`GuestMemoryBackend`](trait.GuestMemory.html): represent a collection of `GuestMemoryRegion`
```
XanClic: I count seven places like this, where the documentation link isn’t updated – did you leave out the `g` on the `sed`? (I know I did (to get a state to compare against) and got the same result at first 😅)

src/guest_memory.rs
```diff
@@ -167,9 +167,9 @@ impl FileOffset {
     }
 }
 
-/// `GuestAddressSpace` provides a way to retrieve a `GuestMemory` object.
+/// `GuestAddressSpace` provides a way to retrieve a `GuestMemoryBackend` object.
 /// The vm-memory crate already provides trivial implementation for
-/// references to `GuestMemory` or reference-counted `GuestMemory` objects,
+/// references to `GuestMemoryBackend` or reference-counted `GuestMemory` objects,
```
XanClic: (Mainly asking about the `g` because of this, where the second instance isn’t updated.)

src/guest_memory.rs
```diff
@@ -581,14 +581,14 @@ impl<'a, M: GuestMemory + ?Sized> IoMemorySliceIterator<'a, MS<'a, M>>
...
+/// Allow accessing [`GuestMemory`] (and [`GuestMemoryBackend`]) objects via [`Bytes`].
 ///
-/// Thanks to the [blanket implementation of `IoMemory` for all `GuestMemory`
-/// types](../io_memory/trait.IoMemory.html#impl-IoMemory-for-M), this blanket implementation
-/// extends to all [`GuestMemory`] types.
-impl<T: IoMemory + ?Sized> Bytes<GuestAddress> for T {
+/// Thanks to the [blanket implementation of `GuestMemory` for all `GuestMemoryBackend`
+/// types](../io_memory/trait.GuestMemory.html#impl-IoMemory-for-M), this blanket implementation
+/// extends to all [`GuestMemoryBackend`] types.
+impl<T: GuestMemory + ?Sized> Bytes<GuestAddress> for T {
```
XanClic: I think the documentation link should be `../guest_memory/trait.GuestMemory.html#impl-GuestMemory-for-M`.

(I think I’m already to blame for the wrong directory, though.)

roypat (approved):
Sorry for the late review!

Issue #337, opened 2025-08-08
Enable clippy::assertions_on_result_states for tests


roypat:
Our unit tests do a lot of `assert!((...).is_ok())` or `assert!((...).is_err())`, which in case of test failures produces very unhelpful messages (e.g. if an assertion on is_ok() fails, it just prints "assert failed: false", without printing what the error variant that we actually received was). The .is_ok() assertions, we can just replace with .unwrap()s, while the is_err() assertions can either become .unwrap_err(), if we dont care about the actual error returned, or `assert_matches!((...).unwrap_err(), expected error)`, if we do. After this is all set and done, we should add `#![warn(clippy(assertions_on_result_states)]` to `lib.rs`.

This is mostly busywork, and there are around 50 instances of this.

PR #365, opened 2025-11-21
Clippy test warnings

danielvallance:
This PR enables clippy::assertions_on_result_states, because asserting that a Result is_ok(), or is_err() causes the unit test failure messages to be quite unhelpful (as described in issue 337 https://github.com/rust-vmm/vm-memory/issues/337).

It also addresses a couple of other clippy warnings.


src/volatile_memory.rs
```diff
@@ -85,10 +85,17 @@ pub type Result<T> = result::Result<T, Error>;
...
 /// #
 /// assert_eq!(108, compute_offset(100, 8).unwrap());
-/// assert!(compute_offset(usize::MAX, 6).is_err());
+/// assert_matches!(
+///     compute_offset(usize::MAX, 6).unwrap_err(),
+///     Error::Overflow {
+///         base: 18446744073709551615,
+///         offset: 6
+///     }
+/// );
```
XanClic: I think in a public example it should really be `usize::MAX` instead of `18446744073709551615`.

src/volatile_memory.rs
```diff
@@ -1677,31 +1684,57 @@ mod tests {
...
+            slice.subslice(2, usize::MAX).unwrap_err(),
+            Error::Overflow {
+                base: 2,
+                offset: 18446744073709551615
+            }
+        );
 
         let maybe_offset_slice = slice.subslice(10, 80);
-        assert!(maybe_offset_slice.is_ok());
         let offset_slice = maybe_offset_slice.unwrap();
```
XanClic: Could be simplified to let `offset_slice = slice.subslice(10, 80).unwrap();` (similarly in another hunk below, with the same variable names).

XanClic (commented):
Looks good, just some nit picks, mainly:
- If we’re asserting a value in an address that’s been part of the previous code, it should have the same form as previously, i.e. assert on `usize::MAX` instead of `18446744073709551615` and on `0x600` instead of `1536` if the value was previously given as `0x600`. Especially so in examples that end up in documentation.
- I’m not sure about dropping `assert!(foo.is_ok())` from documentation examples just because it’s immediately followed by `foo.unwrap()`. I’m all for replacing `.is_ok()` by `.unwrap()` in tests, but in examples, `.is_ok()` adds an explicit “We’re not just unwrapping for simplicity’s sake, but this example is actually supposed to show that this operation will return `Ok(_)`”.

XanClic (approved):
Thanks! :)

src/volatile_memory.rs
```diff
@@ -1677,31 +1684,57 @@ mod tests {
...
-        assert!(slice.write(&[1; 80], 10).is_ok());
+        slice.write(&[1; 80], 10).unwrap();
 
-        assert!(slice.subslice(0, 0).is_ok());
-        assert!(slice.subslice(0, 101).is_err());
+        slice.subslice(0, 0).unwrap();
+        assert_matches!(
+            slice.subslice(0, 101).unwrap_err(),
+            Error::OutOfBounds { addr: 101 }
+        );
```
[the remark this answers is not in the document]
roypat: iirc there are some subtleties where we can request a slice of length 0 one byte past the end of a volatile slice, and it'll work (e.g. if length is 80, then doing subslice(80, 0) will give you an empty slice), and I suspect this is an artifact of that. There's some more oddities like this, such as some read/write impls allowing you to always read/write length-0 slices into guest memory even if the address you pass in is complete bollocks :/

rbradford (approved):

PR #373, opened 2026-03-23
Make the 'rawfd' feature a NOP on Windows

germag:
### Summary of the PR

When `rawfd' feature was introduced on commit 0b7a4377, it was enabled
by default to avoid breaking the default behavior on linux. So, on 
Windows, vm-memory must be included with 'default-features = false' 
otherwise it fails to compile. This is problematic for projects that
depend indirectly on vm-memory and fail if their dependencies don't
set 'default-features = false'.

Since there is no current way to set the default feature set for a
specific target OS, let's make the 'rawfd' feature a NOP for Windows
builds, the feature will still be enabled, but silently just not do
anything.

### Requirements

requirements:

- [] All added/changed functionality has a corresponding unit/integration
  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+### Changed
+
+### Fixed
+
+### Removed
+
+### Deprecated
+
 ## 0.18.0
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 ### Changed
+- \[[#373](https://github.com/rust-vmm/vm-memory/pull/373)\] Make the 'rawfd' feature a NOP on Windows
 
 ### Fixed
 
--- a/benches/mmap/mod.rs
+++ b/benches/mmap/mod.rs
@@ -7,11 +7,11 @@
 extern crate criterion;
 extern crate vm_memory;
 
-#[cfg(feature = "rawfd")]
+#[cfg(all(feature = "rawfd", not(target_os = "windows")))]
 use std::fs::{File, OpenOptions};
 use std::mem::size_of;
 
-#[cfg(feature = "rawfd")]
+#[cfg(all(feature = "rawfd", not(target_os = "windows")))]
 use std::path::Path;
 
 use core::hint::black_box;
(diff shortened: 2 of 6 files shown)


germag:
I can make this a bit more clear using [cfg_aliases](https://docs.rs/cfg_aliases/latest/cfg_aliases/), so instead of 
```rust
#[cfg(all(feature = "rawfd", not(target_os = "windows")))]
```

we can use something like 
```rust
#[cfg("rawfd")]
```

XanClic:
I don’t like it very much that this way, on Windows, the feature will still be enabled, but silently just not do anything. I know that is a very lukewarm stance and a problem you have probably already considered, but it is not weighed on in the commit message, hence me commenting on it.

So: There is no other way, right? To change the default-features list only on Windows, or at least any way to detect the feature being enabled explicitly, and fail then.

(In any case, CHANGELOG.md should get an entry, I think.)

germag:
> I don’t like it very much that this way, on Windows, the feature will still be enabled, but silently just not do anything. I know that is a very lukewarm stance and a problem you have probably already considered, but it is not weighed on in the commit message, hence me commenting on it.

Yes, I'm not happy too, but I don't see another way without breaking the default behavior on linux

> So: There is no other way, right? To change the default-features list only on Windows, or at least any way to detect the feature being enabled explicitly, and fail then.

I'm not aware of another way, sadly there is nothing like  `[target.'cfg(windows)'.features]`.
I know other projects that have the same issue and doing this (some using `cfg_aliases` to make it more palatable)

> (In any case, CHANGELOG.md should get an entry, I think.)

Ok, I'll explain the problem in the commit msg and add and changelog entry

germag:
> any way to detect the feature being enabled explicitly, and fail then.

Maybe (not sure, I never used it) I can use [cargo_metadata](https://crates.io/crates/cargo_metadata) to do some heuristic inside a `build.rs` it feels it will be quite hacky, tho

germag:
> > any way to detect the feature being enabled explicitly, and fail then.
> 
> Maybe (not sure, I never used it) I can use [cargo_metadata](https://crates.io/crates/cargo_metadata) to do some heuristic inside a `build.rs` it feels it will be quite hacky, tho

It seems it will not work for this purpose.

germag:
I did this to keep the SemVer compatibility, but since the next 0.18.0 version already includes a major change in the API, I can just remove the `rawfd` feature from the default set, @XanClic wdyt?

bonzini:
> I can just remove the rawfd feature from the default set, @XanClic wdyt?

0.18.0 has already been released, unfortunately.

germag:
> > I can just remove the rawfd feature from the default set, @XanClic wdyt?
> 
> 0.18.0 has already been released, unfortunately.

Ops my mistake I'll change the CHANGELOG then

XanClic (approved):

PR #375, opened 2026-04-06
`mmap/unix.rs`: Add `with_minimum_alignment` to `MmapRegionBuilder`

jaredwhitedev:
Add an option to specify a minimal alignment for the mmap'd region. This is useful to ensure VM memory backings get huge EPT entries.

### Summary of the PR

To get huge EPT entries callers must be able to align their memory to larger page sizes, such as 2 MiB or 1 GiB.
Using huge EPT entries increases the memory performance by decreasing the TLB pressure.

Expose a `with_minimum_alignment` option to `MmapRegionBuilder` to allow callers to do so.

To support this, we first allocate `size+alignment_size` of virtual address space, then use `MAP_FIXED` to place the actual mapping at an aligned address within the reserved VA space. We then free the excess
### Requirements

requirements:

- [X] All commits in this PR have Signed-Off-By trailers (with
- [X] All added/changed functionality has a corresponding unit/integration
  test.
- [X] All added/changed public-facing functionality has entries in the "Upcoming 
- [X] Any newly added `unsafe` code is properly documented.

### DIFF
--- a/src/mmap/unix.rs
+++ b/src/mmap/unix.rs
@@ -40,14 +40,25 @@ pub enum Error {
     /// The `mmap` call returned an error.
     #[error("{0}")]
     Mmap(io::Error),
+    /// Invalid aligment specified.
+    #[error("The minimum alignment is not a power of two and at least the system page size")]
+    InvalidAlignment,
 }
 
 pub type Result<T> = result::Result<T, Error>;
+
+/// Retrieve the system page size from sysconf
+fn system_page_size() -> usize {
+    // SAFETY: Safe because this call just returns the page size and doesn't have any side
+    // effects.
+    unsafe { libc::sysconf(libc::_SC_PAGESIZE) as usize }
+}
 
 /// A factory struct to build `MmapRegion` objects.
 #[derive(Debug)]
 pub struct MmapRegionBuilder<B = ()> {
     size: usize,
+    alignment: usize,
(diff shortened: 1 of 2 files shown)


jaredwhitedev:
Seems that we can't allocate with more than a 512 MiB alignment when using Miri. Would you rather we:
a) Try best effort for Miri (Let the allocation failure happen whenever caller uses more than 512 MiB alignment). Current behavior.
b) Don't support Miri when using with_alignment (explicit error)
c) Ignore it when using Miri (default 8 byte alignment)

src/mmap/unix.rs
```diff
@@ -94,6 +106,13 @@ impl<B: Bitmap> MmapRegionBuilder<B> {
         self
     }
 
+    /// Create the `MmapRegion` object with the specified minimum alignment.
+    /// `min_align` must be a power of two and at least the system page size.
+    pub fn with_minimum_alignment(mut self, min_align: usize) -> Self {
```
liuw: Non-conformant values are rejected in the build function, but I think it will be better to reject them here. This function can return a `Result<Self>`. In the error case you return `InvalidAlignment`.

This is a cleaner approach. Users of the alignment value don't need to validate it again.

jaredwhitedev: Updated

PR #382, opened 2026-06-18
region: document GuestMemoryRegion::len() non-zero invariant

### LINKED ISSUE #50: vm-memory: Clarify that custom GuestMemoryRegion implementations must not return zero length
The built-in mmap-backed region path already rejects zero-sized mappings, so this doesn't seem to affect the normal `GuestMemoryMmap` flow.

However, `GuestMemoryRegion` is a public extension trait intended for custom memory backends, and I couldn't find documentation stating that `len()` must be non-zero.

A custom implementation returning `0` can be accepted by `GuestRegionCollection::from_regions()`, but later code assumes non-empty regions:

```rust
self.start_addr().unchecked_add(self.len() - 1)
```

This makes the invariant implicit rather than documented.

Aakash456:
## Summary

Documents that `GuestMemoryRegion::len()` must always return a non-zero value, since `last_addr()`'s default implementation computes `self.len() - 1` and would otherwise underflow.

This invariant was previously enforced implicitly for built-in mmap-backed regions but undocumented for the public `GuestMemoryRegion` trait, which custom backends also implement.

## Changes

- Added a `# Invariant` section to the doc comment on `GuestMemoryRegion::len()`
- Added a `debug_assert!` in `last_addr()` to fail fast in debug builds if a custom implementation violates the invariant
- Corrected the existing comment on the `unchecked_add` call, which previously implied the safety guarantee holds universally — it only held for the built-in mmap region, not for arbitrary custom implementations

Fixes rust-vmm/rust-vmm#50

Happy to adjust scope if you'd prefer this as doc-only without the `debug_assert`, or if a hard `assert!` is preferred over `debug_assert!` given this is a safety-adjacent invariant.


src/region.rs
```diff
@@ -43,14 +43,34 @@ pub trait GuestMemoryRegion: Bytes<MemoryRegionAddress, E = GuestMemoryError> {
...
+    /// already enforce this when constructed; custom implementations of `GuestMemoryRegion` are
+    /// responsible for upholding the same invariant.
     fn len(&self) -> GuestUsize;
 
     /// Returns the minimum (inclusive) address managed by the region.
     fn start_addr(&self) -> GuestAddress;
 
     /// Returns the maximum (inclusive) address managed by the region.
+    ///
+    /// # Panics (debug builds only)
```
andreeaflorescu, line 61: We don't typically document debug panics, I don't think this is necessary.

andreeaflorescu (commented):
Thanks for the PR. I just left a small comment on the documentation.

We now moved this crate to the rust-vmm monorepo. To be able to merge this, can you please re-create your PR there: https://github.com/rust-vmm/rust-vmm

